PR #19, opened 2020-09-19
Add an update command to fetch and store the latest rootTreeHash/Size.

dlorenc:
If we have a previous set of values, we also check the consistency proof between
these two to make sure the tree has not been tampered with.

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -1,0 +1,185 @@
+/*
+Copyright © 2020 Luke Hinds <lhinds@redhat.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


cmd/update.go
```diff
@@ -0,0 +1,185 @@
...
+	p, err := stateDir()
+	if err != nil {
+		return nil
+	}
+	var oldState state
+	f, err := ioutil.ReadFile(p)
+	if err != nil {
+		log.Info(err)
+		return nil
+	}
```
lukehinds, line 80: This is really good Dan, making some great progress here.

A tiny nit: First time running we get a `.rekor/rekor.json: no such file or directory`. Its are capturing the err as it should and is correct and proper, but I guess we could put a pretty message there instread, as a file not existing is expected on first run. I will merge and follow up.

lukehinds (approved):

PR #21, opened 2020-09-20
Get Leaf command

lukehinds:
Gets a specific leaf by the leaf Index

`rekor getleaf --index ($index)


cmd/getleaf.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		log.Info("Leaf contents saved to ", outfile)
+
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(getleafCmd)
+	getleafCmd.PersistentFlags().Int64("index", -1, "")
+	getleafCmd.PersistentFlags().String("outfile", GenerateRand(20)+".txt", "file name to output leaf content")
```
dlorenc, line 117: What about naming the file with the index instead of a random prefix? The index should also be unique within the tree.

test.txt
```diff
@@ -0,0 +1,29 @@
+{
+ "signatures": [
```
dlorenc: Did you mean to include this file?

lukehinds: Nope :+1:

cmd/getleaf.go
lukehinds: the index would just be `1.txt` - I am not against this, although its very sexy :)

lukehinds: actually it does make sense

PR #27, opened 2020-10-20
Add verify command.

dlorenc:
I still need to test this out on a few examples. The logic to verify a Rekor entry is getting a little long now
and we might want to start to share it between server and client somehow.


lukehinds (approved):
lgtm!

lukehinds:
I think the conflict from from linkfile -> rekord

PR #34, opened 2020-11-11
Implement upload command

lukehinds:
Accept flags artifact-url,public-key, signature and validate GPG formatting and verify file signature

This will handle both armoured files as well as binary


app/gpg.go
```diff
@@ -0,0 +1,115 @@
...
+
+func FormatPubKey(pubkeyFilePath string) (string, error) {
+	pubKey, err := ioutil.ReadFile(pubkeyFilePath)
+
+	if err != nil {
+		return "Error opening File: ", err
+	}
+
+	if strings.Contains(string(pubKey), "-----BEGIN PGP") {
+		b, err := os.Open(pubkeyFilePath)
```
bobcallaway: you're reading the file twice here; perhaps remove the call on line 22 and just reuse `pubKey` since you've already read all the bytes in ?

app/gpg.go
```diff
@@ -0,0 +1,115 @@
...
+func FormatSignature(sigFilePath string) (string, error) {
+	sigKey, err := ioutil.ReadFile(sigFilePath)
+
+	if err != nil {
+		return "Error opening File: ", err
+	}
+
+	if strings.Contains(string(sigKey), "-----BEGIN PGP") {
+		// It's an armored file
+		b, err := os.Open(sigFilePath)
```
bobcallaway: same as above - you can probably just get away with re-reading `sigKey` instead of re-reading it?

README.md
```diff
@@ -1,47 +1,53 @@
...
 
 Rekor's goals are to provide an immutable tamper resistant ledger of metadata generated within a software projects supply chain.  Rekor will enable software maintainers and build systems to record signed metadata to an immutable record. Other parties can then query said metadata to enable them to make informed decisions on trust and non-repudiation of an object's life-cycle, based on signed metadata stored within a tamper proof binary (merkle) tree.
 
-The Rekor CLI requires a running instance of the [rekor-server](https://github.com/projectrekor/rekor-server).
+Rekor requires a running instance of the [rekor-server](https://github.com/projectrekor/rekor-server).
+
+## Sign your release
+
+Before using rekor, you are required to sign your release. For now we use GPG
+(we plant to extend to other signing formats in the foreseeable future)
```
bobcallaway, line 12: I was thinking about this a bit after the API writeup... it might be best to break out the various signature/key options into separate interfaces and reuse this code between server & client.

cmd/upload.go
```diff
@@ -13,30 +13,248 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		// Generate a file name based off the artifact hash
+		rekorFile := generateRekorFile(generatedSha)
+
+		log.Info("Building rekor file : ", rekorFile)
+
+		// Construct rekor json file
+		// We need to approach this in two ways
+		// as the public key and signature could be either
+		// armored or binary
+		if isArmorProtected(sigkeyRingReader) || isArmorProtected(sigkeyRingReader) {
```
bobcallaway: I think you meant for the one of the predicates to be `pubkeyRingReader`

bobcallaway (commented):
a couple thoughts

app/gpg.go
lukehinds: agree, well spotted.

README.md
lukehinds: Yes, we should make this a plug-able thing, where someone can use a signing system of their preference (if this is what you mean). We should also look at a similar framework for allowing ingestion of different manifest types (in-toto, sbom etc).

cmd/root.go
```diff
@@ -55,17 +55,17 @@ func init() {
...
 
-	rootCmd.PersistentFlags().String("artifact_url", "", "Rekor artifact url")
-	viper.BindPFlag("artifact_url", rootCmd.PersistentFlags().Lookup("artifact_url"))
+	rootCmd.PersistentFlags().String("artifact-url", "", "Rekor artifact url")
+	viper.BindPFlag("artifact-url", rootCmd.PersistentFlags().Lookup("artifact-url"))
 
-	rootCmd.PersistentFlags().String("artifact_sha", "", "Rekor artifact sha")
-	viper.BindPFlag("artifact_sha", rootCmd.PersistentFlags().Lookup("artifact_sha"))
+	rootCmd.PersistentFlags().String("artifact-sha", "", "Rekor artifact sha")
+	viper.BindPFlag("artifact-sha", rootCmd.PersistentFlags().Lookup("artifact-sha"))
```
bobcallaway, line 68: Similar to my latest Gosec patch on the server side, perhaps consolidate into a single BindPFlag call at the end?

lukehinds: let's do that in a separate PR , keep the commits a bit more siloed to the change.

cmd/upload.go
```diff
@@ -13,30 +13,248 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		sigkeyRingReader, err := os.Open(signature)
+		if err != nil {
+			log.Error("Error opening signature: ", err)
+		}
+
+		var keyRing openpgp.EntityList
+		if isArmorProtected(pubkeyRingReader) {
+			keyRing, err = openpgp.ReadArmoredKeyRing(pubkeyRingReader)
+		} else {
+			keyRing, err = openpgp.ReadKeyRing(pubkeyRingReader)
```
dlorenc, line 175: nit: i think we're missing this errcheck

cmd/upload.go
```diff
@@ -13,30 +13,248 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		// armored or binary
+		if isArmorProtected(sigkeyRingReader) || isArmorProtected(pubkeyRingReader) {
+			rekorArmorJSON := RekorArmorEntry{
+				URL:       artifactURL,
+				SHA:       generatedSha,
+				Signature: sig,
+				PublicKey: pub_key,
+			}
+			file, _ := json.MarshalIndent(rekorArmorJSON, "", " ")
+			_ = ioutil.WriteFile(rekorFile, file, 0644)
```
dlorenc, line 217: nit: errcheck

cmd/upload.go
```diff
@@ -13,30 +13,248 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		// as the public key and signature could be either
+		// armored or binary
+		if isArmorProtected(sigkeyRingReader) || isArmorProtected(pubkeyRingReader) {
+			rekorArmorJSON := RekorArmorEntry{
+				URL:       artifactURL,
+				SHA:       generatedSha,
+				Signature: sig,
+				PublicKey: pub_key,
+			}
+			file, _ := json.MarshalIndent(rekorArmorJSON, "", " ")
```
dlorenc, line 216: Any reason for using indent here?

cmd/upload.go
```diff
@@ -13,30 +13,248 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+			_ = ioutil.WriteFile(rekorFile, file, 0644)
+		} else {
+			pubKey, err := ioutil.ReadFile(publicKey)
```
dlorenc, line 219: nit: errcheck

cmd/upload.go
```diff
@@ -13,30 +13,248 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+}
+
+type RekorEntry struct {
+	SHA       string `json:"SHA,omitempty"`
+	URL       string `json:"URL,omitempty"`
+	Signature []byte `json:"Signature"`
+	PublicKey []byte `json:"PublicKey"`
+}
+
+type RekorArmorEntry struct {
```
dlorenc, line 60: RekorArmorEntry and RekorEntry are the same afaict, why do we need both types?

lukehinds: Sure, they use different types for Signature and PublicKey

RekorEntry:
```
	Signature []byte `json:"Signature"`
	PublicKey []byte `json:"PublicKey"`
```
RekorArmorEntry:
```
	Signature string `json:"Signature"`
	PublicKey string `json:"PublicKey"`
```

This is so that we can handle armored and binary gpg artifacts.

cmd/upload.go
lukehinds: No idea, maybe my IDE formatted it, I will change it

cmd/upload.go
lukehinds: `file, _ := json.MarshalIndent(rekorArmorJSON, "", " ")` = `file, err := json.MarshalIndent(rekorArmorJSON, "", " ")`?

cmd/upload.go
dlorenc: Ah, what do you think about merging these two into one struct?

```go
type RekorEntry struct {
	SHA       string `json:"SHA,omitempty"`
	URL       string `json:"URL,omitempty"`
	Signature []byte `json:"Signature"`
	PublicKey []byte `json:"PublicKey"`
        AmoredSignature string `json:"Signature"`
	AmoredPublicKey string `json:"PublicKey"`
}
```

lukehinds: I guess so, will it not fail if a key is missing? For example there is no value for `AmoredPublicKey` as its a binary instance?

PR #36, opened 2020-11-18
Move the Rekor types from server to this repo.

dlorenc:
This also breaks the go module dependency on rekor-server.


lukehinds (approved):
LGTM

PR #43, opened 2020-11-20
Bring back URL into manifest

lukehinds:
The URL was not being stored and subsequently returned when
performing getLeaf by Index calls.

We should includes this information as its useful for monitors
who may want to query for the this dataset.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -36,12 +36,12 @@ import (
 // RekorEntry is the API request.
 type RekorEntry struct {
 	Data      []byte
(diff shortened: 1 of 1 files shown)


bobcallaway (dismissed):

dlorenc:
Hmm, should this be part of the Tlog? Maybe this is a good candidate for the extra data we can store that stays outside the hash.

lukehinds:
I will close this and we go with @dlorenc recommendation of using Extra data

PR #46, opened 2020-11-23
Marshal the JSON correctly in verify, and properly load keys.

dlorenc:


lukehinds:
LGTM!

pkg/types/types.go
```diff
@@ -152,6 +158,10 @@ func (r *RekorEntry) Load(ctx context.Context) error {
 		dataReader = bytes.NewReader(r.Data)
 	}
 
+	if err := r.ParseKeys(); err != nil {
```
bobcallaway, line 161: I'd move this call before the URL fetch. There's no reason to even call `Get()` if we don't have a valid key/sig pair.

pkg/types/types.go
```diff
@@ -90,16 +97,15 @@ func ParseRekorLeaf(r io.Reader) (*RekorLeaf, error) {
...
 	if err != nil {
-		return nil, err
+		return err
 	}
-
-	return &l, nil
+	return nil
 }
 
 func ParseRekorEntry(r io.Reader, leaf *RekorLeaf) (*RekorEntry, error) {
```
bobcallaway, line 111: We should validate the pointer passed in here as well.

pkg/types/types.go
```diff
@@ -77,10 +77,17 @@ func ParseRekorLeaf(r io.Reader) (*RekorLeaf, error) {
 		return nil, err
 	}
 
+	if err := l.ParseKeys(); err != nil {
+		return nil, err
+	}
+	return &l, nil
+}
+
+func (l *RekorLeaf) ParseKeys() error {
```
bobcallaway, line 86: Since you're also "validating" the SHA, perhaps name it something like "ValidateLeaf"?

PR #45, opened 2020-11-20
first pass at openapi doc

bobcallaway:
In an attempt to formalize the server API, I wrote up this OpenAPI description of my proposal in https://docs.google.com/document/d/1Iy9TSGyUwfD6Ww5jro_V_pK8VEGNH_W7lYUT3B_4SDg/edit 

@dlorenc @lukehinds I'd appreciate any feedback on the API itself, along with opinions on whether or not we should consider using this and https://github.com/go-swagger/go-swagger to generate client and/or server code.

PR #49, opened 2020-11-23
simplify upload command to reuse common PKI

bobcallaway:


cmd/cli/app/upload.go
```diff
@@ -17,29 +17,22 @@ package app
...
+	"github.com/projectrekor/rekor/pkg/types"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"golang.org/x/crypto/openpgp"
-	"golang.org/x/crypto/openpgp/armor"
 )
 
 type RespStatusCode struct {
-	Code string `json:"file_recieved"`
+	Code string `json:"file_received"`
```
dlorenc, line 35: Should this actually say "code"?

dlorenc (approved):
One nit

cmd/cli/app/upload.go
bobcallaway: Not sure its important given that its only used for unmarshalling the response. This should probably be pulled out into a common types/something.go file to maintain consistency b/t client and server. 

This would be addressed by #45 (if the code-gen approach was taken) or could be handled in another PR.

PR #57, opened 2020-11-25
Types unit test

bobcallaway:
add unit tests for types.go


pkg/types/types_test.go
```diff
@@ -0,0 +1,473 @@
...
...hpdmUKQ29tbWVudDogZ2l0IGFyY2hpdmUgLS1mb3JtYXQgdGFyIC0tcHJlZml4PWxpbnV4LTUuOC45LyB2NS44LjkKQ29tbWVudDogZ2l0IHZlcnNpb24gMi4yOC4wCgppUUl6QkFBQkNBQWRGaUVFWkg4b1pVaVU0NzFGY1ptK09OdTl5R0NTYVQ0RkFsOWN2YWdBQ2drUU9OdTl5R0NTCmFUNUMvZy8vWnZweTVMZHZHSUVYdkhIZ2xXck1CZDRPUy9ZYzZzbk84aDBZSWpGcnVIR1RQU3ZlbTZLRzByaEYKdy9McGV2aFhwc0IzUjh1N01LN2VNVG5rYVRkanpGbWhreldHZVptcW1PaWRMaGMrWm5WV012dHNnUjJTRXVPNApWd2Vzb2ZvSlkybmswdXhqNW1CcE1ZNkFsRG9oanlteWZzY3diSVNqNkZobTFpcXlNOGpWV3JxVU40WDZCV2dkCnBVS2ovUWRjcmdkS0FHRkpCMGJVeFEyR2I5aHczWmhrQndQRWNhZ2hKNU53Rm44ZzRJTm96UTY2RTBTbEdTK3cKSVNOZEh1d2x6VVg3WitsaUYvVUpsdkllV1ZSZTl3SUNkZnhRSTJqL2RJK1hqblc0VkNoWkxFU3dJeWxadDhqVwpaeDk2d1dPOHlFUmNrd1NablpWSjdrS3dBdmxRVitlNEQ3NkZZQUUzWURySThsTnFuL1hLTXhrZEpqS2hIVGNNCm8waXc3TVFuM2ZFdWxRNFVEd0c3UGEzTHNKeExxMkRVdkI4bG12bng3OG11VDZKTnVhOVVuYzNFTlptbER6YzcKb0pVcUlLeThkWTZZWFFCTWNNcUpoNUR1UkRWRXdKNm55WVlWMTRzWm1rRzgvV29mUkdReHVvY1JjT1RCQk12SgppZkRRb2IyWVdpRUd0UEpIdHh2VWI5RTZ4YnJoR2lscjJncmRDRHFMUUF3QkZTeWlCSy9Kc1dHTE5aMThnUEF6CmJlMGlYMk5mYUxlQlFKZUE2QjMxZk5TcStndHcvMzRBemVOTnNkT1BuZXlYcXN0OUs1WlpCWnFwMUMvZTljWFcKMS9YTlhWeXUwelp1Mkc2VHFlVFpyT3R4NVM4bEVvUjFpM09yQXN0Z0hGT2FibGRFc0VBPQo9WU14VgotLS0tLUVORCBQR1AgU0lHTkFUVVJFLS0tLS0K")
```
dlorenc: ick, having all of these strings makes me worry about maintenance/debugging here. How many unique ones are there in total? Could we use files in testdata/ like the other test?

pkg/types/types_test.go
```diff
@@ -0,0 +1,473 @@
...
+		caseDesc      string
+		json          string
+		expectSuccess bool
+	}
+
+	tests := []Test{
+		{
+			caseDesc: "Valid full entry",
+			json: `{
+				"URL": "https://mirrors.edge.kernel.org/pub/linux/kernel/v5.x/linux-5.8.9.tar.gz",
```
dlorenc: We don't actually hit these URLs, right? I don't think we do here, just want to double check though.

pkg/types/types_test.go
bobcallaway: yeah, I can clean this up.

pkg/types/types_test.go
bobcallaway: correct. I can change this value to something else since the point of this particular test is to cover extra fields in the JSON

lukehinds:
are you good with this now @dlorenc ?

lukehinds (approved):

PR #59, opened 2020-11-27
A few code cleanups.

dlorenc:
- Pass in ctx instead of the full request to the logger
- Handle closing the body in our wrapper

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -22,7 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
+	"io/ioutil"
 	"net/http"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


lukehinds (approved):
LGTM, but will also defer to @bobcallaway as I don't know the requestIDlogger code to well.

pkg/api/api.go
```diff
@@ -205,15 +211,15 @@ func (api *API) getProofHandler(r *http.Request) (interface{}, error) {
 }
 
 func (api *API) addHandler(r *http.Request) (interface{}, error) {
-	defer r.Body.Close()
-
-	var byteEntry bytes.Buffer
-	tee := io.TeeReader(r.Body, &byteEntry)
+	b, err := ioutil.ReadAll(r.Body)
```
bobcallaway, line 214: In this approach, the entire request body will be read into memory immediately, whereas the current code only does this *iff* the leaf content is successfully parsed and checked that it is not already in the tLog. 

While the major reason to do this is to defer parsing the `Data` element of the incoming JSON doc (until we absolutely need to, since it might be huge), it also provides some OOM protection from clients sending huge malformed JSON documents to the server. 

If we're still planning to drop the `Data` element then admittedly the current code doesn't provide as much benefit, but in general I'm suspect of using `ioutil.ReadAll()` in a case where the input stream isn't capped in some way.

pkg/api/api.go
```diff
@@ -112,23 +112,28 @@ type apiHandler func(r *http.Request) (interface{}, error)
...
 		r = r.WithContext(log.WithRequestID(ctx, middleware.GetReqID(ctx)))
 		defer func() {
-			_ = log.RequestIDLogger(r).Sync()
+			_ = log.RequestIDLogger(ctx).Sync()
 		}()
 
 		respObj, err := h(r)
 		if err != nil {
 			writeError(w, err)
+			return
```
bobcallaway, line 125: Returning here will cause a 500 error reported to the client with nothing in the HTTP response body. I'd think we would want to return some sort of an error message back to the client (which the prior code supports, but is unused). Thoughts?

bobcallaway (commented):
I'm fine with the RequestIDLogger changes, but not sure about two of the other ones.

PR #65, opened 2020-12-05
add log_proof impl

bobcallaway:


cmd/cli/app/verify.go
```diff
@@ -35,6 +35,17 @@ var verifyCmd = &cobra.Command{
 	Use:   "verify",
 	Short: "Rekor verify command",
 	Long:  `Verifies a signature and checks that it exists in the transparency log`,
+	PreRun: func(cmd *cobra.Command, args []string) {
+		// these are bound here so that they are not overwritten by other commands
+		if err := viper.BindPFlags(cmd.Flags()); err != nil {
+			log.Logger.Fatal("Error initializing cmd line args: ", err)
+		}
+		if err := validateRekorServerURL(); err != nil {
```
dlorenc, line 43: Maybe we could handle this in a custom flag type instead of needing to put it in the PreRun everywhere?

https://golang.org/pkg/flag/#Value

openapi.yaml
```diff
@@ -36,22 +36,24 @@ paths:
...
           name: firstSize
           type: integer
-          default: 0
-          minimum: 0
+          default: 1
+          minimum: 1
           description: >
-            The size of the tree that you wish to prove consistency from (0 means the beginning of the log)
-            Defaults to 0 if not specified
+            The size of the tree that you wish to prove consistency from (1 means the beginning of the log)
```
dlorenc, line 42: Just curious, why is 1 the start rather than 0?

dlorenc (approved):

cmd/cli/app/verify.go
bobcallaway: +1 on the validation change; however, if we want consistent flags between sub commands we need the BindPFlags call in prerun so that only the command being executed gets the values bound to them.

openapi.yaml
bobcallaway: https://github.com/google/trillian/blob/2053c7648b44d5de45863c3ad12550b511ad6a14/server/validate.go#L83 wants > 0 for both sizes

dlorenc: Ah strange, thanks!

cmd/cli/app/verify.go
dlorenc: Makes sense. I'll handle the validation part.

dlorenc (approved):

PR #66, opened 2020-12-05
add XML & YAML support to API

bobcallaway:
other than one little quirk when sending `Accept: */*` (which is the default with `curl`), this was pretty easy.


lukehinds (changes requested):
Looks good. Could you add some notes on how the the UX will play out in regards to the client / CLI.

Do they use the artifact_path and point to a local / json / yaml / xml file? Be as verbose as you like as its useful for documentation.

lukehinds (approved):

PR #69, opened 2020-12-07
Consolidate flags under root init()

lukehinds:
This sets the same flags for both upload & verify and also
consolidates all commands into the cobras root init

Signed-off-by: Luke Hinds <lhinds@redhat.com>

### DIFF
--- a/cmd/cli/app/verify.go
+++ b/cmd/cli/app/verify.go
@@ -30,7 +30,6 @@ import (
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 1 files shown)


bobcallaway:
I had a similar change pending that I just pushed up in #71 - the big thing I found is that we need to move the viper.BindPFlags call into each subcommand's `PreRun` command. Otherwise the last command added to `rootCmd` will get the flags bound to it.

bobcallaway:
Closing as we moved forward with #71 instead

PR #75, opened 2020-12-18
Add a basic e2e test script and test data.

dlorenc:
This runs against a running setup, and does a single upload of some node
data.


dlorenc:
Thanks for the commits!

PR #77, opened 2020-12-19
Move the API context to a global middleware.

dlorenc:
I think this is the same thing as registering it on each handler, not
100% sure though.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/google/martian v2.1.0+incompatible
 	github.com/google/trillian v1.3.10
 	github.com/gorilla/handlers v1.5.1 // indirect
+	github.com/jessevdk/go-flags v1.4.0
 	github.com/kr/pretty v0.2.1 // indirect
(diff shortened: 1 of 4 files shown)


bobcallaway:
The only comment I have is that by putting it in the global middleware, we will instantiate a GRPC client for every transaction regardless of whether it’s a malformed request or whether it’s an invalid URL; that’s why I did it where I did. I don’t know whether we want to save the connection overhead on the backend by only applying it on a per handler basis after the route has been validated. Let me know what you think.

dlorenc:
> The only comment I have is that by putting it in the global middleware, we will instantiate a GRPC client for every transaction regardless of whether it’s a malformed request or whether it’s an invalid URL; that’s why I did it where I did. I don’t know whether we want to save the connection overhead on the backend by only applying it on a per handler basis after the route has been validated. Let me know what you think.

Ah, good point. I actually moved the connection setup to the server initialization now. I think this is the right way to handle GRPC connections, they're go-routine safe and we're meant to use one for the entire process.

bobcallaway (approved):

PR #79, opened 2021-01-02
Checks correct response code to prevent duplicate insertion

bobcallaway:
QueuedLeafResponse holds the actual insertion response code, so check that instead of just the base GRPC status code.

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/go-openapi/swag"
 
+	"google.golang.org/genproto/googleapis/rpc/code"
 	"google.golang.org/grpc/codes"
 
 	"github.com/projectrekor/rekor/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/api/entries.go
```diff
@@ -91,9 +92,15 @@ func CreateLogEntryHandler(params entries.CreateLogEntryParams) middleware.Respo
...
-	switch resp.status {
-	case codes.OK:
-	case codes.AlreadyExists, codes.FailedPrecondition:
+	//this represents overall GRPC response state (not the results of insertion into the log)
+	if resp.status != codes.OK {
+		return handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("grpc error: %w", resp.err), trillianUnexpectedResult)
+	}
+
+	//this represents the results of inserting the proposed leaf into the log
+	switch resp.getAddResult.QueuedLeaf.Status.Code {
```
dlorenc: tiny nit: I think you can do something like `switch int32(resp.getAddResult.QueuedLeaf.Status.Code)` to save the repeated casting below

dlorenc (approved):
LGTM with one nit. Not sure why the e2e test is failing though...

dlorenc (approved):

PR #82, opened 2021-01-04
Parallelize search when possible

bobcallaway:
The search API allows callers to look for multiple entries in the same request; this change parallelizes the lookups to help speed up processing.

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 
 	"github.com/google/trillian"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/go-openapi/swag"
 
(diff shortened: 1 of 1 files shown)


pkg/api/entries.go
```diff
@@ -251,25 +263,44 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
...
-			case codes.OK, codes.NotFound:
-			default:
-				return handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("grpc error: %w", resp.err), trillianUnexpectedResult)
+		g, _ := errgroup.WithContext(httpReqCtx)
+
+		leaves := make([]*trillian.LogLeaf, len(params.Entry.LogIndexes))
+		for i, logIndex := range params.Entry.LogIndexes {
+			i, logIndex := i, logIndex // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				resp := tc.getLeafByIndex(swag.Int64Value(logIndex))
```
dlorenc, line 272: The trillian client itself (not our TrillianClient, but tc.Client) has a GetLeavesByIndex that takes a slice of indices. Do you think that one would be better here?

bobcallaway: I had looked at that, but discovered that https://github.com/google/trillian/pull/2244 (which has already merged)... so since it appears to be deprecated/removed in the next release, I steered away from it. 

The `GetLeavesByRange` call could be used to optimize this a bit, but I didn't think it was worth doing right now.

dlorenc: Ah, good catch!

dlorenc (approved):

PR #90, opened 2021-01-06
Change the e2e tests to generate new, random data.

dlorenc:
This way we don't have to worry about existing entries in the logs.


bobcallaway (approved):
lgtm

PR #93, opened 2021-01-11
Introduce swagger UI server

lukehinds:


openapi.yaml
```diff
@@ -3,6 +3,9 @@ info:
   title: Rekor
   description: Rekor is a cryptographically secure, immutable transparency log for signed software releases.
   version: 0.0.1
+servers:
```
bobcallaway: I think you'll need to use `host` and `schemes` instead which is the v2 way of expressing this.

https://swagger.io/docs/specification/2-0/api-host-and-base-path/

lukehinds: that fixed it, thanks bob

openapi.yaml
```diff
@@ -4,6 +4,11 @@ info:
   description: Rekor is a cryptographically secure, immutable transparency log for signed software releases.
   version: 0.0.1
 
+host: 104.198.250.40
+basePath: /api/v1
```
bobcallaway: I think you'll either need to leave `basePath` blank or adjust all the URL paths below to omit `/api/v1` from their definition. As written, the URL gets generated with two copies of `/api/v1` in it by SwaggerUI

lukehinds: check that please @bobcallaway https://github.com/projectrekor/rekor/pull/93/commits/cd43323fc17526b28cbe343a38998a9e62f30b85

PR #94, opened 2021-01-11
Add support for ed25519/signify/minisign keys and signatures

bobcallaway:


pkg/pki/ed25519/ed25519.go
```diff
@@ -0,0 +1,189 @@
+/*
+Copyright © 2020 Bob Callaway <bcallawa@redhat.com>
```
lukehinds: tiny weeny nit - 2021

pkg/pki/ed25519/ed25519.go
```diff
@@ -0,0 +1,189 @@
...
+	if key.key == nil {
+		return fmt.Errorf("ed25519 public key has not been initialized")
+	}
+
+	msg, err := ioutil.ReadAll(r)
+	if err != nil {
+		return fmt.Errorf("error reading message to verify signature: %w", err)
+	}
+
+	if ok := ed25519.Verify(ed25519.PublicKey(key.key.PublicKey[:]), msg, s.signature.Signature[:]); !ok {
```
dlorenc: I'm curious what the difference is between using this verify and using https://github.com/jedisct1/go-minisign/blob/master/minisign.go#L93 is.

dlorenc:
Would it make more sense to refer to this package as minisign rather than as ed25519?

My understanding is that the "standard" here we're supporting is really the "minisign/signify" one, which happens to use ed25519 as an implementation detail. If we also add something like x509/pkcs8, it could also support ed25519.

pkg/pki/ed25519/ed25519.go
bobcallaway: The difference is that the link you mention will only validate Minisign signatures. Since signify signatures are a subset of Minisign, the check here validates only the basic signature (not the global signature or a trusted comment).

dlorenc: Ah, makes sense.

dlorenc (approved):

PR #95, opened 2021-01-11
Add x509 support and tests.

dlorenc:
Still need to add some more e2e-style tests that go through the ArtifactFactory flow.


pkg/pki/x509/x509.go
```diff
@@ -0,0 +1,119 @@
+/*
+Copyright © 2020 Bob Callaway <bcallawa@redhat.com>
```
bobcallaway: perhaps change this to 2021 and your name/email?

pkg/pki/x509/x509.go
```diff
@@ -0,0 +1,119 @@
...
+	return s.signature, nil
+}
+
+// Verify implements the pki.Signature interface
+func (s Signature) Verify(r io.Reader, k interface{}) error {
+	if len(s.signature) == 0 {
+		return fmt.Errorf("PGP signature has not been initialized")
+	}
+
+	message, err := ioutil.ReadAll(r)
```
bobcallaway: we could optimize this so that the hash is computed in the same pass over the message, but we can add that in a different PR

pkg/pki/x509/x509.go
```diff
@@ -0,0 +1,119 @@
...
+		if ecdsa.VerifyASN1(pub, hashed[:], s.signature) {
+			return nil
+		}
+		return fmt.Errorf("signature mismatch for %s", string(s.signature))
+	default:
+		return fmt.Errorf("invalid public key type: %T", pub)
+	}
+}
+
+// PublicKey Public Key that follows the PGP standard; supports both armored & binary detached signatures
```
bobcallaway: s/PGP/x509

pkg/pki/x509/x509_test.go
```diff
@@ -0,0 +1,195 @@
+/*
+Copyright © 2020 Bob Callaway <bcallawa@redhat.com>
```
bobcallaway: 2021, your name

pkg/pki/x509/x509.go
```diff
@@ -0,0 +1,119 @@
...
+	return &PublicKey{key: key}, nil
+}
+
+// CanonicalValue implements the pki.PublicKey interface
+func (k PublicKey) CanonicalValue() ([]byte, error) {
+	if k.key == nil {
+		return nil, fmt.Errorf("PGP public key has not been initialized")
+	}
+
+	return x509.MarshalPKIXPublicKey(k.key)
```
bobcallaway: This seems logical, but you couldn't re-read the same object back into NewPublicKey (which assumes PEM encoding)? 

Also, I'd like to see that the `[]byte` returned from `CanonicalValue` could be used to validate the signature in a test case.

pkg/pki/x509/x509.go
```diff
@@ -0,0 +1,119 @@
...
+	signature []byte
+}
+
+// NewSignature creates and validates an x509 signature object
+func NewSignature(r io.Reader) (*Signature, error) {
+	b, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+	return &Signature{
```
bobcallaway, line 43: is there any check that could be done here to validate we have a valid signature object (ahead of the calls failing in Verify?)

pkg/pki/x509/x509.go
dlorenc: Ooh, good point. Fixed and added the test case.

pkg/pki/x509/x509.go
dlorenc: I don't think so... the signatures just look like random bytes.

bobcallaway (approved):

PR #97, opened 2021-01-11
add endpoint for public key and return signed tree head with loginfo

bobcallaway:


cmd/cli/app/log_info.go
```diff
@@ -42,6 +51,47 @@ var logInfoCmd = &cobra.Command{
...
+			log.Fatal(err)
+		}
+		sth := trillian.SignedLogRoot{
+			KeyHint:          keyHint,
+			LogRoot:          logRoot,
+			LogRootSignature: signature,
+		}
+
+		// validate result is signed with the key we're aware of
+		keyResp, err := rekorClient.Tlog.GetPublicKey(nil)
```
dlorenc: This is probably a larger question - but do you think we should store this public key as well? It shouldn't change, really ever.

dlorenc (approved):
LGTM, one comment/question.

cmd/cli/app/log_info.go
bobcallaway: I thought about throwing it into the Rekor config file client side, which would be a nice thing to do IMO.

lukehinds: That sounds like a good approach, this way developers machines can be locked down (to degree), so they only use trusted / nominated public keys.

dlorenc: Right, it's sort of like SSH's TOFU model. I don't think it can be rotated without invalidating the entire t-log, so it should basically never change.

dlorenc: Feel free to leave as a TODO though.

PR #104, opened 2021-01-13
Add a helper to wrap Cobbra functions to make them easier to test.

dlorenc:
This switches out cobra commands to return obj, err. This should make it
easier to test them, and also keep formatting across the commands consistent.

Formatting should happen in one place, instead of having each command print directly.

cc @bobcallaway let me know what you think of this approach before I add it to the other commands.

### DIFF
--- a/cmd/cli/app/get.go
+++ b/cmd/cli/app/get.go
@@ -16,10 +16,14 @@ limitations under the License.
 package app
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 
+	"github.com/projectrekor/rekor/cmd/cli/app/format"
 	"github.com/projectrekor/rekor/pkg/generated/client/entries"
+	"github.com/projectrekor/rekor/pkg/generated/models"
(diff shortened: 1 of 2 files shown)


cmd/cli/app/get.go
```diff
@@ -75,21 +76,36 @@ var getCmd = &cobra.Command{
...
-			return
 		}
 
-		log.Fatal("either --uuid or --log-index must be specified")
-	},
+		return nil, errors.New("either --uuid or --log-index must be specified")
+	}),
+}
+
+func parseEntry(e models.LogEntryAnon) (interface{}, error) {
```
bobcallaway, line 94: These feels fragile to me but barring a cast of `e.Body` to string and doing the base64 decode directly, I can't think of a better way to accomplish this.

bobcallaway (approved):

cmd/cli/app/get.go
dlorenc: Yeah I tried a bunch of ways and got confused because the base64 decoding was happening automatically sometimes (I think []byte serialization takes care of that?), but not others.

PR #108, opened 2021-01-14
re-enable check for missed files in commit in CI workflow

bobcallaway:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
           go-version: ${{ env.GOVERSION }}
       # Download go-swagger
       - name: download go-swagger
-        run : cd $HOME && go get -u github.com/go-swagger/go-swagger/cmd/swagger@master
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           go-version: ${{ env.GOVERSION }}
       # Download go-swagger
       - name: download go-swagger
-        run : go get -u github.com/go-swagger/go-swagger/cmd/swagger@master
+        run : cp go.* $HOME && go get -u github.com/go-swagger/go-swagger/cmd/swagger@master && cp $HOME/go.* .
```
dlorenc, line 31: Is this needed just for the go.mod/go.sum stuff?

bobcallaway: yes

dlorenc: I **think** there's a way to do this with something like `go -mod=readonly` but this is fine for now.

dlorenc (approved):

PR #109, opened 2021-01-14
Add minisign e2e tests.

dlorenc:
These require the minisign binary to be installed.

### DIFF
--- a/tests/util.go
+++ b/tests/util.go
@@ -24,15 +24,23 @@ func outputContains(t *testing.T, output, sub string) {
 	}
 }
 
-func runCli(t *testing.T, arg ...string) string {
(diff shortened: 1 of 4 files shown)


tests/util.go
```diff
@@ -55,6 +63,17 @@ func readFile(t *testing.T, p string) string {
 
 // createdSignedArtifact gets the test dir setup correctly with some random artifacts and keys.
 func createdSignedArtifact(t *testing.T, artifactPath, sigPath string) {
```
bobcallaway: should this really be moved into pgp.go since it calls a PGP-specific signing method?

dlorenc: Whoops, good catch.

bobcallaway (approved):
lgtm

PR #115, opened 2021-01-17
Finish transforming commands into the interface{}, error pattern.

dlorenc:
Next up is a JSON output mode to make testing/scripting easier.

### DIFF
--- a/cmd/cli/app/verify.go
+++ b/cmd/cli/app/verify.go
@@ -17,12 +17,15 @@ package app
 
 import (
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math/bits"
 	"os"
 	"strconv"
 
+	"github.com/google/trillian/merkle"
 	"github.com/google/trillian/merkle/rfc6962"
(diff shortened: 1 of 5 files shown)


cmd/cli/app/verify.go
```diff
@@ -126,16 +150,17 @@ var verifyCmd = &cobra.Command{
 				right = result
 			}
 			result = hasher.HashChildren(left, right)
-			fmt.Printf("SHA256(0x01 | %v | %v) =\n\t%v\n\n", hex.EncodeToString(left), hex.EncodeToString(right), hex.EncodeToString(result))
+			o.Hashes = append(o.Hashes,
```
bobcallaway: do you really want to have this be the output within the JSON structure too, or just have the raw hashes? I might vote for keeping the hashes as what is in the return object and moving the bulk of this code into String().

dlorenc: Good call. Fixed!

bobcallaway (approved):

PR #120, opened 2021-01-20
Fix get with --uuid.

dlorenc:


tests/e2e_test.go
```diff
@@ -104,7 +104,8 @@ func TestGet(t *testing.T) {
 	if err := json.Unmarshal([]byte(out), &g); err != nil {
 		t.Error(err)
 	}
-	// TODO: check the actual data in here.
+	// Get it with the logindex as well
+	runCli(t, "get", "--uuid", g.LogIndex)
```
bobcallaway: `--log-index` should be for LogIndex, `--uuid` should be `uuid`

bobcallaway (approved):

PR #124, opened 2021-01-21
Fix string formatting for the IntegratedTime stamp.

dlorenc:
Also check it in the test.

### DIFF
--- a/cmd/cli/app/get.go
+++ b/cmd/cli/app/get.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/projectrekor/rekor/cmd/cli/app/format"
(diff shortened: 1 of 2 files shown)


cmd/cli/app/get.go
```diff
@@ -38,7 +39,8 @@ type getCmdOutput struct {
 
 func (g *getCmdOutput) String() string {
 	s := fmt.Sprintf("Index: %d\n", g.LogIndex)
-	s += fmt.Sprintf("IntegratedTime: %d\n", g.IntegratedTime)
+	dt := time.Unix(g.IntegratedTime, 0).UTC().Format(time.RFC822)
```
bobcallaway: personally, I would do RFC3339 but either way its better than Unix epochs.

dlorenc: Whoops, I had that but switched it for some reason.

bobcallaway (approved):

PR #127, opened 2021-01-21
use result rather than rootHash to compute inclusion proof

bobcallaway:


dlorenc (approved):
Whoops!

Issue #128, opened 2021-01-22
CLI: Store the log info locally between invocations (the latest signed tree head)


dlorenc:
When loginfo is called, we should store the latest signed tree head locally (in a .rekor directory).

Then, when loginfo is called again, we should perform an inclusion proof between those two signed tree heads.

This can help audit the log and prove it is append-only.

PR #129, opened 2021-01-22
Store the last tree state during loginfo so we can prove the tree state.

dlorenc:
We also need to use this somehow in the log_proof command...


dlorenc:
Ref #128

lukehinds (approved):
lgtm, thanks @dlorenc

PR #135, opened 2021-01-25
Add initial libraries for ssh support.

dlorenc:
I'll rebase this up after some more clean ups.

This doesn't hook anything up yet, but adds some functions and tests for the SSH signature format.

### DIFF
--- a/pkg/pki/ssh/encode.go
+++ b/pkg/pki/ssh/encode.go
@@ -1,0 +1,73 @@
+package ssh
+
+import (
+	"encoding/pem"
+	"errors"
+	"fmt"
+
+	"golang.org/x/crypto/ssh"
+)
+
+const (
+	namespace = "file"
+	pemType   = "SSH SIGNATURE"
+)
+
+func Armor(s *ssh.Signature, p ssh.PublicKey) string {
+	sig := WrappedSig{
+		Version:       1,
+		PublicKey:     string(p.Marshal()),
+		Namespace:     namespace,
+		HashAlgorithm: hashAlgorithm,
+		Signature:     string(ssh.Marshal(s)),
+	}
+	copy(sig.MagicHeader[:], []byte(magicHeader))
+
+	enc := pem.EncodeToMemory(&pem.Block{
+		Type:  pemType,
+		Bytes: ssh.Marshal(sig),
+	})
+	return string(enc)
+}
+
+func Decode(s string) (*ssh.Signature, ssh.PublicKey, error) {
+	pemBlock, _ := pem.Decode([]byte(s))
+	if pemBlock == nil {
+		return nil, nil, errors.New("unable to decode pem file")
+	}
+
+	if pemBlock.Type != pemType {
+		return nil, nil, fmt.Errorf("wrong pem block type: %s. Expected SSH-SIGNATURE", pemBlock.Type)
+	}
+
+	// Now we unmarshal it into the Signature block
+	sig := WrappedSig{}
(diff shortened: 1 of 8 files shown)


pkg/pki/ssh/sign.go
```diff
@@ -0,0 +1,82 @@
...
+	Signature     string
+}
+
+const (
+	magicHeader   = "SSHSIG"
+	hashAlgorithm = "sha512"
+)
+
+func sign(s ssh.AlgorithmSigner, m io.Reader) (*ssh.Signature, error) {
+	data, err := ioutil.ReadAll(m)
```
bobcallaway: I would do an `io.Copy` from `m` to `hf` rather than `ReadAll` and then `Write`.

pkg/pki/ssh/sign.go
```diff
@@ -0,0 +1,82 @@
...
+	dataMessageWrapper = append([]byte(magicHeader), dataMessageWrapper...)
+
+	sig, err := s.SignWithAlgorithm(rand.Reader, dataMessageWrapper, ssh.SigAlgoRSASHA2512)
+	if err != nil {
+		return nil, err
+	}
+	return sig, nil
+}
+
+func Sign(sshPrivateKey string, data []byte) (string, error) {
```
bobcallaway: can you make `data` an `io.Reader` and just pass it through to the `sign()` call?

pkg/pki/ssh/verify.go
```diff
@@ -0,0 +1,35 @@
+package ssh
+
+import (
+	"crypto/sha512"
+
+	"golang.org/x/crypto/ssh"
+)
+
+func Verify(message []byte, armoredSignature string, publicKey []byte) error {
```
bobcallaway: use `io.Reader` instead of `byte[]` in case it is a large payload

pkg/pki/ssh/verify.go
```diff
@@ -0,0 +1,35 @@
...
+		return err
+	}
+
+	desiredPk, _, _, _, err := ssh.ParseAuthorizedKey(publicKey)
+	if err != nil {
+		return err
+	}
+
+	// Hash the message so we can verify it against the signature.
+	h := sha512.New()
```
bobcallaway, line 22: `io.Copy` from reader straight to writer `&h`

pkg/pki/ssh/encode.go
```diff
@@ -0,0 +1,68 @@
...
+		Version:       1,
+		PublicKey:     string(p.Marshal()),
+		Namespace:     "file",
+		HashAlgorithm: hashAlgorithm,
+		Signature:     string(ssh.Marshal(s)),
+	}
+	copy(sig.MagicHeader[:], []byte(magicHeader))
+
+	enc := pem.EncodeToMemory(&pem.Block{
+		Type:  "SSH SIGNATURE",
```
bobcallaway: make this into `const`?

pkg/pki/ssh/sign.go
dlorenc: Nice! done.

pkg/pki/ssh/verify.go
dlorenc: Done.

pkg/pki/ssh/verify.go
dlorenc: Done!

bobcallaway (approved):
lgtm

dlorenc:
I'm going to squash this and then merge if tests still pass!

PR #136, opened 2021-01-27
Add a production logging configuration and a flag.

dlorenc:
This is only toggled on for the server right now.


lukehinds (approved):
nice change!

dlorenc:
The JSON format looks much nicer in the consoles :)

PR #139, opened 2021-01-28
block to return for insertion until leaf has been added

bobcallaway:


pkg/api/trillian_client.go
```diff
@@ -76,6 +77,52 @@ func (t *TrillianClient) addLeaf(byteValue []byte) *Response {
...
+
+	root, err := t.root()
+	if err != nil {
+		return &Response{
+			status:       status.Code(err),
+			err:          err,
+			getAddResult: resp,
+		}
+	}
+	verifier, err := client.NewLogVerifierFromTree(t.tree)
```
dlorenc: nit: is there a way we could create this once and cache it on the t object? It doesn't appear to rely on anything in the request context.

pkg/api/trillian_client.go
```diff
@@ -76,6 +77,52 @@ func (t *TrillianClient) addLeaf(byteValue []byte) *Response {
...
+	if err != nil {
+		return &Response{
+			status:       status.Code(err),
+			err:          err,
+			getAddResult: resp,
+		}
+	}
+	logClient := client.New(t.logID, t.client, verifier, root)
+	if err := logClient.WaitForInclusion(t.context, byteValue); err != nil {
```
dlorenc, line 106: Is there a way to set a timeout here? We don't have a way to expose metrics yet, but this time span is exactly what I would want to monitor eventually too.

dlorenc (commented):
Nice! Two nits/questions.

dlorenc:
Oh, and you can also remove the time.Sleeps from our e2e tests, that will actually verify this code in a way.

pkg/api/trillian_client.go
bobcallaway: yup, will fix

pkg/api/trillian_client.go
bobcallaway: no, it will be min({HTTP request timeout (as set by client) which will cause context to cancel},{10 seconds}). https://github.com/google/trillian/blob/549baa4214f1fd64b89344d55dc9aa894ab80c27/client/log_client.go#L130 uses an backoff approach here. I suspect that if we add metrics we can just wrap the call and be good with it.

dlorenc:
Nice!

dlorenc (approved):

PR #140, opened 2021-01-28
Add and plumb through the ssh type.

dlorenc:


pkg/pki/ssh/ssh.go
```diff
@@ -0,0 +1,107 @@
...
+	return []byte(Armor(s.signature, s.pk)), nil
+}
+
+// Verify implements the pki.Signature interface
+func (s Signature) Verify(r io.Reader, k interface{}) error {
+	if s.signature == nil {
+		return fmt.Errorf("ssh signature has not been initialized")
+	}
+
+	message, err := ioutil.ReadAll(r)
```
bobcallaway: no need to do this, just pass through r to Verify

pkg/pki/ssh/ssh.go
```diff
@@ -0,0 +1,107 @@
...
+	pk        ssh.PublicKey
+}
+
+// NewSignature creates and Validates an ssh signature object
+func NewSignature(r io.Reader) (*Signature, error) {
+	b, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+	sig, pk, err := Decode(string(b))
```
bobcallaway: nit: why do `Armor` and `Decode` return/take string arguments when it appears the first thing you do is convert back to `[]byte`?

bobcallaway: also, should Decode just return the `Signature` struct as defined on line 28?

pkg/pki/ssh/ssh.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, err
+	}
+
+	return &PublicKey{key: key}, nil
+}
+
+// CanonicalValue implements the pki.PublicKey interface
+func (k PublicKey) CanonicalValue() ([]byte, error) {
+	if k.key == nil {
+		return nil, fmt.Errorf("x509 public key has not been initialized")
```
bobcallaway: s/x509/ssh

pkg/pki/ssh/ssh.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return &Signature{
+		signature: sig,
+		pk:        pk,
+	}, nil
+}
+
+// CanonicalValue implements the pki.Signature interface
+func (s Signature) CanonicalValue() ([]byte, error) {
+	return []byte(Armor(s.signature, s.pk)), nil
```
bobcallaway, line 47: from above - change return type of Armor to []byte?

bobcallaway (approved):
lgtm, good work!

lukehinds (approved):
+1

PR #143, opened 2021-02-01
persist URL along with state for comparison

### LINKED ISSUE #142: incorrect CLI persisted log state error
There are scenarios for the CLI where the server queried may not be the same as where the log state information that persisted in the `~/.rekor/` directory was obtained from. If this is the case, the CLI prints an incorrect error state in comparing states.

bobcallaway:

fixes #142 

Also squelches `Using config file: ...` message if the format requested != default

### DIFF
--- a/cmd/cli/app/state/state.go
+++ b/cmd/cli/app/state/state.go
@@ -26,10 +26,7 @@ import (
 	"github.com/mitchellh/go-homedir"
 )
 
-type persistedLogRoot struct {
-	ServerURL string
-	LogRoot   *types.LogRootV1
-}
+type persistedState map[string]*types.LogRootV1
 
 func Dump(url string, lr *types.LogRootV1) error {
 	rekorDir, err := getRekorDir()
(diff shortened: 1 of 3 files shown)


dlorenc:
Nice!

cmd/cli/app/state/state.go
```diff
@@ -26,14 +26,24 @@ import (
 	"github.com/mitchellh/go-homedir"
 )
 
-func Dump(lr *types.LogRootV1) error {
+type persistedLogRoot struct {
```
dlorenc: Would it make sense to make this a map of url -> logroot? instead? So that way we can track multiple at the same time?

lukehinds (approved):
good idea!

cmd/cli/app/state/state.go
bobcallaway: well if we're going to do that, should we do `map[url]map[treeSize]*types.LogRootV1` so we can save not just the latest queried state from that server?

dlorenc: IMO we probably don't want to store the entire tree state locally for every user - but we do need some way to do that for people that want to run full mirrors.

dlorenc (approved):

PR #148, opened 2021-02-04
use variable instead of hardcoded DB name

bobcallaway:


lukehinds:
This reminds me that we should likely stop carrying storage.sql and instead point users upstream to trillian, as they might make changes to schema and we fall out of sync.

lukehinds (approved):

PR #150, opened 2021-02-06
Codeql analysis

naveensrinivasan:
https://codeql.github.com/docs/

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,36 @@
+# https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
+name: "CodeQL"
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '45 10 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,36 @@
+# https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
+name: "CodeQL"
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
```
bobcallaway, line 9: im not sure we need a schedule if this will run on all PRs targeted at main.

bobcallaway (commented):
thanks for adding this; I'm guessing there is no analysis because this hasn't run properly on main yet?

.github/workflows/codeql-analysis.yml
naveensrinivasan: Even I had the same concern initially, then I realized there new upstream vulnerabilities identified(codeql identifies a new way of catching code issues) that will not be reported till there is a new PR. For example, could be a zero-day, especially with a security-critical project it would make sense to run them on a schedule to mitigate that risk.

naveensrinivasan:
> thanks for adding this; I'm guessing there is no analysis because this hasn't run properly on main yet?

Yes because it hasn't yet run on main.

.github/workflows/codeql-analysis.yml
bobcallaway: makes sense. Thanks!

PR #152, opened 2021-02-07
Fix broken link pointing to rekord_schema

bistaastha:
This is a tiny change to fix the broken link which pointed to `rekord_schema.md` and not `rekord_schema.json`.


lukehinds:
nice catch, thanks @bistaastha

Issue #171, opened 2021-02-20
Rename x509 PKI to PKIX


dlorenc:
I think this is probably closer to the "correct" name the way we're using these keys, since we don't actually verify a certificate chain.

We could either extend this type to take an optional cert chain to verify the public key against, or have two types of PKI here.

dlorenc:
I guess the cert chains could be stored in ExtraData, since we could theoretically have multiple valid cert chains for a given signature.

dlorenc:
Or, one other idea: allow for the public key to refer to a certificate (with an embedded public key) instead of just a public key.

We would then verify the certificate itself and store that in extra data (because many certificates can contain the same public key), extract the public key and store that as usual.

PR #172, opened 2021-02-20
Allow uploading x509 signatures based on a cert rather than the publi…

dlorenc:
…c key.

See #171 for more context.


bobcallaway (approved):
lgtm

PR #175, opened 2021-02-25
Refactor of org name

lukehinds:
All instances of projectrekor are now renamed to SigStore

This includes:

* Import paths
* Tests
* Readme's

Signed-off-by: Luke Hinds <lhinds@redhat.com>


lukehinds:
@bobcallaway for changes to swagger pkg/generated files, if you prefer to generate these using the tool, I can restore these files, but I figured that won't be needed (but hey, always prudent to check these sorts of things).

dlorenc (approved):
The renaming works for me as long as it passes ci!

PR #177, opened 2021-02-26
changed directory permissions for .rekor directory

jehanshah8:
Changed the permissions for .rekor directory because they were too restrictive to allow new writes

### DIFF
--- a/cmd/cli/app/state/state.go
+++ b/cmd/cli/app/state/state.go
@@ -82,7 +82,7 @@ func getRekorDir() (string, error) {
 	}
 	rekorDir := filepath.Join(home, ".rekor")
 	if _, err := os.Stat(rekorDir); os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


dlorenc:
Nice! This was causing issues saving state for fresh installs.

lukehinds:
750 would get this past gate.

cmd/cli/app/state/state.go
```diff
@@ -82,7 +82,7 @@ func getRekorDir() (string, error) {
 	}
 	rekorDir := filepath.Join(home, ".rekor")
 	if _, err := os.Stat(rekorDir); os.IsNotExist(err) {
-		if err := os.MkdirAll(rekorDir, 0644); err != nil {
+		if err := os.MkdirAll(rekorDir, 0755); err != nil {
```
dlorenc:
```suggestion
		if err := os.MkdirAll(rekorDir, 0750); err != nil {
```

PR #181, opened 2021-03-03
no longer require SHA to upload artifacts to log

### LINKED ISSUE #180: computation of hash when fetching remote artifact is redundant
When using the CLI to create an entry in the log for a remote entry, we currently require the caller to specify the SHA256 hash as well as provide the digital signature. Since verification of the signature will by definition include validation of the hash, we can remove this out of the Rekor processing flow.

@SantiagoTorres

bobcallaway:
Since the verification of a signature will, by definition, include
verifying the content has not been altered, it is unnecessary to require
users of the CLI or REST API to specify the SHA256 hash of the content
when creating a new entry into the log.

Note that the server will still compute the hash and store it in the log
for ease of comparison.

Fixes #180


dlorenc:
Nice!

PR #184, opened 2021-03-03
fix race condition in e2e tests

### LINKED ISSUE #183: E2E tests can fail prematurely because docker-compose services are not healthy before the test run starts
In our CI step for E2E tests, the CLI is rebuilt from scratch in a new GitHub environment; this means that the go module dependencies must be fetched. The delay introduced during this fetch masks a race-condition in the`e2e-test.sh` script in that the tests can start running before the rekor services and dependencies have come up (this is usually because we are creating a MySQL database for the first time which requires some time to initialize).

bobcallaway:
In our CI environment there is an artifical delay in between starting
the Rekor services via docker-compose and when the E2E tests are
actually executed due to Go modules being downloaded. In a local
development environment, the download may not be required so the tests
can start before the docker-compose services are actually running.

This introduces a healthcheck for services (where possible), and blocks
the start of the e2e tests until the services are reporting as healthy.
It also forces the use of an empty homedir and rekor config file to
ensure no collision between the tests and the developer's environment.

Fixes #183

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,12 @@ services:
       - MYSQL_USER=test
       - MYSQL_PASSWORD=zaphod
     restart: always # keep the MySQL server running
+    healthcheck:
+      test: ["CMD", "/etc/init.d/mysql", "status"]
+      interval: 30s
+      timeout: 3s
+      retries: 3
+      start_period: 10s
   redis-server:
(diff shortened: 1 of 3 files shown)


lukehinds (approved):
LGTM and loving the descriptive commit msg's, it really helps give context to the PR.

dlorenc (approved):

PR #189, opened 2021-03-05
Add the UUID to the outpupt for the get command.

dlorenc:
This lets you retrieve entries by UUID or index, and see the other
value.


lukehinds (approved):
good idea!

PR #193, opened 2021-03-08
Add timestamps to the log_info output.

dlorenc:
Also clean up a few small panics along the way.


lukehinds (approved):
lgtm, just a small conflict to fix

Issue #191, opened 2021-03-07
pubsub/bigquery


dlorenc:
We could make some other types of (UNTRUSTED) feeds available to simplify integrations.

* All entries could get sent to a public pubsub/kafka topic for others to use.
* All entries could get synced to a public bigquery instance for ad-hoc querying/analysis.
* Publish daily/hourly (signed) snapshots and backups of the log to some kind of CDN.

Hayden-IO:
We chatted about this issue today. As more clients are interested in monitoring Rekor, we'll need to figure out how to scale reads for Rekor. One option is read replicas. Pub/sub is another good option.

cc @flxw

Hayden-IO:
This is complete.

PR #197, opened 2021-03-09
Add a second job that watches rekor for new STH entries and publishes

dlorenc:
them to GCS.

We don't always have a new one, and under load there might be "batches" of entries
all integrated into the same STH. This means there is no guaranteed frequency of updates
or even a guarantee that every index will exist.

The values (and timestamps) should be monotonically increasing though.


config/watcher.yaml
```diff
@@ -0,0 +1,40 @@
...
+      containers:
+      - name: rekor-watcher
+        image: ko://github.com/sigstore/rekor/cmd/server
+        command: ["/ko-app/server"]
+        ports:
+        - containerPort: 3000
+        - containerPort: 2112 # metrics
+        args: [
+          "watch",
+          "--rekor_server.address=rekor-server",
```
lukehinds, line 30: Does that not need to be rekor-watcher (the binding hostname)?

dlorenc: Nope! This is how the watcher finds the server. I reused the same flag for consistency, but could also see how that's more confusing...

cmd/server/app/watch.go
```diff
@@ -0,0 +1,182 @@
...
+			if err := uploadToGcs(ctx, bucket, lr); err != nil {
+				log.Logger.Warnf("error uploading result: %s", err)
+				continue
+			}
+			last = lr
+		}
+	},
+}
+
+func init() {
```
font, line 134: There's a duplicate `init` at the bottom of this file.

font (commented):
@dlorenc this is cool! So now we can add others (likely as sub-commands to `watch`) to support https://github.com/sigstore/rekor/issues/191.

font (commented):
@dlorenc Is there any way to add an e2e test so that when we implement others from https://github.com/sigstore/rekor/issues/191 we know we won't have broken GCS?

cmd/server/app/watch.go
dlorenc: Thanks!

dlorenc:
> @dlorenc Is there any way to add an e2e test so that when we implement others from #191 we know we won't have broken GCS?

Yes, really good idea. I need to think through the right layer to mock out GCS at...

dlorenc:
> Yes, really good idea. I need to think through the right layer to mock out GCS at...

All set!

tests/e2e_test.go
```diff
@@ -283,3 +286,28 @@ func TestUploadNoAPIKeyInOutput(t *testing.T) {
 		t.Errorf("CLI output contained API key when it should have squelched it")
 	}
 }
+
+func TestWatch(t *testing.T) {
+
+	td := t.TempDir()
+	cmd := exec.Command("../rekor-server", "watch", "--interval=1s")
```
font: `s|"../rekor-server"|server|`

cmd/server/app/watch.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		pub, err := x509.ParsePKIXPublicKey(block.Bytes)
+		if err != nil {
+			return err
+		}
+
+		verifier := tclient.NewLogVerifier(hasher.DefaultHasher, pub, crypto.SHA256)
+		ctx := context.Background()
+		bucketURL := os.Getenv("REKOR_STH_BUCKET")
```
font: The next function call will catch this, but we may want to consider fatally logging a specific error message if this env var is unset.

font (approved):
@dlorenc Nice!

cmd/server/app/watch.go
dlorenc: Done!

tests/e2e_test.go
dlorenc: I don't think so... We compile it as "rekor-server" above, and the test directory here is set to test/, so we need to go back up one.

font: Oh sorry I should have been more specific! The string replacement was to use the `server` constant var defined below as I assumed that's what you intended and I didn't see the constant used anywhere else.

font (approved):

font (approved):
LGTM. Can we merge this?

PR #221, opened 2021-03-18
Remove gzip processing flow completely from rekor

### LINKED ISSUE #208: Failed to upload an gzipped artifact
Following https://sigstore.dev/get_started/client/, I have the following output
```
$ gpg --armor -u jdoe@example.com --output mysignature.asc --detach-sig myrelease.tar.gz
$ rekor upload --rekor_server http://localhost:3000 --signature mysignature.asc --public-key mypublickey.key --artifact "http://localhost:8000/myrelease.tar.gz"
2021/03/12 07:41:27 error retrieving external entities: openpgp: invalid signature: hash tag doesn't match
```

The reason why `hash tag doesn't match` is that the signature is for the payload `myrelease.tar.gz` not for its decompressed content. However, the rekor cli decompress the content automatically and then do the verification.

bobcallaway:
Issue #208 reported different handling of gzipped content via fetch vs
direct upload to rekor server. The code should be consistent, regardless
of whether content was compressed or not - by always attempting to
verify the signature against the (unmodified) byte stream.

This patch removes the gzip decoding completely from rekor and verifies
the bytes supplied or referenced.

Also fixes issue in E2E tests where sending SIGKILL to watch process
caused message to be printed to stderr, which fails the test when
running on MacOS.

Fixes #208

### DIFF
--- a/pkg/api/index.go
+++ b/pkg/api/index.go
@@ -53,7 +53,7 @@ func SearchIndexHandler(params index.SearchIndexParams) middleware.Responder {
 	}
 	if params.Query.PublicKey != nil {
 		af := pki.NewArtifactFactory(swag.StringValue(params.Query.PublicKey.Format))
-		keyReader, err := util.FileOrURLReadCloser(httpReqCtx, params.Query.PublicKey.URL.String(), params.Query.PublicKey.Content, true)
+		keyReader, err := util.FileOrURLReadCloser(httpReqCtx, params.Query.PublicKey.URL.String(), params.Query.PublicKey.Content)
 		if err != nil {
 			return handleRekorAPIError(params, http.StatusBadRequest, err, malformedPublicKey)
 		}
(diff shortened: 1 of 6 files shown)


dlorenc:
I'm fine removing this and I agree with it, but the reason we added it was because linux kernel releases are signed before zipping. They're zipped in a few ways, but the only the tar is signed.

bobcallaway:
@dlorenc is that worth adding in some retry logic such that if we detect a gzip content type we retry if the SHA or signature validation fails with the compressed content, we try again with the uncompressed content?

dlorenc:
> @dlorenc is that worth adding in some retry logic such that if we detect a gzip content type we retry if the SHA or signature validation fails with the compressed content, we try again with the uncompressed content?

IMO  the server-side simplification here is probably the best long-term thing, and we can handle strangely signed content like the kernel in client-side tooling. I don't really know how common it is though, so we can always revisit if it turns out to be really common.

dlorenc (approved):

dlorenc:
LGTM. Merge when ready!

PR #226, opened 2021-03-19
Return location of existing entry if upload tried

### LINKED ISSUE #222: rekor CLI should print entry URL even in case the entry already exists
Currently it just prints `Entry already exists.`

bobcallaway:
Adds a Location response header when a 409 Conflict error is returned
from the server when a duplicate entry is sent for insertion into the
log. Also changes message printed by CLI to improve usability.

Fixes #222

### DIFF
--- a/pkg/api/error.go
+++ b/pkg/api/error.go
@@ -21,6 +21,7 @@ import (
 	"regexp"
 
 	"github.com/go-openapi/runtime/middleware"
+	"github.com/go-openapi/strfmt"
 	"github.com/mitchellh/mapstructure"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/generated/restapi/operations/entries"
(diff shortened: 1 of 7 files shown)


pkg/api/error.go
```diff
@@ -98,7 +99,19 @@ func handleRekorAPIError(params interface{}, code int, err error, message string
...
 		case http.StatusConflict:
-			return entries.NewCreateLogEntryConflict().WithPayload(errorMsg(message, code))
+			resp := entries.NewCreateLogEntryConflict().WithPayload(errorMsg(message, code))
+			locationFound := false
+			for _, field := range fields {
+				if locationFound {
+					existingURL := field.(strfmt.URI)
+					resp.SetLocation(existingURL)
+					break
+				} else if field.(string) == "entryURL" {
```
dlorenc, line 109: Does this require the fields to be in the right order? Would it be safer to do two passes? one for the url then another for the URI?

lukehinds:
seems like the `u.Location` is missing still (not had a chance to debug, just ran a local test)

image

lukehinds:
woops (hit wrong button, did not meant to close)

bobcallaway:
> seems like the `u.Location` is missing still (not had a chance to debug, just ran a local test)
> 
> image

the patch isn't live against api.rekor.dev so you'll need to try against localhost

lukehinds:
> > seems like the `u.Location` is missing still (not had a chance to debug, just ran a local test)
> > image
> 
> the patch isn't live against api.rekor.dev so you'll need to try against localhost

just realised that (I need more coffee)

pkg/api/error.go
bobcallaway: reviewed with @dlorenc

dlorenc (approved):

PR #246, opened 2021-03-31
rekor-cli: logproof -last-size default fix

ahmetb:
-last-size is defaulting to 1, but fixing it to 0 so we can properly error out
if it's unspecified.

### DIFF
--- a/cmd/rekor-cli/app/log_proof.go
+++ b/cmd/rekor-cli/app/log_proof.go
@@ -55,14 +55,17 @@ var logProofCmd = &cobra.Command{
 		if err := viper.BindPFlags(cmd.Flags()); err != nil {
 			return fmt.Errorf("Error initializing cmd line args: %s", err)
 		}
-		if viper.GetUint64("first-size") > viper.GetUint64("last-size") {
-			return errors.New("last-size must be >= to first-size")
-		}
 		if viper.GetUint64("first-size") == 0 {
 			return errors.New("first-size must be > 0")
 		}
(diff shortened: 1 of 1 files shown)


bobcallaway:
Thanks for the PR!

I think I understand what you're going after here... but we already have checks to see if the flag was specified or not on https://github.com/sigstore/rekor/blob/b5e49b4dd80aa37a9164da07d84015c2d2d0b574/cmd/rekor-cli/app/log_proof.go#L98 

Can you clarify what args you are passing to `./rekor-cli` and the expected output?

ahmetb:
We do already have checks, but if we leave default as `1`, it shows `(default: 1)` next to the flag. Is there a particular reason why it is set to 1?

bobcallaway:
I didn't like setting it to 0 as a default knowing that it was an invalid value... but I now see that this has the added benefit of removing it from the help message. 

I tried to push a small addition to your commit to help add and reorder the flag checks so they're a bit more logical. Any chance you could turn on the "allow edits from maintainers" option in the PR so I can amend this?

Thanks for pointing this out!

ahmetb:
> allow edits from maintainers

is already on. 🤔

bobcallaway:
> > allow edits from maintainers
> 
> is already on. 🤔

My bad. Helps if you push to the correct branch :(

dlorenc:
Is this ready? Did you have anything else to push still @bobcallaway ?

ahmetb:
I think it's ready to go.

PR #250, opened 2021-04-02
Update rekor REST API to match Trillian semantics

### LINKED ISSUE #229: Update Rekor API after get by UUID removal
Looks like the GetLeavesByHash call got removed somewhere, our CI running against the trillian_log_server at head broke. I pinned it, but we'll need to fix that before we can update

bobcallaway:
This patch removes the `/api/v1/log/entries/{uuid}/proof` endpoint. If you have the UUID (aka the leaf Merkle hash), you likely want proof that the content represented by that hash is included in the log. There's no need for a separate `/proof` endpoint to deliver the same content.

This commit also ensures that the `getLogEntryByIndex` and `getLogEntryByUUID` endpoints return an inclusion proof as part of their response content. The search endpoint also now returns the inclusion proof of all entries returned from the query.

With this patch, Rekor no longer uses the deprecated `GetLeavesByHash` Trillian API.

Fixes #229


dlorenc (approved):
This looks good, I wonder how we should make the switch over in our clients.

bobcallaway:
Only thing that bugs me (slightly) is that including the inclusion proof makes the content uncacheable, since the proof will change as the log grows; that’s a minor nit. 

With regards to the client changes - I think since we're still making promises about being pre-production we could just roll out the changes. I know cosign could have some code cleaned up and probably sigstore as well. 

I will rebase this and if we're good with that approach we can merge.

dlorenc:
> this

Yup, LGTM

PR #259, opened 2021-04-08
Serve a fulcio signed certificate of the rekor public key

### LINKED ISSUE #256: Serve a fulcio signed certificate of the rekor public key
In cosign we're currently working on bundling all necessary proofs into the signature (https://github.com/sigstore/cosign/issues/181) -- the idea is that we can verify the signature is in rekor without having to actually access the rekor tlog.

Ideally, we would like cosign to verify these signatures with as few public keys required as possible. Instead of having to store rekor's public key in the code, we could:
1. Create a fulcio signed cert that contains the rekor public key
2. Store that cert in the cosign signature

That way we can get rekor's public key, but only have to trust the fulcio root CA to get it from the cert.

I was planning to create the cert, and then making it accessible from the rekor API in the same way the rekor public key is currently accessible.

priyawadhwa:
I hardcoded the cert for now, but if anyone knows of a more correct place to put it that would probably be better!

Fixes #256


dlorenc:
What do you think about putting this in a file (then plumbing it through via kodata or a configmap) and having it available as a flag at start time?

lukehinds:
A separate file is going to be useful for local development environments

priyawadhwa:
Opted to store the public key in cosign instead, so I'm gonna close this!

Issue #257, opened 2021-04-07
Get signed proof from rekor for bundled cosign signatures


asraa:
**Description**
This is for getting bundling (https://github.com/sigstore/cosign/issues/181) working.

When cosign uploads the {signature, public key, payload} to rekor, we need rekor to provide proof that the entry is in the tlog. 
1. We'll have Fulcio serve a cert for Rekor https://github.com/sigstore/rekor/issues/256 to single out Fulcio as a trust source.
2. We need to get Rekor to provide a signed proof (with a separate key) that the entry is in the log. 

For step (2): We would (a) create a KMS key for Rekor and publish the key somewhere and (b) add an endpoint to get SignedProofs for an entry. The signed proof would contain {payload (timestamp that Rekor got the proof + inclusion proof), signature}. 

We can probably just add the SignedProof to the entry and create it when we create the log entry...

When cosign retrieves that, it can bundle the SignedLogRoot and SignedProof into the signature object so that it can (1) Verify that fulcio attests to rekor's public key (2)  Verify the signature on the proof and STH (3) Use the proof and the STH to verify that the entry was in the tlog. 

cc @dlorenc @priyawadhwa

-------------------------------------------------
About promises: it seems like Rekor already waits for the entry to be included https://github.com/sigstore/rekor/blob/e41105480a8638741ca169907c776db8ac2145ba/pkg/api/trillian_client.go#L107 so we don't necessarily need promises over the inclusion proof. If we did want the promise, ... sort of unsure how to make one nicely

dlorenc:
I was looking for some inspiration on what to do with the keys (since we now have two). Unfortunately the ctclient stuff doesn't handle this today: https://tools.ietf.org/html/rfc6962

> Note that this document does not describe how clients obtain the logs' public keys.

There's some more info here:
https://github.com/google/certificate-transparency-community-site/blob/master/docs/google/known-logs.md

dlorenc:
I can't find any logs that use two separate keys. It sounds like we might be able to use the same one?

https://github.com/google/certificate-transparency-go/blob/master/trillian/docs/ManualDeployment.md#key-generation

lukehinds:
Will rekor still be able operate independently of a fulcio instance being present, e.g. this proposal is implemented in a modular way so the operator can run either with this new mode of operation or the current mode?

dlorenc:
Yeah, this portion is really about deciding whether we have one key or two i think, separate from fulcio.

asraa:
[...]

PR #260, opened 2021-04-09
Add signed tree head to inclusion proof

asraa:
Part of https://github.com/sigstore/rekor/issues/257

Adds a signed tree head to an inclusion proof so that a user can use the InclusionProof to verify that an entry exists in the transparency log as long as they trust Rekor's key to trust the root hash.

I implemented this PR off of https://github.com/sigstore/rekor/pull/250 to avoid re-using `GetLeavesByHash`


openapi.yaml
```diff
@@ -409,11 +390,32 @@ definitions:
           type: string
           description: SHA256 hash value expressed in hexadecimal format
           pattern: '^[0-9a-fA-F]{64}$'
+      signedTreeHead:
```
dlorenc, line 396: Can we reuse the signedTreeHead definition from line 329?

bobcallaway (approved):
LGTM!

asraa:
Thanks! I'll merge in main but just FYI I'm looking into replacing this with a "signed promise" like SCT for ctlog that can be used to replace the need for copying the whole inclusion proof when verifying without calling rekor.
@priyawadhwa

Edit: Also in flux after the recent trillian change.

asraa:
Closed with https://github.com/sigstore/rekor/pull/291

Issue #262, opened 2021-04-12
Trillian is removing the log signing


dlorenc:
We need to update our trillian clients and production instances to deal with the removal of log signing. This has a few implications:

* It makes the API changes @priyawadhwa and @asraa were thinking about much more straightforward
* It simplifies our key management story
* It potentially means we need to change our private key. Can we use KMS?
* We need to drop all of our signature verification, the client libraries are rapidly changing. 

https://github.com/google/trillian/issues/2450

dlorenc:
I don't think this will matter to the CT log. We can update that too though easily.

lukehinds:
Good catch @dlorenc , I have read the original trillian issue and its very parse on details. Maybe we can chat about rekor changes required on the community meeting.

dlorenc:
I think the breakage is 100 percent our fault for not pinning the tag on the signer, but the details are light on the original pr.

asraa:
Just to talk it out yet again :P This means rekor should now be signing its own tree heads to publish on the log and can then be signing promises to include? 

BTW cosign never verified the signature on SignedLogRoot (https://github.com/sigstore/cosign/blob/e9f5113e2a4a8cda2e6fa2ea561b12f9049fa4cf/pkg/cosign/verify.go#L138). Just want to confirm my mental model that the onus on verifying public logs is on monitoring instances

dlorenc:
That's my understanding. I think we'll want to wait a few days until the trillian client changes settle and get more clear, then come up with a plan to tackle all the changes at once. Does that make sense?

priyawadhwa:
Sounds good! +1 to changing our private key & using KMS instead

PR #263, opened 2021-04-12
Drop signature verification in rekor server.

dlorenc:
This is changing as part of the general trillian signature changes.
The trust model is still client -> database, our server trusts our
database so we can pass signed messages on directly to users without
double verification.


dlorenc:
Ref #262

priyawadhwa (approved):

PR #264, opened 2021-04-13
[api] Add max 32 mb on content data

### LINKED ISSUE #220: Bug: Request entity too large
I hit this when trying to upload a 3.7M artifact using the --artifact flag. It worked for me when passing in a URL. I think a reasonable limit makes sense here, but 3.7M seems like it should work. Not sure where to fix this.

asraa:

Fixes https://github.com/sigstore/rekor/issues/220
Adds a max of 32 MB to content data to avoid hitting ingress max size.

Not sure I particularly like this since the error is sorta opaque, but at the same time it's defined in the schema. Otherwise, I can add to `Validate` for each type and return a more verbose error message on expected/actual size.

### DIFF
--- a/pkg/types/rekord/v0.0.1/rekord_v0_0_1_schema.json
+++ b/pkg/types/rekord/v0.0.1/rekord_v0_0_1_schema.json
@@ -86,7 +86,8 @@
                 "content": {
                     "description": "Specifies the content inline within the document",
                     "type": "string",
-                    "format": "byte"
+                    "format": "byte",
+                    "maxLength" : 32000000
                 }
             },
             "oneOf": [
--- a/pkg/generated/models/rekord_v001_schema.go
+++ b/pkg/generated/models/rekord_v001_schema.go
@@ -175,6 +175,7 @@ func (m *RekordV001Schema) UnmarshalBinary(b []byte) error {
 type RekordV001SchemaData struct {
 
 	// Specifies the content inline within the document
+	// Max Length: 32000000
 	// Format: byte
 	Content strfmt.Base64 `json:"content,omitempty"`
 
@@ -190,6 +191,10 @@ type RekordV001SchemaData struct {
 func (m *RekordV001SchemaData) Validate(formats strfmt.Registry) error {
 	var res []error
 
+	if err := m.validateContent(formats); err != nil {
+		res = append(res, err)
+	}
+
 	if err := m.validateHash(formats); err != nil {
 		res = append(res, err)
 	}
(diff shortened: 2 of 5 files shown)


pkg/types/rekord/v0.0.1/rekord_v0_0_1_schema.json
```diff
@@ -86,7 +86,8 @@
                 "content": {
                     "description": "Specifies the content inline within the document",
                     "type": "string",
-                    "format": "byte"
+                    "format": "byte",
+                    "maxLength" : 32000000
```
dlorenc, line 90: I guess we can't comment here in the JSON. I'll make a note in the nginx config to remember to update both places.

dlorenc (approved):

pkg/types/rekord/v0.0.1/rekord_v0_0_1_schema.json
dlorenc: OK, filed a TODO over there.

bobcallaway: While this change would limit the length of the inline base64 encoded content field, it's not sufficient to develop an equivalent limit to what the NGINX ingress was enforcing (i.e. the entire number of bytes passed in the HTTP request body). Also since OpenAPI is enforcing this via the length of the string, the 32M value really puts a limit on the content of ~67% of that or approx 21MB. 

I can see use cases for passing large files to Rekor (e.g. signed ISO images or jumbo containers), and since certain types can be fully streamed (rekord and RPM) while others can't (soon-to-be-added JAR support), I think we need to  revert this PR and consider it more broadly. I'm happy to take some time to develop a proposal on this. 

I could also imagine some users might want a size limit to be configurable, so that is something to consider as well.

lukehinds: > I could also imagine some users might want a size limit to be configurable, so that is something to consider as well.

Agree, this could be a config value in rekor-server.yaml, that way you can place `32000000` as the default value in the config, but users of rekor who have needs for larger file can still set accordingly.

asraa: That sounds great, would love to chat more about office hours! I had exactly that problem with not being able to measure the request itself, and couldn't see a great solution here.

dlorenc: The issue right now is that it's enforced at ingress, not in our rekor config. That's the main reason client-validation would be nice, otherwise we end up with random errors from the load balancer that clients can't catch.

It might be better to just let clients hit it though.

asraa: We could also return back the 413 w/ error message

PR #267, opened 2021-04-13
Add support for signing and retrieving public key from KMS

priyawadhwa:
This PR does a few different things:
- Adds a `Signer` interface which can be used to sign payloads and get public keys (currently supports only GCP KMS)
- Moves the API for getting the public key out of the tlog
- Removes public key from trillian client

I ran this locally and was able to successfully get the public key via the API. 

Up next: sign root hashes using the Signer interface


pkg/api/public_key.go
```diff
@@ -0,0 +1,39 @@
...
+func GetPublicKeyHandler(params pubkey.GetPublicKeyParams) middleware.Responder {
+	b, err := x509.MarshalPKIXPublicKey(api.pubkey)
+	if err != nil {
+		handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("marshal public key err: %w", err), marshalPublicKeyError)
+	}
+	key := pem.EncodeToMemory(&pem.Block{
+		Type:  "PUBLIC KEY",
+		Bytes: b,
+	})
+	return pubkey.NewGetPublicKeyOK().WithPayload(string(key))
```
dlorenc: I think this is the part @bobcallaway mentioned today - to use something related to the OIDC spec. I'm not sure I understood it correctly.

pkg/signer/signer.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/sigstore/cosign/pkg/cosign/kms/gcp"
+	"github.com/sigstore/sigstore/pkg/signature"
+)
+
+func New(ctx context.Context, signer string) (signature.Signer, error) {
+	switch {
+	case strings.HasPrefix(signer, gcp.ReferenceScheme):
+		return gcp.NewGCP(ctx, signer)
+	default:
```
dlorenc, line 34: Feel free to track this as a TODO or separate issue, but we might want to support an in memory signer or something that uses a k8s secret as a standard fallback or for local development.

cmd/rekor-cli/app/root_test.go
```diff
@@ -57,5 +57,5 @@ func TestAPIKey(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	_, _ = client.Tlog.GetPublicKey(nil)
+	_, _ = client.Pubkey.GetPublicKey(nil)
```
dlorenc, line 59: I'm not terribly opinionated about the location, but would it be easier for backwards compat to keep it under client.TLog?

dlorenc (commented):
Nice!

lukehinds:
Would be nice to get some test coverage where we can.

pkg/api/public_key.go
bobcallaway: I think we can handle that in a separate PR (it was representing the public key(s) used to sign log entries in a JSON Web Key Set format (RFC 7517)). It's probably only worth considering that if we would have multiple signing keys in use at the same time.

pkg/signer/signer.go
bobcallaway: +1 I think we should default to something that can run completely disconnected.

cmd/rekor-cli/app/root_test.go
bobcallaway: This only changes the client bindings, not the actual REST API in any way; so it would be easier to leave it in Tlog... but I don't feel strongly about this.

pkg/api/public_key.go
```diff
@@ -0,0 +1,39 @@
...
+	"encoding/pem"
+	"fmt"
+	"net/http"
+
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/sigstore/rekor/pkg/generated/restapi/operations/pubkey"
+)
+
+func GetPublicKeyHandler(params pubkey.GetPublicKeyParams) middleware.Responder {
+	b, err := x509.MarshalPKIXPublicKey(api.pubkey)
```
bobcallaway: Is there a reason to re-marshal & re-encode this on every request? We seem to be initializing `api.pubkey` at startup, and never changing it ever again in the life of the process.

bobcallaway (commented):
generally looks good to me - some unit test coverage on Signer would be good.

pkg/signer/signer.go
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+package signer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/sigstore/cosign/pkg/cosign/kms/gcp"
```
dlorenc: We should probably try to avoid importing cosign into rekor here too, can we move the signer interface up into sigstore itself?

pkg/api/public_key.go
priyawadhwa: I opened https://github.com/sigstore/rekor/issues/271 so that we don't forget about this!

pkg/signer/signer.go
priyawadhwa: Yup! I think the signer interface is already there but the kms/gcp specific code isn't. Just opened https://github.com/sigstore/sigstore/pull/41 to move it.

pkg/api/public_key.go
priyawadhwa: Good point, changed it to store `api.pubkey` as a string and return it directly.

pkg/signer/signer.go
priyawadhwa: sgtm, added the in-memory signer! I think it's needed in this PR for the integration tests.

pkg/signer/signer.go
priyawadhwa: cosign imports should be removed now!

tests/e2e_test.go
```diff
@@ -110,6 +110,7 @@ func TestUploadVerifyRpm(t *testing.T) {
 }
 
 func TestLogInfo(t *testing.T) {
+	t.Skip("TODO (priyawadhwa@): skipping for now, turn on once rekor signs the log root itself")
```
priyawadhwa: cc @dlorenc @lukehinds @bobcallaway -- this test fails when log-info runs `VerifySignedLogRoot` -- I'm guessing it's because the public key we return now is not the one in trillian.

I added a skip in with plans to turn the test on once we start signing roots ourselves, but if anyone has other ideas lmk!

bobcallaway: I hit this earlier but after a rebase all was well.

cmd/rekor-cli/app/root_test.go
priyawadhwa: I might leave it this way then, since the public key isn't really associated with the tlog anymore. Happy to switch back if anyone prefers that!

tests/e2e_test.go
priyawadhwa: still failing 😞

priyawadhwa: Added in signing of the signature for the log root, this should be working now!

dlorenc (dismissed):
Nice! The real test will come from trying to update trillian now :) I'll give that a try locally.

dlorenc:
OK, I've verified that this works with trillian at 1.3.13. It doesn't completely work with trillian at HEAD, we need to make a few more changes still, but I did manage to do it locally with a bit of hacking.

We need to mostly fix up the signature verification logic client side and do a bit more refactoring. I can't think of any way to do it in more of a piecemeal fashion so I think we need to get this in then keep updating more parts as we go.

Specifically, we need to get rid of all calls to `NewLogVerifier` and do that ourselves.

dlorenc:
I'll leave this open for others to look at tomorrow but then we should merge and keep the refactor going, nice work @priyawadhwa!

cmd/rekor-server/app/root.go
```diff
@@ -61,6 +61,8 @@ func init() {
 	rootCmd.PersistentFlags().Uint16("trillian_log_server.port", 8091, "Trillian log server port")
 	rootCmd.PersistentFlags().Uint("trillian_log_server.tlog_id", 0, "Trillian tree id")
 	rootCmd.PersistentFlags().String("rekor_server.address", "127.0.0.1", "Address to bind to")
+	rootCmd.PersistentFlags().String("rekor_server.signer", "", "Rekor signer to use. Current valid options include: [gcpkms]")
```
bobcallaway: do we need to mention the memory signing option here (or make it the explicit default)?

bobcallaway (dismissed):
one minor thing that needs to be resolved before merge - but otherwise looks good! thanks @priyawadhwa

cmd/rekor-server/app/root.go
priyawadhwa: good idea, updated!

dlorenc:
Nice! Looks like some merge conflicts slipped in, want to squash/rebase then merge?

priyawadhwa:
Fixed the conflict, I'll probably choose the "squash+merge" option!

dlorenc (approved):

PR #272, opened 2021-04-14
Add new type for JAR archives

bobcallaway:
This adds support for a new pluggable type that can extract signatures from signed JAR files. Per the JAR spec, a special manifest file is created with the digest hashes of all included content in the JAR file. It is this special manifest file that is then signed, and included in a file within the archive in PKCS7 format. The PKCS7 file also includes the X509 certificate that can be used to verify the signed manifest file inside of the JAR.


cmd/rekor-cli/app/pflags.go
```diff
@@ -130,14 +131,79 @@ func validateArtifactPFlags(uuidValid, indexValid bool) error {
...
+		var jarBytes []byte
+		jarURL, err := url.Parse(jar)
+		if err == nil && jarURL.IsAbs() {
+			/* #nosec G107 */
+			rpmResp, err := http.Get(jar)
+			if err != nil {
+				return nil, fmt.Errorf("error fetching 'jar': %w", err)
+			}
+			defer rpmResp.Body.Close()
+			jarBytes, err = ioutil.ReadAll(rpmResp.Body)
```
dlorenc: nit: jarresp?

cmd/rekor-cli/app/pflags.go
```diff
@@ -130,14 +131,79 @@ func validateArtifactPFlags(uuidValid, indexValid bool) error {
...
+
+			//TODO: ensure this is a valid JAR file; look for META-INF/MANIFEST.MF?
+			re.JARModel.Archive.Content = strfmt.Base64(artifactBytes)
+		}
+
+		if err := re.Validate(); err != nil {
+			return nil, err
+		}
+
+		if re.HasExternalEntities() {
```
dlorenc, line 195: Would a jar ever have external entities? I think this is always false?

pkg/types/jar/v0.0.1/entry.go
```diff
@@ -0,0 +1,342 @@
...
+
+func (v V001Entry) APIVersion() string {
+	return APIVERSION
+}
+
+func NewEntry() types.EntryImpl {
+	return &V001Entry{}
+}
+
+func Base64StringtoByteArray() mapstructure.DecodeHookFunc {
```
dlorenc, line 74: we have this defined in two places already, should we put it somewhere common?

pkg/types/jar/v0.0.1/entry_test.go
```diff
@@ -0,0 +1,229 @@
+/*
+Copyright © 2021 The Rekor Authors
```
dlorenc: should these all say sigstore?

dlorenc (commented):
A couple nits, looks awesome!

cmd/rekor-cli/app/pflags.go
bobcallaway: This is meant to declare whether the JSON doc posted to Rekor needs to fetch an external entity (i.e. download the JAR file from an HTTP(S) server).

pkg/types/jar/v0.0.1/entry.go
bobcallaway: yes, this and the semver stuff could be abstracted out. I will open an issue for this.

pkg/types/jar/v0.0.1/entry_test.go
bobcallaway: I had this question - should we set this org-wide in all projects to just be "sigstore authors"?

lukehinds: I think @cpanato is planning to do this, he already made a couple of PRs to unity the license header. 

I think the name thing originates from when I first started the project with just me in a personal repo.

cmd/rekor-cli/app/pflags.go
dlorenc: Ahhh right, the entire jar can be external.

dlorenc (dismissed):

pkg/types/jar/v0.0.1/entry_test.go
cpanato: Yes, doing for the missing repos today or tomorrow (15.04.2021) :)

dlorenc (approved):

Issue #247, opened 2021-03-31
Add email search index to redis


lukehinds:
To allow searching via email, let's add it to the redis. Clients can then query rekor for entries according to an indenity.

dlorenc:
@asraa is working on this one!

bobcallaway:
Fixed by #276

PR #276, opened 2021-04-15
[redis] add email search index

asraa:

https://github.com/sigstore/rekor/issues/247

Adds email index to enable search queries like:
`./rekor search --email=asraa@google.com`

I added both subject alt name emails and emails in the Name (which is how cosign will add them for keyless)

TODO: Does anyone have any good email validators in go? Or maybe just a regex is fine?


lukehinds:
cool, thanks for picking this up.

> TODO: Does anyone have any good email validators in go? Or maybe just a regex is fine?

go-playground/validator is pretty good

pkg/types/rekord/v0.0.1/entry.go
```diff
@@ -103,6 +106,23 @@ func (v V001Entry) IndexKeys() []string {
 		}
 	}
 
+	// if this is a certificate, add email addresses to result
+	p, _ := pem.Decode(key)
+	if p != nil {
+		cert, err := x509.ParseCertificate(p.Bytes)
+		if err == nil {
+			var emailAddressOID asn1.ObjectIdentifier = []int{1, 2, 840, 113549, 1, 9, 1}
+			for _, email := range cert.EmailAddresses {
```
dlorenc: I think you can just pick one here, probably the cert.Subject.Names. Cosign fills them all in right now and it's a bit silly. Either way they'll always be the same.

pkg/types/rpm/v0.0.1/entry.go
```diff
@@ -107,6 +110,23 @@ func (v V001Entry) IndexKeys() []string {
 		}
 	}
 
+	// if this is a certificate, add email addresses to result
+	p, _ := pem.Decode(key)
+	if p != nil {
+		cert, err := x509.ParseCertificate(p.Bytes)
+		if err == nil {
+			var emailAddressOID asn1.ObjectIdentifier = []int{1, 2, 840, 113549, 1, 9, 1}
```
dlorenc: Could you make this a constant somewhere and link to the definition of it for reference?

dlorenc (commented):
Nice! One nit.

pkg/types/rekord/v0.0.1/entry.go
```diff
@@ -100,6 +102,19 @@ func (v V001Entry) IndexKeys() []string {
 		}
 	}
 
+	// if this is a certificate, add email addresses to result
```
bobcallaway: would it be better to attempt to typecast `v.keyObj` to pki/x509/PublicKey and just use the value we've already parsed? The struct fields aren't exported at the moment but I don't see why they couldn't be. 

I'd rather handle this at the time the keyObj is created rather than re-parse at this point.

pkg/types/rpm/v0.0.1/entry.go
```diff
@@ -104,6 +106,19 @@ func (v V001Entry) IndexKeys() []string {
 		}
 	}
 
+	// if this is a certificate, add email addresses to result
```
bobcallaway: RPMs are only signed with PGP so I think we would need to consider how to extract it from that key. Also see typecast from above. 

Should we consider extending the base class for public key to have a method for this and then simply call it here?

pkg/api/error.go
```diff
@@ -40,6 +40,7 @@ const (
 	malformedUUID                  = "UUID must be a 64-character hexadecimal string"
 	malformedHash                  = "Hash must be a 64-character hexadecimal string created from SHA256 algorithm"
 	malformedPublicKey             = "Public key provided could not be parsed"
+	malformedEmail                 = "Email must be a well formed email"
```
bobcallaway: nit: `s/well formed/well-formed`

bobcallaway (commented):
love the idea, just think we need to keep details of parsing this out of key objects within the PKI package instead of sprinkling it out into the types.

pkg/types/rpm/v0.0.1/entry.go
asraa: I like extending the base class! I'll do that and try to support pgp -- will try to get the email out of the User ID (and hope it has an angle bracket enclosed email)

pkg/types/rekord/v0.0.1/entry.go
asraa: done

pkg/types/rpm/v0.0.1/entry.go
asraa: Done, I got the PGP User ID Email (https://github.com/golang/crypto/blob/eec23a3978ad/openpgp/packet/userid.go#L16), but in case it's in Name, I don't parse it.

asraa:
(Have to fix up the pkcs7 merge)

asraa:
PKCS7 emails/tests added

cmd/rekor-cli/app/pflags.go
```diff
@@ -47,6 +48,8 @@ func addSearchPFlags(cmd *cobra.Command) error {
 	cmd.Flags().Var(&fileOrURLFlag{}, "artifact", "path or URL to artifact file")
 
 	cmd.Flags().Var(&uuidFlag{}, "sha", "the SHA256 sum of the artifact")
+
+	cmd.Flags().Var(&emailFlag{}, "email", "email used in the x509 certificate")
```
bobcallaway: small nit - maybe make this description not specific to x509?

pkg/api/index.go
```diff
@@ -77,6 +77,16 @@ func SearchIndexHandler(params index.SearchIndexParams) middleware.Responder {
 		}
 		result = append(result, resultUUIDs...)
 	}
+	if params.Query.Email != "" {
+		if !govalidator.IsEmail(params.Query.Email) {
```
bobcallaway: we're using different validator packages in client VS server; was this intentional?

bobcallaway (commented):
lgtm other than these two questions

pkg/api/index.go
asraa: I went with the already used one in this file. I'll make it uniform:
https://github.com/go-playground/validator to check sha would be a length check (rather than a `IsSHA256` check) and both have email. I would prefer to use this one, WDYT?

bobcallaway: https://github.com/go-playground/validator has cross-field validation... sweet! I like that one better too.

pkg/api/index.go
```diff
@@ -36,11 +36,13 @@ import (
 
 func SearchIndexHandler(params index.SearchIndexParams) middleware.Responder {
 	httpReqCtx := params.HTTPRequest.Context()
+	var validate *validator.Validate
 
 	var result []string
 	if params.Query.Hash != "" {
 		// validate this is only a valid sha256 hash
-		if !govalidator.IsSHA256(params.Query.Hash) {
+		errs := validate.Var(params.Query.Hash, "len=64")
```
bobcallaway: Since we define a pattern in https://github.com/sigstore/rekor/blob/4d98f80a4907314f3d1f641abb956d4996731bf3/openapi.yaml#L327 I think this check is unnecessary.

pkg/api/index.go
```diff
@@ -77,6 +79,17 @@ func SearchIndexHandler(params index.SearchIndexParams) middleware.Responder {
 		}
 		result = append(result, resultUUIDs...)
 	}
+	if params.Query.Email != "" {
+		errs := validate.Var(params.Query.Hash, "email")
```
bobcallaway: shouldn't this be params.Query.Email here?

openapi.yaml
```diff
@@ -308,6 +308,8 @@ definitions:
   SearchIndex:
     type: object
     properties:
+      email:
+        type: string
```
bobcallaway, line 312: I think you could add `format: email` on the line below this and it will be validated by the go-openapi runtime.

pkg/pki/pgp/pgp.go
```diff
@@ -277,3 +277,17 @@ func (k PublicKey) KeyRing() (openpgp.KeyRing, error) {
...
 	return k.key, nil
 }
+
+// EmailAddresses implements the pki.PublicKey interface
+func (k PublicKey) EmailAddresses() []string {
+	var names []string
+	// Extract from cert
+	for _, entity := range k.key {
+		for _, identity := range entity.Identities {
+			if len(identity.UserId.Email) > 0 {
```
bobcallaway: should we do any checking here for valid formatting?

asraa: Yeah, looks like it's not enforced https://www.rfc-editor.org/rfc/rfc4880.html#section-5.11

bobcallaway (approved):

dlorenc:
Ready to merge? I can deploy whenever.

asraa:
Just rebased!

PR #281, opened 2021-04-15
 Update trillian dependency 

priyawadhwa:
This PR updates to trillian @ HEAD, which no longer contains verification logic.  It removes calls to the trillian verifier and replaces them with calls to a new rekor verify package!

The `verify` package currently only verifies the signed log root.


docker-compose.yml
```diff
@@ -16,7 +16,7 @@
 version: '3.4'
 services:
   mysql:
-    image: gcr.io/trillian-opensource-ci/db_server:v1.3.13
+    image: gcr.io/trillian-opensource-ci/db_server:5e12fb368c8fd19e10aeb5a5cf785107f8069c08
```
dlorenc, line 19: Oh nice, these are the actual commits it was built at!

pkg/verify/log_root.go
```diff
@@ -0,0 +1,64 @@
...
+
+	var lr types.LogRootV1
+	if err := lr.UnmarshalBinary(logRoot); err != nil {
+		return nil, err
+	}
+	return &lr, nil
+}
+
+// Verify cryptographically verifies the output of Signer.
+func Verify(pub crypto.PublicKey, hasher crypto.Hash, data, sig []byte) error {
```
dlorenc: Could this be made private? It's soooorttt of duplication from stuff we have in sigstore/sigstore but I understand why it's here. I just don't want this to spread into a general purpose crytpo verifier.

dlorenc (commented):
one nit

pkg/verify/log_root.go
priyawadhwa: good call!

dlorenc:
Nice! I want @bobcallaway to have a look here too.

cmd/rekor-cli/app/log_info.go
```diff
@@ -76,10 +74,6 @@ var logInfoCmd = &cobra.Command{
 
 		logInfo := result.GetPayload()
 
-		keyHint, err := base64.StdEncoding.DecodeString(logInfo.SignedTreeHead.KeyHint.String())
-		if err != nil {
-			return nil, err
-		}
 		logRoot, err := base64.StdEncoding.DecodeString(logInfo.SignedTreeHead.LogRoot.String())
```
bobcallaway: This is my fault (per git blame), but 
`logRoot := *logInfo.SignedTreeHead.LogRoot` will give the same `[]byte` with out encoding and decoding to base64 for this and the line for signature below

pkg/api/trillian_client.go
```diff
@@ -216,7 +217,9 @@ func (t *TrillianClient) getLeafAndProofByIndex(index int64) *Response {
 		})
 
 	if resp != nil && resp.Proof != nil {
-		if err := t.verifier.VerifyInclusionAtIndex(&root, resp.Leaf.LeafValue, index, resp.Proof.Hashes); err != nil {
+		leaf := t.verifier.BuildLeaf(resp.Leaf.LeafValue)
```
bobcallaway: we already have `resp.Leaf` at this point; why do we rebuild it here?

cmd/rekor-cli/app/log_info.go
bobcallaway: there are several other places where this simplification can be done too... would you mind cleaning those up too? thanks!

pkg/verify/log_root.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/google/trillian/types"
+	"github.com/pkg/errors"
+)
+
+// this verification copied from https://github.com/google/trillian/blob/v1.3.13/crypto/verifier.go
+// which has since been deleted
+
+// SignedLogRoot verifies the signed log root and returns its contents
+func SignedLogRoot(pub crypto.PublicKey, hash crypto.Hash, logRoot, logRootSignature []byte) (*types.LogRootV1, error) {
```
bobcallaway: nit: if we're hardcoding the hash to SHA256, is there a reason why we're exposing this as an argument here?

bobcallaway (commented):
a couple nits but overall good stuff @priyawadhwa !

cmd/rekor-cli/app/log_info.go
priyawadhwa: Thanks! fixed those up where i could find them.

pkg/verify/log_root.go
priyawadhwa: nope, removed

pkg/api/trillian_client.go
priyawadhwa: thanks! removed

docker-compose.yml
```diff
@@ -45,7 +45,7 @@ services:
       retries: 3
       start_period: 5s
   trillian-log-server:
-    image: gcr.io/trillian-opensource-ci/log_server:v1.3.13
+    image: gcr.io/trillian-opensource-ci/log_server:5e12fb368c8fd19e10aeb5a5cf785107f8069c08
```
dlorenc, line 48: OK, so actually I think we need to split this change up into two so we're safe. It's good you tested with these containers, but we need to also prove that the code works with the old containers so we can roll forward/back safely.

So could you drop this part from this PR, when tests pass we'll merge, then we send another PR to bump this part, when tests pass there we then roll out in prod?

dlorenc:
OK, we tested that this can get rolled out safely.

dlorenc (approved):

PR #285, opened 2021-04-21
Add a flag to specify an artifact by hash. This is required when uplo…

dlorenc:
…ading by URL.

### DIFF
--- a/cmd/rekor-cli/app/pflags.go
+++ b/cmd/rekor-cli/app/pflags.go
@@ -79,6 +79,7 @@ func addArtifactPFlags(cmd *cobra.Command) error {
 	cmd.Flags().Var(&fileOrURLFlag{}, "public-key", "path or URL to public key file")
 
 	cmd.Flags().Var(&fileOrURLFlag{}, "artifact", "path or URL to artifact file")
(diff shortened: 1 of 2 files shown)


cmd/rekor-cli/app/pflags.go
```diff
@@ -79,6 +79,7 @@ func addArtifactPFlags(cmd *cobra.Command) error {
 	cmd.Flags().Var(&fileOrURLFlag{}, "public-key", "path or URL to public key file")
 
 	cmd.Flags().Var(&fileOrURLFlag{}, "artifact", "path or URL to artifact file")
+	cmd.Flags().String("artifact-hash", "", "hex encdoed hash of artifact (when using URL)")
```
bobcallaway: I'd change this to be of type `uuidFlag` so you get the checking for hex string of 64 char long; also maybe put that it must be SHA256 in the help text?

dlorenc: Nice,done.

lukehinds (approved):
looks ok to me, will `https://get.jenkins.io/war-stable/2.277.3/jenkins.war` persist?

dlorenc:
It's an LTS release, so about as long as anything else I'd guess :)

PR #291, opened 2021-04-27
Add signedEntryTimestamp to LogEntry for offline verification

priyawadhwa:
This PR adds a signature field to LogEntry (I'm open to different names for signature here since it might be overused). It can be used to verify the LogEntry was signed for offline verification, which cosign will be using for bundling!

Basically, rekor will:
1. Sign the LogEntry
2. Add the signature to the "Verification.SignedEntryTimestamp" field of the LogEntry
3. Send that to the client whenever they create a new entry

and clients will:
1. Get the LogEntry when they create a new entry, and set signature=nil to get the payload (making people set signature=nil is kinda weird, but follows the precert model where the poison extension has to be removed to validate the cert). I went this route so that the signature could be bundled with the LogEntry which is already being returned when users add an entry to the log.
2. Validate the payload and signature against the rekor public key

Also adds an integration test for this.

cc @dlorenc @asraa

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -299,12 +299,15 @@ definitions:
           additionalProperties: true
         integratedTime:
           type: integer
-        inclusionProof:
-          $ref: '#/definitions/InclusionProof'
-        signature:
-          type: string
-          format: byte
-          description: signature over the log entry (acts as an inclusion promise if inclusion proof is not included)
+        verification:
+          type: object
+          properties:
+            inclusionProof:
+              $ref: '#/definitions/InclusionProof'
+            signedEntryTimestamp:
+              type: string
+              format: byte
+              description: signature over the logIndex, body and integratedTime
       required:
         - "logIndex"
         - "body"
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -308,10 +308,9 @@ definitions:
               type: string
               format: byte
               # To verify the signedEntryTimestamp:
-                # 1. Set LogEntry.Verification = nil
-                # 2. Get the payload by running LogEntry.MarshalBinary()
-                # 3. Canonicalize the payload by running jsoncanonicalizer.Transform(payload)
-                # 4. Verify the canonicalized payload and signedEntryTimestamp against rekor's public key
+                # 1. Remove the Verification object from the JSON Document
+                # 2. Canonicalize the remaining JSON document by following RFC 8785 rules
+                # 3. Verify the canonicalized payload and signedEntryTimestamp against rekor's public key
               description: Signature over the logIndex, body and integratedTime. 
       required:
         - "logIndex"
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -307,7 +307,12 @@ definitions:
             signedEntryTimestamp:
               type: string
               format: byte
-              description: signature over the logIndex, body and integratedTime
+              # To verify the signedEntryTimestamp:
+                # 1. Set LogEntry.Verification = nil
+                # 2. Get the payload by running LogEntry.MarshalBinary()
+                # 3. Canonicalize the payload by running jsoncanonicalizer.Transform(payload)
+                # 4. Verify the canonicalized payload and signedEntryTimestamp against rekor's public key
+              description: Signature over the logIndex, body and integratedTime. 
       required:
         - "logIndex"
         - "body"
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -301,6 +301,10 @@ definitions:
           type: integer
         inclusionProof:
           $ref: '#/definitions/InclusionProof'
+        signature:
+          type: string
+          format: byte
+          description: signature over the log entry (acts as an inclusion promise if inclusion proof is not included)
       required:
         - "logIndex"
         - "body"
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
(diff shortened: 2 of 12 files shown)


openapi.yaml
```diff
@@ -301,6 +301,10 @@ definitions:
           type: integer
         inclusionProof:
           $ref: '#/definitions/InclusionProof'
+        signature:
```
dlorenc: we should probably be a bit more verbose with the name, `signedEntryTimestamp` maybe?

tests/e2e_test.go
```diff
@@ -359,3 +374,117 @@ func TestWatch(t *testing.T) {
...
+	}
+	params := entries.NewCreateLogEntryParams()
+	params.SetProposedEntry(&returnVal)
+	resp, err := rekorClient.Entries.CreateLogEntry(params)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// unmarshal the response
+	le := resp.GetPayload()
+	if len(le) != 1 {
```
dlorenc: we should make a helper for this at some point (getting the one field out), it's so annoying to get right and we do it all over the place

dlorenc (commented):
This is awesome, a couple nits then I want @bobcallaway to check the API change

openapi.yaml
asraa: I almost feel like log entry should be a type in itself with logIndex, body, and integrated time so it's clear log entry includes those three (and not e.g. inclusion proof)

asraa: btw, do we want this to be replaced with an RFC 3161 signed timestamp over (logIndex, body) in the future?

dlorenc: Probably not replaced by, but in addition to? I don't love the idea of promoting usage of the RFC3161 format outside of where we need it for compatibility.

asraa: Oh yeah, definitely. the other option is a JSON based one, which would look roughly like a packaged signature field + message digest + cert chain

openapi.yaml
```diff
@@ -301,6 +301,10 @@ definitions:
...
+          type: string
+          format: byte
+          description: signature over the log entry (acts as an inclusion promise if inclusion proof is not included)
```
bobcallaway: right now Rekor is blocking on the insertion loop so this wouldn't really be a promise... but more of a static attestation of inclusion. Not sure how much to stress over the semantics here.

pkg/api/entries.go
```diff
@@ -156,6 +155,21 @@ func CreateLogEntryHandler(params entries.CreateLogEntryParams) middleware.Respo
 		}()
 	}
 
+	payload, err := logEntryAnon.MarshalBinary()
```
bobcallaway, line 161: This is just marshaling to JSON under the covers... we should probably marshal it in a JCS-compatible way before signing so that it is repeatable for clients to verify the hash.  

Also since we support returning data in YAML format I think we need to be explicit in documenting how we format the data before signing since there would be a potential for mismatch there too.

openapi.yaml
bobcallaway: I'm also in favor of shifting the inclusion proof and the signature into their own level of the LogEntry object. We should also probably add the log ID (in case we ever needed to rotate to a different log for performance / maintenance reasons)

perhaps something like:
```
LogEntry:
    type: object
    additionalProperties:
      type: object
      properties:
        logId:
          type: string
        logIndex:
          type: integer
          minimum: 0
        body:
          type: object
          additionalProperties: true
        integratedTime:
          type: integer
        verification:
          type: object
          properties:
            inclusionProof:
              $ref: '#/definitions/InclusionProof'
            signedEntryTimestamp:
              $ref: '#/definitions/SignedEntryTimestamp'
      required:
        - "logId"
        - "logIndex"
        - "body"
        - "integratedTime"
```

openapi.yaml
dlorenc: Right, I was thinking we might want to leave it open for interpretation later in case we ever need to decouple these. I'm fine keeping it simple though.

openapi.yaml
priyawadhwa: Cool, I renamed it to `signedEntryTimestamp` and moved `signedEntryTimestamp` and `inclusionProof` under `verification`

I figure we can add in the logID in a separate PR so that this one doesn't get too big!

openapi.yaml
priyawadhwa: sgtm, I made the description more general here

pkg/api/entries.go
priyawadhwa: Cool, I found this json canonicalization library & added it in: https://github.com/cyberphone/json-canonicalization

Where would be the best spot for the documentation? I could make a `docs` folder in the repo if there's isn't an existing spot for this kind of thing.

dlorenc: Nothing really right now outside of the openapi.yaml. I'd try to be as verbose there as you can and then see if we need more space.

tests/e2e_test.go
priyawadhwa: I added it as a helper func for testing, but if we're doing it in code I could move it to a general util package or something

pkg/api/entries.go
priyawadhwa: cool, I added in instructions for how to verify as comments

dlorenc (commented):
Nice, LGTM now.

@asraa and @bobcallaway any other thoughts?

openapi.yaml
```diff
@@ -299,8 +299,20 @@ definitions:
...
-          $ref: '#/definitions/InclusionProof'
+        verification:
+          type: object
+          properties:
+            inclusionProof:
+              $ref: '#/definitions/InclusionProof'
+            signedEntryTimestamp:
+              type: string
+              format: byte
+              # To verify the signedEntryTimestamp:
```
bobcallaway, line 310: if you end up making another change to this, I might add change the wording to be a bit less Golang specific - for example:

- Remove the verification object from the JSON document
- Follow RFC 8785 rules to canonicalize the remaining JSON document

bobcallaway:
> Nice, LGTM now.
> 
> @asraa and @bobcallaway any other thoughts?

LGTM - should we make any changes to `rekor-cli upload` to print the signature or verify it (could be a separate PR)?

asraa (approved):
LGTM!

openapi.yaml
priyawadhwa: good call, added that in

priyawadhwa:
>LGTM - should we make any changes to rekor-cli upload to print the signature or verify it (could be a separate PR)?

SGTM. I'll open a PR once this is merged, tracking here: https://github.com/sigstore/rekor/issues/292

dlorenc (approved):

bobcallaway (approved):

Issue #232, opened 2021-03-24
Proposal: Add an RFC3161 Timestamp server to our API


dlorenc:
This would ease integration with Java and other tooling that expects timestamps in this format. We could expose two APIs:

- A normal RFC3161 variant, where users send us a TSR and we reply with a timestamp
- A less difficult to use one, where users just send us a blob and we return a timestamp, in the RFC3161 format.

Both variants would log signatures to Rekor, and could include the index in the response.

dlorenc:
This gets a lot easier with #262

dlorenc:
@bobcallaway - does this help with your maven plugin at all? I know java expects timestamps.

dlorenc:
https://blog.cyborch.com/trusted-time/ for reference on what a more "usable" rfc3161 could look like :)

bobcallaway:
> @bobcallaway - does this help with your maven plugin at all? I know java expects timestamps.

I think it could, but need to think more about the sequencing of all of this.

Hayden-IO:
Closing - A timestamp authority was once implemented in rekor, then removed, and reimplemented in https://github.com/sigstore/timestamp-authority

PR #293, opened 2021-04-29
Implement RFC 3161 timestamp responses

asraa:

Part of https://github.com/sigstore/rekor/issues/232
Adds RFC 3161 timestamp response generation. The CLI takes in an artifact file or artifact hash to timestamp. It writes the response locally (by default, to response.tsr). For e.g.

```
rekor timestamp --artifact file.txt --out /tmp/response.tsr
```

OR

```
$ ./rekor timestamp --artifact-hash edeaaff3f1774ad2888673770c6d64097e391bc362d7d6fb34982ddf0efd18cb --rekor_server http://localhost:3000

Wrote response to: response.tsr
$ openssl ts -verify -digest edeaaff3f1774ad2888673770c6d64097e391bc362d7d6fb34982ddf0efd18cb -in response.tsr --CAfile root.pem
Using configuration from /usr/lib/ssl/openssl.cnf
Verification: OK
```

Or without the CLI, just pass in an RFC 3161 timestamp query
```
$ curl -H "Content-Type: application/timestamp-query" --data-binary '@file.tsq' http://localhost:3000/api/v1/timestamp --output response.tsr
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1923  100  1832  100    91   596k  30333 --:--:-- --:--:-- --:--:--  625k
$ openssl ts -verify -data file.txt -in response.tsr --CAfile root.pem
Using configuration from /usr/lib/ssl/openssl.cnf
Verification: OK
```

You can grab the timestamp cert and root CA cert used to verify through the `api/v1/timestamp/certchain` endpoint.

**FOLLOW UPS**
* Right now there are TODOs marking that this does not upload to the tlog. Will implement in a follow up
* A production instance of rekor using GCP KMS will not generate timestamp responses yet (it will not have a timestamping cert chain to use and return early). I will need to add the ability to pass the cert chain through a server flag 
* Handful of follow ups to tighten the RFC 3161 spec (unique serial numbers)


bobcallaway:
> * Eventually I want to add a --upload flag here that would also upload the response to the log.
> * Or upload would be similar to the JAR type, you would generate and then upload the response `rekor upload --artifact <response.tsr> --type timestamp`
> * You could combine the two and ask rekor to upload a timestamped request  `rekor upload --file <file.txt> --format rfc3161 --type timestamp`, returning the response and index locally.
> 
> If the third sounds best, I'll refactor this and add the timestamp type first (still proxying to generate the response if accept requests to upload). I think [2] is the next step, personally.

I guess I would vote we always make an entry in the log for a timestamp whenever one is requested (or in the future generated ourselves)... this would be done server-side before returning the original RFC3161 response to the client. If the client then wants to embed it in a JAR they could do that, but I can't think of a reason not to put it into the log when the server has it.

dlorenc:
> I guess I would vote we always make an entry in the log for a timestamp whenever one is requested (or in the future generated ourselves)... this would be done server-side before returning the original RFC3161 response to the client. If the client then wants to embed it in a JAR they could do that, but I can't think of a reason not to put it into the log when the server has it.

+1, otherwise you could just use the other TSA directly.

asraa:
This has been revamped as step 1 of:
1. Add rekor ability to serve TSRs
   - Add flag to start rekor server with a timestamping cert chain (for gcpkms)
2. Add upload to tlog when server generates response
3. Add JSON format support

cmd/rekor-cli/app/timestamp.go
```diff
@@ -0,0 +1,198 @@
...
+		if err != nil {
+			return nil, err
+		}
+		body, err := base64.StdEncoding.DecodeString(resp.Payload.RfcResponse.String())
+		if err != nil {
+			return nil, err
+		}
+
+		// Sanity check response and check if the TimeStampToken was successfully created
+		_, err = timestampReq.ParseResponse(body)
```
dlorenc: tiny nit, this can get inlined.

cmd/rekor-cli/app/timestamp.go
```diff
@@ -0,0 +1,198 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Write response to file
+		outStr := viper.GetString("out")
+		if outStr == "" {
```
dlorenc, line 232: Should we default to stdout maybe?

cmd/rekor-cli/app/timestamp.go
```diff
@@ -0,0 +1,198 @@
...
+			outStr = "response.tsr"
+		}
+		f, err := os.Create(outStr)
+		if err != nil {
+			return nil, err
+		}
+		defer f.Close()
```
dlorenc: I think this could get shortended to ioutil.WriteFile()

openapi.yaml
```diff
@@ -89,6 +89,23 @@ paths:
         default:
           $ref: '#/responses/InternalServerError'
 
+  /api/v1/log/timestampCertChain:
```
dlorenc: should this be `/api/v1/tsr/certchain` maybe?

pkg/api/api.go
```diff
@@ -113,12 +115,31 @@ func NewAPI() (*API, error) {
 		return nil, errors.Wrap(err, "new verifier")
 	}
 
+	// verify cert chain if present
+	if len(certChain) > 0 {
+		roots := x509.NewCertPool()
+		intermediates := x509.NewCertPool()
+		for _, cert := range certChain[1:(len(certChain) - 1)] {
```
dlorenc: Would this easier overall if the API served the cert and chain in separate fields? I can't decide myself but might have done that in Fulcio.

pkg/api/api.go
```diff
@@ -113,12 +115,31 @@ func NewAPI() (*API, error) {
...
+			intermediates.AddCert(cert)
+		}
+		roots.AddCert(certChain[len(certChain)-1])
+		_, err = certChain[0].Verify(x509.VerifyOptions{
```
dlorenc: This could get inlined

dlorenc (commented):
Awesome! A few nits/questions.

cmd/rekor-cli/app/timestamp.go
asraa: Hmm it's kinda gross binary file. But tbh most of the time you pipe into a response.tsr after making a request, so yeah, why not.

pkg/api/api.go
asraa: Yes!! WDYT about the spinning up the rekor server -- should you provide a timestamp cert and a root CA cert separately too? since i guess the root CA cert can also be used in other cert chains

cmd/rekor-cli/app/timestamp.go
```diff
@@ -0,0 +1,198 @@
...
+func validateTimestampFlags() error {
+	requestStr := viper.GetString("request")
+	fileStr := viper.GetString("file")
+	if requestStr == "" && fileStr == "" {
+		return errors.New("request or file must be specified")
+	}
+	return nil
+}
+
+func createRequestFromFlags() (*pkcs9.TimeStampReq, error) {
```
bobcallaway, line 128: Is there existing tooling to easily create the TSR... I'm thinking a better UX would be to support some additional command line options like https://www.openssl.org/docs/man1.0.2/man1/ts.html instead of requiring a pre-formed binary input.

cmd/rekor-cli/app/timestamp.go
bobcallaway: +1

openapi.yaml
```diff
@@ -224,6 +241,33 @@ paths:
...
+  
+  /api/v1/tsr:
+    post:
+      # TODO: Add uploads to transparency log.
+      summary: Returns a timestamp response generated by Rekor
+      operationId: getTimestampResponse
+      tags:
+        - timestamp
+      consumes:
+        - application/json
```
bobcallaway: Is there a reason to wrap this in JSON vs just sending as `application/timestamp-query` and receiving `timestamp-reply` per the RFC? That would allow us to better integrate with tools like jarsigner.

pkg/signer/signer.go
```diff
@@ -18,20 +18,27 @@ package signer
...
 	"github.com/sigstore/sigstore/pkg/signature"
 )
 
-func New(ctx context.Context, signer string) (signature.Signer, error) {
+func New(ctx context.Context, signer string) (signature.Signer, []*x509.Certificate, error) {
 	switch {
 	case strings.HasPrefix(signer, gcp.ReferenceScheme):
-		return gcp.NewGCP(ctx, signer)
+		kms, err := gcp.NewGCP(ctx, signer)
+		// TODO: Add timestamping cert issued by Fulcio and Fulcio's Root CA.
```
bobcallaway: I feel like we're intermingling too many things on this API... if the timestamp cert is imported (and not part of the interaction b/t Rekor and GCP KMS, this feels a bit awkward...

openapi.yaml
bobcallaway: I would rather see this end up as `/api/v1/timestamp/certchain` or perhaps update this and the public key endpoint to be grouped together?

openapi.yaml
```diff
@@ -89,6 +89,23 @@ paths:
...
+      operationId: getTimestampCertChain
+      tags:
+        - timestamp
+      produces:
+        - application/x-pem-file
+      responses:
+        200:
+          description: The PEM encoded cert chain
+          schema:
+            type: string
```
bobcallaway: do we need a 404 if not we don't have time stamping configured?

openapi.yaml
```diff
@@ -224,6 +241,33 @@ paths:
           $ref: '#/responses/BadContent'
         default:
           $ref: '#/responses/InternalServerError'
+  
+  /api/v1/tsr:
```
bobcallaway: how about `/api/v1/timestamp`

openapi.yaml
bobcallaway: we could then have the `application/json` type be the non-ASN1 format that we'd rather folks use once it is implemented.

openapi.yaml
```diff
@@ -448,6 +492,22 @@ definitions:
       - treeSize
       - hashes
 
+  TimestampRequest:
+    type: object
```
bobcallaway: if we back off from JSON above these will need to change

pkg/api/api.go
```diff
@@ -54,7 +54,9 @@ type API struct {
 	pubkey     string // PEM encoded public key
 	pubkeyHash string // SHA256 hash of DER-encoded public key
 	signer     signature.Signer
-	verifier   *client.LogVerifier
+	// timestamping cert chain
+	certChain []*x509.Certificate
```
bobcallaway: any chance we could store both this and the PEM encoded version so we don't waste cycles re-encoding?

pkg/util/rfc3161.go
```diff
@@ -0,0 +1,225 @@
...
+	msg := new(pkcs9.TimeStampReq)
+	if rest, err := asn1.Unmarshal(data, msg); err != nil {
+		return nil, fmt.Errorf("error umarshalling request")
+	} else if len(rest) != 0 {
+		return nil, fmt.Errorf("error umarshalling request, trailing bytes")
+	}
+	return msg, nil
+}
+
+func CreateResponse(ctx context.Context, req pkcs9.TimeStampReq, certChain []*x509.Certificate, signer signature.Signer) (*pkcs9.TimeStampResp, error) {
```
bobcallaway: can we name this method something more descriptive since it is in `util`?

cmd/rekor-cli/app/timestamp.go
dlorenc: Ah, fair. Is there a standard pem encoding? If not then yeah keep it as the file I guess.

pkg/signer/signer.go
asraa: My other thought was to explicitly request one, but you would need provide the CA

openapi.yaml
asraa: Request yes, I'll do that.
The reply seemed to conflict if rekor uploads to the tlog by default, we'd be sending back some extra info with the response

Edit: upload to tlog directly would also conflict with directly writing the response to stdout...

bobcallaway: I think writing the RFC-compliant output to a file makes sense, and then just including the rekor information via stdout works too. You could mimic the CreateEntry behavior and return the URL to the entry in the Location HTTP response header without changing the body.

cmd/rekor-cli/app/timestamp.go
asraa: TSQ :P

The golang tools I've seen that relied on https://github.com/digitorus/timestamp. They implement everything openssl provides besides the random seed and specifying the digest bytes.

(We can't use their CreateResponse since we don't implement a crypto signer)

It would be pretty easy to do add support for these options myself.

cmd/rekor-cli/app/timestamp.go
asraa: Right I forgot the rationale here was that if I wrote to file, I could also include info on what the log index is when uploaded.

openapi.yaml
asraa: Ooooh that's a great idea -- I'll make this change, and update the TODOs about the uploading parts

cmd/rekor-cli/app/timestamp.go
asraa: OK added nonce bool to specify a nonce, string for TSA policy, string to specify the digest hash.

the only one from openssl i didn't include was the cert bool -- rekor always adds the cert chain. i could add it anyway though.

asraa:
Updated:
* The content type matches RFC 3161 timestap authorities with application/timestamp-query and application/timestamp-reply. See updated usage in PR description
* I changed the endpoint for the cert chain to timestamp/certchain. 

I think in real deployment rekor should just specify a root CA to make a request for a timestamp cert in order to build the chain. That hopefully reduces the too-much-api in running rekor with an already formed external cert chain

cmd/rekor-cli/app/timestamp.go
```diff
@@ -0,0 +1,251 @@
...
+	var digest []byte
+	if digestStr != "" {
+		decoded, err := hex.DecodeString(digestStr)
+		if err != nil {
+			return nil, err
+		}
+		digest = decoded
+	}
+	if digestStr == "" {
+		artifactStr := viper.GetString("artifact")
```
bobcallaway, line 157: perhaps for a followup PR - the rest of the rekor CLI commands take either URL or file path for artifact... might be worth mimicking that behavior here

cmd/rekor-cli/app/timestamp.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func (f *oidFlag) Type() string {
+	return "oidFlag"
+}
+
+func addTimestampFlags(cmd *cobra.Command) error {
+	cmd.Flags().Var(&fileFlag{}, "artifact", "path to an artifact to timestamp")
+	cmd.Flags().Var(&uuidFlag{}, "artifact-hash", "hex encoded SHA256 hash of the the artifact to timestamp")
+	cmd.Flags().Bool("nonce", false, "specify a pseudo-random nonce in the request")
```
bobcallaway: is there a reason not to default to `true` here for the nonce?

openapi.yaml
```diff
@@ -224,6 +224,53 @@ paths:
...
+          name: request
+          required: true
+          schema:
+            type: string
+            format: binary
+      responses:
+        200:
+          description: Returns a timestamp response 
+          schema:
+            type: string
```
bobcallaway, line 250: should this have `format: binary` as well?

openapi.yaml
```diff
@@ -224,6 +224,53 @@ paths:
...
+
+  /api/v1/timestamp/certchain:
+    get:
+      summary: Retrieve the certfiicate chain for timestamping that can be used to validate trusted timestamps
+      description: Returns the certfiicate chain for timestamping that can be used to validate trusted timestamps
+      operationId: getTimestampCertChain
+      tags:
+        - timestamp
+      produces:
+        - application/x-pem-file
```
bobcallaway: If we're returning multiple certs here from the chain, this should really be `application/pem-certificate-chain`

pkg/api/timestamp.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, err
+	}
+
+	return body, nil
+}
+
+func TimestampResponseHandler(params timestamp.GetTimestampResponseParams) middleware.Responder {
+	// Fail early if we don't haven't configured rekor with a certificate for timestamping.
+	if len(api.certChain) == 0 {
+		return handleRekorAPIError(params, http.StatusNotFound, errors.New("rekor is not configured to serve timestamps"), "")
```
bobcallaway: nit: this should be a 501 error (Not Implemented) not 404

tests/e2e_test.go
```diff
@@ -201,6 +204,7 @@ func TestGet(t *testing.T) {
 	outputContains(t, out, uuid)
 }
 
+/*
```
bobcallaway: are you intending to leave this in?

bobcallaway:
Are we expecting to fill out any of the other fields (accuracy, TSA) in our response:

```
rekor git:(ts) ✗ /usr/local/opt/openssl@1.1/bin/openssl ts -reply -in response.tsr -text
Using configuration from /usr/local/etc/openssl@1.1/openssl.cnf
Status info:
Status: Granted.
Status description: unspecified
Failure info: unspecified

TST info:
Version: 1
Policy OID: tsa_policy1
Hash Algorithm: sha256
Message data:
    0000 - 4a 9d 0a b4 e1 05 97 49-7e d6 c4 61 7c 98 3c 35   J......I~..a|.<5
    0010 - fa 9e 96 4e 75 cd d6 f9-fa e3 a0 da 19 29 ac c6   ...Nu........)..
Serial number: 0xFD290B0D05379DDE4F34858C
Time stamp: May 13 20:09:09 2021 GMT
Accuracy: unspecified
Ordering: no
Nonce: unspecified
TSA: unspecified
Extensions:
```

cmd/rekor-cli/app/root.go
```diff
@@ -124,6 +124,8 @@ func GetRekorClient(rekorServerURL string) (*client.Rekor, error) {
 	rt.Consumers["application/yaml"] = util.YamlConsumer()
 	rt.Consumers["application/x-pem-file"] = runtime.TextConsumer()
 	rt.Producers["application/yaml"] = util.YamlProducer()
+	rt.Producers["application/timestamp-query"] = runtime.TextProducer()
```
bobcallaway: This may work, but wondering if `runtime.ByteStreamProducer()` and `runtime.ByteStreamConsumer()` are more appropriate

pkg/generated/restapi/configure_rekor_server.go
```diff
@@ -70,6 +71,8 @@ func configureAPI(api *operations.RekorServerAPI) http.Handler {
 	api.YamlProducer = util.YamlProducer()
 
 	api.ApplicationXPemFileProducer = runtime.TextProducer()
+	api.ApplicationTimestampQueryConsumer = runtime.TextConsumer()
```
bobcallaway: This may work, but wondering if `runtime.ByteStreamProducer()` and `runtime.ByteStreamConsumer()` are more appropriate

pkg/api/error.go
```diff
@@ -32,17 +32,18 @@ import (
...
+	entryAlreadyExists                = "An equivalent entry already exists in the transparency log with UUID %v"
+	firstSizeLessThanLastSize         = "firstSize(%d) must be less than lastSize(%d)"
+	malformedUUID                     = "UUID must be a 64-character hexadecimal string"
+	malformedPublicKey                = "Public key provided could not be parsed"
+	failedToGenerateCanonicalKey      = "Error generating canonicalized public key"
+	redisUnexpectedResult             = "Unexpected result from searching index"
+	lastSizeGreaterThanKnown          = "The tree size requested(%d) was greater than what is currently observable(%d)"
+	signingError                      = "Error signing"
+	failedToGenerateTimestampResponse = "Error generating timestamp response"
 )
```
bobcallaway, line 49: Need to add case below for the new timestamp methods in `handleRekorAPIError`

pkg/api/timestamp.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/pkg/errors"
+	"github.com/sigstore/rekor/pkg/generated/restapi/operations/timestamp"
+	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/util"
+)
+
+func RequestFromRekor(ctx context.Context, requestBytes []byte) ([]byte, error) {
+	req, err := util.ParseTimestampRequest(requestBytes)
```
bobcallaway: perhaps we should bump this out into the calling method and return HTTP 400 (Bad Request) with this error message if we can't parse the input?

asraa:
> Are we expecting to fill out any of the other fields (accuracy, TSA) in our response:

All the fields left out are optional per spec:
* TSA: this is a TODO, but I'll go ahead to make that the subject name from the timestamping cert: `If present, it MUST correspond to one of the
   subject names included in the certificate that is to be used to
   verify the token.`
* Accuracy: this would just be the accuracy of the gentime, it's optional and could be retrieved from other means. i've seen some people use anything from a few seconds to 30 seconds. 

I could add extensions to the CLI as a follow up, they will be there if you curl a request with one

cmd/rekor-cli/app/timestamp.go
asraa: Yeah, I thought about that, can do in a followup -- here's my question about it: I want to server to accept a TSQ, so this would mean the client would need to fetch the URL which I wanted to avoid. Instead, I allowed the artifact-hash which is basically the same.

Otherwise, I could add this as an option for the JSON formatted and expose a URL field that the server can fetch.

cmd/rekor-cli/app/timestamp.go
asraa: done. was my mistake in thinking it was openssl default to false, but actually I was just using -no_nonce

asraa:
TSA added, format binary added.

```
Using configuration from /usr/lib/ssl/openssl.cnf
Status info:
Status: Granted.
Status description: unspecified
Failure info: unspecified

TST info:
Version: 1
Policy OID: tsa_policy1
Hash Algorithm: sha256
Message data:
    0000 - e3 b0 c4 42 98 fc 1c 14-9a fb f4 c8 99 6f b9 24   ...B.........o.$
    0010 - 27 ae 41 e4 64 9b 93 4c-a4 95 99 1b 78 52 b8 55   '.A.d..L....xR.U
Serial number: 0x1E1A44AD6D97A9380D28012A
Time stamp: May 19 00:15:24 2021 GMT
Accuracy: unspecified
Ordering: no
Nonce: 0xD6D202F8D818130314D1066A
TSA: DirName:/C=US/ST=/L=San Francisco/street=Golden Gate Bridge/postalCode=94016/O=Rekor Test
Extensions:
```

we should add accuracy when we have a better sense of a good accuracy value in real time instances.

pkg/api/timestamp.go
```diff
@@ -0,0 +1,78 @@
...
+		return handleRekorAPIError(params, http.StatusInternalServerError, err, failedToGenerateTimestampResponse)
+	}
+
+	// TODO: Upload to transparency log and add entry UUID to location header.
+	log.Logger.Errorf("generated OK")
+	return timestamp.NewGetTimestampResponseOK().WithPayload(ioutil.NopCloser(bytes.NewReader(resp)))
+}
+
+func GetTimestampCertChainHandler(params timestamp.GetTimestampCertChainParams) middleware.Responder {
+	return timestamp.NewGetTimestampCertChainOK().WithPayload(api.certChainPem)
```
bobcallaway, line 80: If we don't have certChainPem configured, are we OK with sending back an empty response with HTTP 200 OK, or should we send back a 404 to indicate that we don't have it?

asraa: yeah, i like that. done!

asraa:
Thanks @bobcallaway! Is there anything else? Let me know, will start staging follow-up

dlorenc (approved):

bobcallaway (approved):

pkg/signer/memory.go
```diff
@@ -33,6 +39,71 @@ type Memory struct {
...
+			Locality:      []string{"San Francisco"},
+			StreetAddress: []string{"Golden Gate Bridge"},
+			PostalCode:    []string{"94016"},
+		},
+		IPAddresses:  []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback},
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().AddDate(10, 0, 0),
+		SubjectKeyId: []byte{1, 2, 3, 4, 6},
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageTimeStamping},
+		KeyUsage:     x509.KeyUsageContentCommitment,
```
mstjohns, line 89: This is an unusual keyUsage value to pair with the id-kp-timestamping.  Typical keyUsage is digitalSignature, not contentCommitment.  If you want the most interoperability, I'd suggest adding digitalSignature here.  See for example https://www.digistamp.com/support/repository-of-timestamp-public-key-certificates.  This came up as an issue in OpenJDK and these certificates would be rejected without a change to the JDK.

dlorenc: Sorry just noticed this comment here. We should switch that over!

asraa: can do!

asraa: Update in https://github.com/sigstore/rekor/pull/504

PR #294, opened 2021-04-30
Add Log ID to LogEntry field

bobcallaway:
Since the signed entry timestamp will be able to prove insertion into the log, we will need to include the log ID in order for the index to be meaningful.

### DIFF
--- a/cmd/rekor-cli/app/get.go
+++ b/cmd/rekor-cli/app/get.go
@@ -40,10 +40,12 @@ type getCmdOutput struct {
 	LogIndex       int
 	IntegratedTime int64
 	UUID           string
+	LogID          string
 }
 
 func (g *getCmdOutput) String() string {
-	s := fmt.Sprintf("Index: %d\n", g.LogIndex)
+	s := fmt.Sprintf("LogID: %v\n", g.LogID)
+	s += fmt.Sprintf("Index: %d\n", g.LogIndex)
 	dt := time.Unix(g.IntegratedTime, 0).UTC().Format(time.RFC3339)
 	s += fmt.Sprintf("IntegratedTime: %s\n", dt)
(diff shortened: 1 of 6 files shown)


dlorenc:
> Since the signed entry timestamp will be able to prove insertion into the log, we will need to include the log ID in order for the index to be meaningful.

I'm not quite sure I follow. Is the LogID essentially just the random int we get from createtree? I don't think we expose it anywhere else.

dlorenc:
> I'm not quite sure I follow. Is the LogID essentially just the random int we get from createtree? I don't think we expose it anywhere else.

Maybe the log URL would be better? Sort of the way LE refers to them: https://letsencrypt.org/docs/ct-logs/

Although I guess they do also publish the log id.

bobcallaway:
> > I'm not quite sure I follow. Is the LogID essentially just the random int we get from createtree? I don't think we expose it anywhere else.
> 
> Maybe the log URL would be better? Sort of the way LE refers to them: https://letsencrypt.org/docs/ct-logs/
> 
> Although I guess they do also publish the log id.

OK, I think I see how I got confused here... https://tools.ietf.org/html/rfc6962#section-3.2 refers to Log ID but instead of it being the actual trillian log ID, it is the sha256 hash of the log's signing public key. The index of an entry in an RFC6962 compliant CT log isn't part of the SCT at all. 

Right now, the Rekor SET (signed entry timestamp) covers the index, actual entry, and integrated timestamp. I think we should ignore my earlier idea of adding the trillian log ID into the SET, but instead follow the practice from RFC6962 and include the hash of the log's public key. This allows for easier identification of the appropriate public key required to validate the signature on the SET.

dlorenc (approved):

PR #312, opened 2021-05-21
Use default trillian_log_server ports

axelsimon:
Switch away from using rpc_endpoint=8091 and back to 8090 which is trillian_log_server's default.

This simplifies the commands to start trillian log_server, reducing it down to:
`trillian_log_server --logtostderr ...`
and avoids diverging from upstream when there is no need.

This change also updates the corresponding docker-compose and k8s configs.

A corresponding documentation update is in the works.

Signed-off-by: axelsimon <github@axelsimon.net>


axelsimon:
And here is the corresponding sigstore website PR: https://github.com/sigstore/sigstore.github.io/pull/64

bobcallaway (approved):
lgtm

PR #321, opened 2021-06-02
Attestation Transparency!

dlorenc:
This adds an "Attestation" method to the entry interface. Entries can
return an attestation that they would like to store.

The attestations are currently stored in GCS, but it supports any blob store.
The feature is turned off with a flag, and we can set a max size as well.

TODO: 

- [x] End to End tests
- [x] Size limits for attestations
- [x] Figure out storage key (UUID of entry or of attestation? How to verify attestation hash? We store only the "statement", not the overall envelope. Maybe store both?)

### DIFF
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -23,6 +23,7 @@ import (
 	"crypto/ecdsa"
 	"crypto/sha256"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"encoding/pem"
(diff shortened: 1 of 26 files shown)


tests/e2e_test.go
```diff
@@ -344,8 +350,32 @@ func TestIntoto(t *testing.T) {
 	// If we do it twice, it should already exist
 	out := runCli(t, "upload", "--artifact", attestationPath, "--type", "intoto", "--public-key", pubKeyPath)
 	outputContains(t, out, "Created entry at")
+	uuid := getUUIDFromUploadOutput(t, out)
+
+	// The atteestation should be stored at /var/run/attestations/$uuid
+	cmd := exec.Command("docker", "run", "-v", "/var/run/attestations:/var/run/attestations", "alpine", "cat", "/var/run/attestations/"+uuid)
```
bobcallaway, line 356: could this be simplified to just read directly from `/var/run/attestations/` on the host?

bobcallaway (approved):
one question, but LGTM otherwise

tests/e2e_test.go
dlorenc: Not always, especially in CI you're typically running in a container too with the docker socket mounted in.

It's the same roughly for Mac or anywhere that you have docker in a VM.

PR #322, opened 2021-06-03
Convert STH to checkpoint format

### LINKED ISSUE #313: Use new STH format from Trillian
Trillian is working on a new format for signed-tree-heads that should allow witnesses/monitors/auditors to run against any transparency log without needing to know or care what's in the log. We should switch to this when it's ready.

https://github.com/google/trillian-examples/tree/master/formats/log

cc @alcutter

bobcallaway:
This switches away from sending the (now deprecated) Trillian LogRootV1
format over to the checkpoint format documented at
https://github.com/google/trillian-examples/tree/master/formats/log

Fixes: #313


pkg/util/checkpoint.go
```diff
@@ -0,0 +1,325 @@
...
+func (s SignedCheckpoint) Verify(public crypto.PublicKey) bool {
+	if len(s.Signatures) == 0 {
+		return false
+	}
+
+	msg, _ := s.Checkpoint.MarshalText()
+	//TODO: generalize this
+	digest := sha256.Sum256(msg)
+
+	var errorArray []error
```
dlorenc: I don't think I follow the errorArray logic here. I'd expect the logic to be something like:

if all signatures verify, return true.
if any signature does not verify, return false.

The len(errorArray) == count check inside the loop has me confused a bit. It looks like it might return on the first successful verification.

bobcallaway: Yeah, the error handling can be simplified; There can be multiple signatures over the checkpoint content so if the public key verifies successfully against one of them then we return true.

dlorenc (approved):

pkg/util/checkpoint.go
```diff
@@ -0,0 +1,319 @@
...
+	pubKeyBytes, err := x509.MarshalPKIXPublicKey(signer.Public())
+	if err != nil {
+		return nil, errors.Wrap(err, "marshalling public key")
+	}
+
+	pkSha := sha256.Sum256(pubKeyBytes)
+
+	signature := note.Signature{
+		Name:   identity,
+		Hash:   binary.BigEndian.Uint32(pkSha[:]),
```
asraa, line 143: is it important that the algorithm is encoded in this hash? 
it is included in the sumdb note code
https://cs.opensource.google/go/x/mod/+/refs/tags/v0.4.2:sumdb/note/note.go;drc=ed3ec21bb8e252814c380df79a80f366440ddb2d;l=326

PR #324, opened 2021-06-04
Add new type timestamp for RFC 3161 timestamp responses

loosebazooka:
Should be ready to go

example usage on local build:
```
rekor-cli upload --rekor_server http://localhost:3000 --type rfc3161 --artifact test.tsr
```


pkg/types/timestamp/v0.0.1/timestamp_v0_0_1_schema.json
```diff
@@ -0,0 +1,48 @@
...
+        "tsr": {
+            "description": "Information about the tsr file associated with the entry",
+            "type": "object",
+            "properties" : {
+                "content": {
+                    "description": "Specifies the tsr file content inline within the document",
+                    "type": "string",
+                    "format": "byte"
+                },
+                "hash": {
```
loosebazooka: it's not clear to me that storing the hash of the tsr has any value? We can just generate it and use it for indexing? But I dunno, so it's in here for now.

dlorenc:
> * [ ]  perhaps also allow urls (?)

I think we can skip URLs for small content like these.

pkg/types/timestamp/v0.0.1/entry.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, err
+	}
+
+	return bytes, nil
+}
+
+// Validate performs cross-field validation for fields in object
+func (v V001Entry) Validate() error {
+
+	// TODO: validate the tsr content here?
```
dlorenc: @asraa any idea what would make sense to validate here?

pkg/types/timestamp/v0.0.1/timestamp_v0_0_1_schema.json
dlorenc: Yeah I think indexing is the main use case.

pkg/types/timestamp/v0.0.1/entry.go
asraa: Hi hi! So probably you want to validate the ContentInfoSignedData contained in the TimeStampResp. I would:
(1) asn1 unmarshal the response into a [pkcs9.TimeStampResp](https://pkg.go.dev/github.com/sassoftware/relic@v7.1.0+incompatible/lib/pkcs9#TimeStampResp) resp
(2) If the resp.Status.Status is not Granted/GrantedWithMods (0 or 1) then there should not be any TimeStampToken attached and maybe we allow people uploading failed but well-formed timestamps. So stop here
(3) Otherwise, we have a TimeStampToken (type [pkcs7.ContentInfoSignedData](https://pkg.go.dev/github.com/sassoftware/relic@v7.1.0+incompatible/lib/pkcs7#ContentInfoSignedData)). Verify that the ContentType is a signed data OID attribute `{1, 2, 840, 113549, 1, 7, 2}`. 
(4) Verify the pkcs7.SignedData attribute (TimeStampToken.Content) in the token using this [Verify](https://github.com/sassoftware/relic/blob/v7.1.0/lib/pkcs7/verify.go#L44) method, with no external content, but skipDigests false.
(5) Final note: we can't really verify the cert chain with a given root as suggested in the doc string from (4), so skip that. We've validated the sig/content.

pkg/types/timestamp/v0.0.1/timestamp_v0_0_1_schema.json
```diff
@@ -0,0 +1,48 @@
...
+    "properties": {
+        "tsr": {
+            "description": "Information about the tsr file associated with the entry",
+            "type": "object",
+            "properties" : {
+                "content": {
+                    "description": "Specifies the tsr file content inline within the document",
+                    "type": "string",
+                    "format": "byte"
+                },
```
bobcallaway: I would also extract and store in a field called `publicKey` the x509 certificate that verified the signature of the timestamp. It's returned from https://github.com/sassoftware/relic/blob/a47b9184570dc9f78acc224b69e812a1b3ee0eb9/lib/pkcs7/verify.go#L93

bobcallaway (commented):
anticipating that we might add the JSON-formatted TSR, or roughtime support in the future... should we pick a more specific name than `timestamp` for the type?

pkg/types/timestamp/v0.0.1/entry.go
loosebazooka: Thanks! This makes sense, for now though, I failed if status is not: Granted/GrantedWithMods

loosebazooka:
- Updated to remove hash schema
- Added validation to the canonicalize function
- ~~Not yet done: Updating the name to something more specific as @bobcallaway suggested~~
  - rename type to `rfc3161`

pkg/types/timestamp/v0.0.1/timestamp_v0_0_1_schema.json
loosebazooka: From my understanding, all of that can be extracted from the tsr file at verification time? Is it just to make it easier on the final verifier?
I get that this is gross though:

`$ openssl ts -reply -in test.tsr -token_out | openssl pkcs7 -inform der -print_certs | sed -n '/-----BEGIN/,/-----END/p' > chain.pem`

loosebazooka:
added in extracting and storing publicKey

pkg/types/timestamp/v0.0.1/timestamp_v0_0_1_schema.json
loosebazooka: I'm storing it now.

asraa (approved):
Yay!!! LGTM The timestamp related stuff looks good to me!

pkg/types/rfc3161/v0.0.1/entry.go
```diff
@@ -0,0 +1,189 @@
...
+
+	// ExtraData is copied through unfiltered
+	canonicalEntry.ExtraData = v.Rfc3161Obj.ExtraData
+
+	// wrap in valid object with kind and apiVersion set
+	ref3161Obj := models.Rfc3161{}
+	ref3161Obj.APIVersion = swag.String(APIVERSION)
+	ref3161Obj.Spec = &canonicalEntry
+
+	bytes, err := json.Marshal(&ref3161Obj)
```
bobcallaway: nit: `return json.Marshal(&ref3161Obj)` instead of the `if err != nil` below

pkg/types/rfc3161/v0.0.1/entry.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, err
+	}
+
+	if v.Rfc3161Obj.Tsr == nil || v.Rfc3161Obj.Tsr.Content == nil {
+		return nil, fmt.Errorf("tsr content is nil")
+	}
+
+	content := *v.Rfc3161Obj.Tsr.Content
+
+	// validate the content
```
bobcallaway: should we move this validation logic into `Validate` and store the cert in a struct field?

pkg/types/rfc3161/v0.0.1/rfc3161_v0_0_1_schema.json
```diff
@@ -0,0 +1,39 @@
...
+    "$id": "http://rekor.sigstore.dev/types/timestamp/timestamp_v0_0_1_schema.json",
+    "title": "Timestamp v0.0.1 Schema",
+    "description": "Schema for RFC3161 entries",
+    "type": "object",
+    "properties": {
+        "tsr": {
+            "description": "Information about the tsr file associated with the entry",
+            "type": "object",
+            "properties" : {
+                "content": {
```
bobcallaway, line 12: just curious - why is the hash not included in the `tsr` object (as in the other types)? if we don't end up persisting the content of the TSR into the tlog, including the hash ensures we don't run into collisions when we insert multiple TSRs signed with the same public key, and improves usability for people searching the log for a given artifact

pkg/types/rfc3161/v0.0.1/entry.go
```diff
@@ -0,0 +1,189 @@
...
+	sig, err := tsr.TimeStampToken.Content.Verify(nil, false)
+	if err != nil {
+		return nil, fmt.Errorf("Tsr verification error: %v", err)
+	}
+
+	cert := strfmt.Base64(sig.Certificate.Raw)
+
+	canonicalEntry := models.Rfc3161V001Schema{
+		Tsr: &models.Rfc3161V001SchemaTsr{
+			Content: &content,
```
bobcallaway: should we be persisting the entire TSR content into the log (this deviates from the other types, where we do not store content itself)? I could see extracting the message imprint (content digest and algorithm)... @dlorenc what are your thoughts here?

pkg/types/rfc3161/v0.0.1/rfc3161_v0_0_1_schema.json
loosebazooka: no particular reason, I had it in there, but it didn't seem to add any value? I though the TSR content should be enough to differentiate between the entries? Is that not true?

see: https://github.com/sigstore/rekor/pull/324#pullrequestreview-676663135

pkg/types/rfc3161/v0.0.1/entry.go
loosebazooka: Looking through the code (I'm new), it doesn't seem like validate is called before "writing" an entry to log? I only see it in SearchLogQueryHandler? Am I missing something?

pkg/types/rfc3161/v0.0.1/rfc3161_v0_0_1_schema.json
bobcallaway: If we don't store the content (given precedent from other types) then you'd just have the public key left here which runs into the issues of potential collisions for a TSA signing two or more timestamps which would appear to the tlog as duplicate entries. Since we're not storing a separate signature value in the type, we need something to provide uniqueness.

pkg/types/rfc3161/v0.0.1/entry.go
bobcallaway: In RPM/JAR/Rekord, we call it in `FetchExternalEntities`; in in-toto it's called at the end of `Unmarshal`; perhaps it was an errant delete since `FetchExternalEntities` is a no-op in this type?

pkg/types/rfc3161/v0.0.1/entry.go
dlorenc: I **think** it would be safe to store the whole thing, but I honestly don't understand the format well enough. @asraa what do you think? It's definitely simplest to just store it all...

asraa: i don't think we should just store the message imprint, since the TSR contains that *plus* the time added (and two of the same messages with different timestamps would be equivalent)

if you wanted a minimal portion of the "signed" TSR content, that would be the TSTInfo (that's the thing the server generates with the matching message imprint and time). Even that, that's not the actual bytes the server signs, but it should be unique if considered with the public key.

asraa: the actual signed bytes would be the AuthenticatedAttributes. https://github.com/sigstore/rekor/blob/c4b1a3941339dd67f6419ba368142b70ecfc5750/pkg/util/rfc3161.go#L230

that would be unique, it contains the hash of the TSTInfo and the hash of the signing cert.

pkg/types/rfc3161/v0.0.1/rfc3161_v0_0_1_schema.json
asraa: if you were to opt to store a hash, i would either store a hash of the entire TSR or just the signed content like I mentioned below https://github.com/sigstore/rekor/pull/324#discussion_r649953085

both should provide uniqueness, even across the same TSA, same content, same time (TSA creates a serial number included in the signed content that shouldn't be repeated).

pkg/types/rfc3161/v0.0.1/entry.go
loosebazooka: Yeah the combined typing for sending data and storing is little confusing. This type probably doesn't need to fetch external entries, but I could do validation in `Unmarshal`.

pkg/types/rfc3161/v0.0.1/entry.go
loosebazooka: I don't have a strong opinion here, I can change to store anything. Should it just be TSTInfo, Hash and the Cert/PublicKey?

dlorenc: Let's store the entire thing.

pkg/types/rfc3161/v0.0.1/entry.go
loosebazooka: I also find the model of types exporting all their methods and then reusing them internally to be somewhat confusing. Is an external system orchestrating all the actions or is the struct supposed to do everything itself? It really doesn't help define what the scope of responsibility is.

for example in Jar: https://github.com/sigstore/rekor/blob/main/pkg/types/jar/v0.0.1/entry.go#L210 calling FetchExternalEntities internally really makes it murky what the point of the method even is?

loosebazooka: Anyway, I think was able to follow the current convention and store only hash+cert. Lemme know if that is sufficient? Or if we should extract a signature as well.

pkg/types/rfc3161/v0.0.1/rfc3161_v0_0_1_schema.json
loosebazooka: Using hash of the entire file

pkg/types/rfc3161/rfc3161_schema.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://rekor.sigstore.dev/types/rfc3161/rfc3161_schema.json",
+    "title": "Timstamp Schema",
```
bobcallaway: nit: if you end up pushing something else to this PR again, `s/Timstamp/Timestamp`

pkg/types/rfc3161/v0.0.1/entry.go
bobcallaway: after thinking about this as well, I think we should store the entire thing.

bobcallaway (commented):
I think the feedback on the EntryImpl interface is valid and we could clean that up in a separate PR for all types. 

I think we can move forward with this (notwithstanding the change to store the entire TSR... apologies for the churn on that point but given that it was a change in behavior it is good to make sure we're all in agreement there).

loosebazooka:
It's all good I'll update with storing the full tsr.

loosebazooka:
As a precaution, rekor will fail if size is too big? My sample tsr are 2K. Should we set an upper limit? like 20K? Would this unduly burden the log/users? Maybe something smaller/bigger?

asraa:
> Should we set an upper limit? like 20K? Would this unduly burden the log/users? Maybe something smaller/bigger?

This sounds fine to me to start -- the message to timestamp is signed as a hash, so potentially the only fields I see growing are the cert sizes/chain sizes. ~O(10)kb sounds okay to me

pkg/types/rfc3161/v0.0.1/entry.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, fmt.Errorf("tsr content is nil")
+	}
+
+	if len(*v.Rfc3161Obj.Tsr.Content) > (10 * 1024) {
+		return nil, fmt.Errorf("tsr content exceed max limit (10kB)")
+	}
+
+	content := *v.Rfc3161Obj.Tsr.Content
+
+	// validate the content
```
loosebazooka: I think it feels weird to validate content here, but the Validate function itself seems like its for crossfield validation, not content validation?

bobcallaway: I think it makes sense to move this to Validate, and turn `FetchExternalEntities` into a no-op for this type.

loosebazooka: kk, moved into validate, and called validate in Unmarshal

pkg/types/rfc3161/v0.0.1/entry.go
```diff
@@ -0,0 +1,178 @@
...
+	if data == nil {
+		return errors.New("missing tsr data")
+	}
+
+	content := *data.Content
+	if len(content) == 0 {
+		return errors.New("'content' must be specified for data")
+	}
+
+	if len(*v.Rfc3161Obj.Tsr.Content) > (10 * 1024) {
```
bobcallaway: this will check the length of the b64 string which isn't the same as checking the length of the byte array `b`... might want to slide this down after the decoding.

bobcallaway (commented):
LGTM other than the spelling comment and size computation mentioned above

dlorenc (approved):
LGTM!

loosebazooka:
What a roller coaster 🎢

PR #336, opened 2021-06-17
Upload generated timestamps

asraa:
Thanks to @loosebazooka rekor stores RFC 3161 timestaps, so this PR adds create log entries whenever a timestamp is generated.

The output now looks like
```
$ ./rekor timestamp --artifact file.txt --rekor_server http://localhost:3000
Artifact timestamped at 2021-06-17 19:24:00 +0000 UTC
Wrote timestamp response to response.tsr
Created entry at index 104, available at: http://localhost:3000/api/v1/log/entries/333bf5fd9d3d09057d3b6e5bd7a909940e1889e449c27bc72b7fa1d80958cc7e
```


cmd/rekor-cli/app/timestamp.go
```diff
@@ -238,7 +245,10 @@ var timestampCmd = &cobra.Command{
 
 		// TODO: Add log index after support for uploading to transparency log is added.
```
loosebazooka: Remove this TODO now?

pkg/api/entries.go
```diff
@@ -127,26 +127,22 @@ func GetLogEntryByIndexHandler(params entries.GetLogEntryByIndexParams) middlewa
 	return entries.NewGetLogEntryByIndexOK().WithPayload(logEntry)
 }
 
-// CreateLogEntryHandler creates new entry into log
-func CreateLogEntryHandler(params entries.CreateLogEntryParams) middleware.Responder {
-	ctx := params.HTTPRequest.Context()
-	httpReq := params.HTTPRequest
+func createLogEntry(ctx context.Context, params entries.CreateLogEntryParams) (models.LogEntry, int, string, error) {
```
loosebazooka: It is a little weird that we return and error and a error message? But I dunno, maybe that was the only sane way to refactor this and it doesn't matter since it's internal?

pkg/api/entries.go
```diff
@@ -211,6 +207,23 @@ func CreateLogEntryHandler(params entries.CreateLogEntryParams) middleware.Respo
...
+}
+
+// CreateLogEntryHandler creates new entry into log
+func CreateLogEntryHandler(params entries.CreateLogEntryParams) middleware.Responder {
+	ctx := params.HTTPRequest.Context()
+	httpReq := params.HTTPRequest
+
+	logEntry, code, message, err := createLogEntry(ctx, params)
+	if err != nil {
+		return handleRekorAPIError(params, code, err, message)
```
loosebazooka: I think I might prefer the order of these params to be the same for both, 
like if it could be `..., code, err, message` to match the inputs to `handleRekorApiError`

pkg/api/timestamp.go
```diff
@@ -62,15 +82,32 @@ func TimestampResponseHandler(params timestamp.GetTimestampResponseParams) middl
...
-	// TODO: Upload to transparency log and add entry UUID to location header.
-	log.Logger.Errorf("generated OK")
-	return timestamp.NewGetTimestampResponseOK().WithPayload(ioutil.NopCloser(bytes.NewReader(resp)))
+	// Upload to transparency log and add entry UUID to location header.
+	entryParams := entries.CreateLogEntryParams{
+		HTTPRequest:   httpReq,
+		ProposedEntry: createRFC3161FromBytes(resp),
+	}
+
+	logEntry, code, msg, err := createLogEntry(ctx, entryParams)
```
loosebazooka: I vaguely remember something about this being behind a flag like `--log` or something, not sure if we care about that anymore.

pkg/api/entries.go
loosebazooka: Maybe this doens't make sense. Feel free to ignore.

loosebazooka:
LGTM.

pkg/api/timestamp.go
asraa: Yeah, I think looking back on it we probably want all timestamps to be logged (even if only using as a timestamping machine) just for sake of transparency. But I'm open to creating a default upload flag and making it configurable per req.

pkg/api/entries.go
asraa: I might be able to return a logEntry, middleware.Responser, and err. I wasn't entirely sure how to chain together middleware.Responders, it was a little opaque to me

asraa: That seems pretty OK

pkg/api/entries.go
asraa: resolved by way of removing :)

asraa:
/request review from @bobcallaway. would love input on this "middleware chain" (really not sure if this was the best way)

openapi.yaml
```diff
@@ -249,6 +249,17 @@ paths:
           schema:
```
bobcallaway, line 248: We should probably change the response code from `200` to `201` here since we've actually created a new object.

openapi.yaml
```diff
@@ -249,6 +249,17 @@ paths:
...
             type: string
             format: binary
+          headers:
+            ETag:
+              type: string
+              description: UUID of log entry
```
bobcallaway: maybe be a bit more verbose here (and in other description strings below) "UUID of log entry made for TSR"

pkg/api/timestamp.go
```diff
@@ -62,15 +82,35 @@ func TimestampResponseHandler(params timestamp.GetTimestampResponseParams) middl
...
+	paramReq.URL.Path = "/api/v1/log/entries"
+	entryParams := entries.CreateLogEntryParams{
+		HTTPRequest:   &paramReq,
+		ProposedEntry: createRFC3161FromBytes(resp),
+	}
+
+	// If middleware is returned, this indicates an error.
+	logEntry, middleware := createLogEntry(entryParams)
+	if middleware != nil {
+		return middleware
```
bobcallaway, line 85: I think this should be fine (returning an error responder from CreateLogEntry) but curious if you have artificially forced an error path here? (they both implement the same interface, so it should be fine... but since that is all part of the generated code from go-openapi I'm not 100% sure they aren't doing something funky).

pkg/api/timestamp.go
```diff
@@ -62,15 +82,35 @@ func TimestampResponseHandler(params timestamp.GetTimestampResponseParams) middl
...
 	if err != nil {
 		return handleRekorAPIError(params, http.StatusInternalServerError, err, failedToGenerateTimestampResponse)
 	}
 
-	// TODO: Upload to transparency log and add entry UUID to location header.
-	log.Logger.Errorf("generated OK")
-	return timestamp.NewGetTimestampResponseOK().WithPayload(ioutil.NopCloser(bytes.NewReader(resp)))
+	// Upload to transparency log and add entry UUID to location header.
+	paramReq := *httpReq
+	paramReq.URL.Path = "/api/v1/log/entries"
```
bobcallaway: I might use the URL builder from the code generation here VS hard coding this. I think it would look like:

```go
cleURL := entries.CreateLogEntryURL{}
paramReq.URL = cleURL.Must(cleURL.Build())
```

pkg/api/timestamp.go
```diff
@@ -45,6 +49,22 @@ func RequestFromRekor(ctx context.Context, req pkcs9.TimeStampReq) ([]byte, erro
 	return body, nil
 }
 
+func createRFC3161FromBytes(resp []byte) models.ProposedEntry {
```
bobcallaway: This feels a little out of place here (vs being implemented inside the actual type and being called here), but this would certainly work.

bobcallaway (commented):
a couple thoughts, but overall looking pretty good.

pkg/api/timestamp.go
asraa: Hmmm yeah, I really didn't understand too well how repurposing the generated code would work. What do you mean by "forced an error path"? 
The error path shouldn't hit if the log entry was created. In some cases an error could legitimately happen (some internal server error). 
The other option is creating our own error struct and change the interface a little bit so that `handlRekorAPIError` can take that input

bobcallaway: Sorry I was unintentionally obtuse. "forced an error path" meant "temporarily change line 102 to `middleware == nil`" which would cause the rare behavior to always occur. If this causes the server to panic, then you'd know that while the compiler is happy, the generated code somewhere really isn't and we need to investigate a different approach.

asraa: AH! That's a great idea, I was looking through what errors I could generate easily earlier to test the path, forcing it is great. Thanks!

asraa: Seemed to go through fine, I forced a signing error to happen, returning non-nil middleware, and I ended up with
```
2021/06/22 13:14:19 [POST /api/v1/timestamp][500] getTimestampResponse default  &{Code:500 Message:Error signing}
``` 
server-side looks fine as well

pkg/api/timestamp.go
asraa: addressed, now I share code between pflags and here

pkg/api/timestamp.go
bobcallaway: ok, then this part LGTM then.

bobcallaway (approved):
LGTM

PR #339, opened 2021-06-23
Improve separation between type implementations and CLI code

### LINKED ISSUE #273: refactor CLI type handling
**Description**

Right now we have a bit of spaghetti code when we handle command line arguments to the various CLI commands. Some types (RPM, JAR) do not take certain args because the signature and/or public key needed are embedded within the file, versus being detached and processed separately.

It would be good to generalize this and make this explicitly declared within the type implementation, and re-implement the existing code in pflags.go to leverage the type-specific combinations.

bobcallaway:
This is the first of a few patches to try to clean up & simplify the CLI implementation, as well as make responsibility more clear for the pluggable type interface.

- Adds support for specifying specific versions of pluggable types (using `--type type(:version_string)?` command line flag)
- Use go-playground/validator instead of custom validation functions where possible
- Have CLI help enumerate supported types (and versions), as well as PKI formats
- Remove all type-specific code from CLI (except for RFC3161 timestamping)
- Move calls to create format-agnostic entry objects (i.e. `models.ProposedEntry`) into versioned type implementations instead of within CLI
- Remove duplicated flag checking code in various CLI functions
- Added `SupportedVersions()` method to `TypeImpl` for enumerating the versions of pluggable types supported
- Added `DefaultVersion()` method to `TypeImpl` so if a type is requested without a version, the default version will be used

Fixes #273


cmd/rekor-cli/app/pflags.go
```diff
@@ -16,790 +16,223 @@
...
-	return "fileOrURLFlag"
-}
-
-type typeFlag struct {
-	value string
-}
-
-func (t *typeFlag) Type() string {
-	return "typeFormat"
+// IsURL returns true if the supplied value is a valid URL and false otherwise
```
loosebazooka: `IsURL` -> `isURL`
it does not appear like this needs to be exported?

cmd/rekor-cli/app/root.go
```diff
@@ -16,16 +16,24 @@
...
 	"strings"
 
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+
+	"github.com/sigstore/rekor/pkg/log"
+
+	// these imports are to call the packages' init methods
```
loosebazooka, line 29: This feels strange to me, but maybe I'm not familiar enough with the flags/command library in use here? If this has to know about its sub commands -- should it not just be responsible for initializing them? the whole init back and forth here feels like it's hard to trace things in code.

pkg/log/log.go
```diff
@@ -54,8 +54,9 @@ var CliLogger = createCliLogger()
 func createCliLogger() *zap.SugaredLogger {
 	cfg := zap.NewDevelopmentConfig()
 	cfg.EncoderConfig.TimeKey = ""
-	cfg.EncoderConfig.LevelKey = ""
+	// cfg.EncoderConfig.LevelKey = ""
```
loosebazooka: do we want to leave this in for something? can it have a context comment?

pkg/types/entries.go
```diff
@@ -36,11 +37,23 @@ type EntryImpl interface {
 	Unmarshal(e models.ProposedEntry) error           // unmarshal the abstract entry into the specific struct for this versioned type
 	Validate() error                                  // performs any cross-field validation that is not expressed in the OpenAPI spec
 	Attestation() (string, []byte)
+	CreateFromPFlags(context.Context, ArtifactProperties) (models.ProposedEntry, error)
```
loosebazooka: nit: Should this name still reference `PFlags`, since you've already done the translate into `ArtifactProperties`?

cmd/rekor-cli/app/pflag_groups.go
```diff
@@ -0,0 +1,175 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/sigstore/rekor/pkg/pki"
+	"github.com/sigstore/rekor/pkg/types"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// addFlagToCmd adds the
```
loosebazooka: incomplete comment?

loosebazooka (commented):
looks good to me, just some personal confusion about how init should be used in this cli library.

cmd/rekor-cli/app/root.go
bobcallaway: before this change, the types weren't imported because we were directly creating objects of each type. Since the object creation code now lies in `pkg/types/...`, we need to trigger the type's `init()` methods to fire to populate the type & version maps accordingly, so that's why we have the imports here.

bobcallaway:
> looks good to me, just some personal confusion about how init should be used in this cli library.

@loosebazooka I addressed the points you raised - appreciate any other feedback you might have.

cmd/rekor-cli/app/root.go
loosebazooka: yeah I think my problem is that we need to trigger the `init()` functions explicitly like without ever actually using anything from the package explicitly. The pattern just feels off to me, but I dunno, I don't have any better/alternative solutions.

loosebazooka (approved):

dlorenc (approved):

dlorenc:
Sorry for the delay - I got confused about whether we were moving this one forward or the other one.

Issue #315, opened 2021-05-25
New Timestamp Format


lukehinds:
Issue to track new timestamping efforts

* will update shortly.

dlorenc:
Let's use this issue to track the new timestamp format @asraa is working on!

asraa:
Hi!

So for context, originally I hashed out and was working on a `TimestampNote` inspired by `SignedNote` that I was generalizing from Bob's impl of the `SignedCheckpoint`. The TimestampNote would follow the roughtime protocol and include a blind, radius, etc. 

However, I think we may as well use Roughtime structs following https://github.com/cloudflare/roughtime so that when Rekor is fully production and we have a time source locked down, we would be ready to formally integrate it into the ecosystem and add a PR here https://github.com/cloudflare/roughtime/tree/1fe56bcbcfd4dd1ed52614ccc73a29c9b062275d#ecosystem-guidelines

Hayden-IO:
There are newer issues, closing as outdated

PR #347, opened 2021-06-30
Generalize SignedCheckpoint to take arbitrary Notes

asraa:
Generalizes SignedCheckpoint to a SignedNote:

SignedNote contains a textual representation of a `Note` and `note.Signatures`.
SignedCheckpoint uses a Checkpoint as the note, and marshals it as before into the `SignedNote.Note`.

This lays the groundwork for a SignedTimestamp which would contain a marshalled TimestampResponse as the Note.

For https://github.com/sigstore/rekor/issues/315


asraa:
Merged main -- just looking for a signal on whether I'm missing something I can do in Go about specializing a type in an interface implementation

i.e. can i specify that a RekorSTH, which is a type with a SignedNote (containing a Note interface), always has a Checkpoint Note?

Maybe the best way is to write methods for RekorSTH to get the Hash/Size

asraa:
@dekkagaijin if you have a few minutes could you please take a look? mostly on the comment above!

bobcallaway:
Looking at `golang.org/x/mod/sumdb/note` it seems like they just leave it as a `string`. 

> Maybe the best way is to write methods for RekorSTH to get the Hash/Size

That would probably be where I'd start.

asraa:
> Looking at golang.org/x/mod/sumdb/note it seems like they just leave it as a string.

Storing it as a text representation worked okay, won't need special methods to access Checkpoint data now.

pkg/util/signed_note.go
```diff
@@ -0,0 +1,207 @@
...
+type SignedNote struct {
+	// Textual representation of a note to sign.
+	Note string
+	// Signatures are one or more signature lines covering the payload
+	Signatures []note.Signature
+}
+
+// Sign adds a signature to a SignedCheckpoint object
+// The signature is added to the signature array as well as being directly returned to the caller
+func (s *SignedNote) Sign(identity string, signer crypto.Signer, opts crypto.SignerOpts) (*note.Signature, error) {
```
dekkagaijin: can replace this with `signature.Signer` from s/s to eliminate the hashing boilerplate

pkg/util/signed_note.go
```diff
@@ -0,0 +1,207 @@
...
+		Base64: base64.StdEncoding.EncodeToString(sig),
+	}
+
+	s.Signatures = append(s.Signatures, signature)
+	return &signature, nil
+}
+
+// Verify checks that one of the signatures can be successfully verified using
+// the supplied public key
+func (s SignedNote) Verify(public crypto.PublicKey) bool {
```
dekkagaijin: ditto this with `signature.Verifier`

asraa:
Moved over to signature.Signer/Verifier. Sorry it took a second! Vacation + some bugs.

pkg/util/signed_note.go
```diff
@@ -0,0 +1,190 @@
...
+
+	msg := []byte(s.Note)
+	digest := sha256.Sum256(msg)
+
+	for _, s := range s.Signatures {
+		sigBytes, err := base64.StdEncoding.DecodeString(s.Base64)
+		if err != nil {
+			return false
+		}
+		pk, _ := verifier.PublicKey()
```
dlorenc: Do we need to check this error?

pkg/util/signed_note.go
```diff
@@ -0,0 +1,190 @@
...
+		opts := []signature.VerifyOption{}
+		switch pk.(type) {
+		case *rsa.PublicKey:
```
dlorenc: there's no fallthrough by default, I think you can do a multi-line case statement or use the explicit fallthrough operator.

pkg/util/signed_note.go
```diff
@@ -0,0 +1,190 @@
...
+
+// String returns the String representation of the SignedNote
+func (s SignedNote) String() string {
+	var b strings.Builder
+	b.WriteString(s.Note)
+	b.WriteRune('\n')
+	for _, sig := range s.Signatures {
+		var hbuf [4]byte
+		binary.BigEndian.PutUint32(hbuf[:], sig.Hash)
+		sigBytes, _ := base64.StdEncoding.DecodeString(sig.Base64)
```
dlorenc, line 121: Do we need to check this error?

dlorenc (commented):
A few nits.

pkg/util/signed_note.go
asraa: I should, I was coding too transparently into the Verifiers (it never returns an error https://github.com/sigstore/sigstore/blob/a9aeb218f4d17318e29c0df3315c0496edf25a1a/pkg/signature/ecdsa.go#L138)

pkg/util/signed_note.go
asraa: Unmarshalling checks this, and creating a signature will guarantee this to be OK, so I think there's an implicit guarantee that any SignedNote is well-formed in  that respect. On the other hand, then I would probably make the field private. WDYT? (This is existing)

dlorenc (approved):

pkg/util/signed_note.go
dlorenc: Yeah this is probably fine.

PR #349, opened 2021-07-01
Move GetRekorClient into util directory

bobcallaway:
Since other sigstore projects are using GetRekorClient, this moves it
into the pkg/util directory so that the number of dependencies this
brings with it can be minimized.


cpanato (approved):

cpanato (commented):
why not use a pkg called `client` instead of `util`?

dlorenc (approved):

PR #354, opened 2021-07-06
Added Helm type

sabre1041:
Added a new type for [Helm](https://helm.sh/)


dlorenc (commented):
Just took a quick look, this looks awesome!

A couple high level questions, I'll dig in more tomorrow.

Is the helm provenance spec "firm"? Could you provide a link to the spec somewhere in the README?

Does helm require pgp, or can it user other key types?

sabre1041:
@dlorenc 

1. Documentation to the Helm Provenance file can be found [here](https://helm.sh/docs/topics/provenance/#the-provenance-file). Happy to add to the documentation. Please advise where you feel it should be placed
2. Sadly Helm supports only pgp. However, the approach is to implement/integrate with the current set of capabilities in Helm and then work with the Helm community about the possibilities of incorporating other key types which could then be downstreamed here

pkg/types/helm/providence.go
```diff
@@ -0,0 +1,132 @@
...
+	"bytes"
+	"io"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
+
+	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/armor"
+	"golang.org/x/crypto/openpgp/clearsign"
```
lukehinds: Are there any non `x/crypto` pgp modules we could use. x/crypto is going to get rejected by the FIPS folks. 

Gah, I can't wait to kill pgp with fire (but I totally understand you're working with what you have here).

lukehinds (approved):

pkg/types/helm/providence.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+type Provedance struct {
```
bobcallaway: s/Provedance/Provenance

pkg/types/helm/providence.go
```diff
@@ -0,0 +1,132 @@
...
+
+		return parts[1], nil
+	}
+
+	// Return error if no keys found
+	return "", errors.New("No checksums found")
+
+}
+
+func (p *Provedance) VerifySignature(keyring openpgp.KeyRing, armoredSignatureBody io.Reader) error {
```
bobcallaway: Is there a reason you didn't reuse the `pki/pgp` implementation? I would strongly prefer to try to reuse existing modules rather than directly importing openpgp here.

pkg/types/helm/v0.0.1/entry.go
```diff
@@ -0,0 +1,359 @@
...
+	"github.com/sigstore/rekor/pkg/types/helm"
+	"github.com/sigstore/rekor/pkg/util"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	APIVERSION = "0.0.1"
+)
+
+var (
```
bobcallaway: This is already defined as a const as `HelmV001SchemaChartHashAlgorithmSha256`

pkg/types/helm/providence.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Error occurred parsing SumCollection")
+	}
+
+	p.SumCollection = sc
+
+	return nil
+}
+
+func (p *Provedance) GetChartHash() (string, error) {
```
bobcallaway: Since you're using this as the search index, is there a reason to not just return the entire string (with `sha256` prefix)?

pkg/types/helm/v0.0.1/entry.go
```diff
@@ -0,0 +1,359 @@
...
+	canonicalEntry.Chart = &models.HelmV001SchemaChart{}
+
+	chartHash, err := v.provenanceObj.GetChartHash()
+
+	if err != nil {
+		return nil, err
+	}
+
+	canonicalEntry.Chart.Hash = &models.HelmV001SchemaChartHash{}
+	canonicalEntry.Chart.Hash.Algorithm = &SHA256
```
bobcallaway: use const `models.HelmV001SchemaChartHashAlgorithmSha256`

pkg/types/helm/providence.go
```diff
@@ -0,0 +1,132 @@
...
+
+type SumCollection struct {
+	Files  map[string]string `json:"files"`
+	Images map[string]string `json:"images,omitempty"`
+}
+
+func (p *Provedance) Unmarshal(reader io.Reader) error {
+
+	buf := &bytes.Buffer{}
+	_, err := buf.ReadFrom(reader)
```
bobcallaway: nit: you could obtain the count of bytes read here rather than doing `len(rawProvenanceFile)` below

pkg/types/helm/providence.go
Foxboron: You want to use `github.com/ProtonMail/go-crypto/openpgp` instead.

https://github.com/ProtonMail/go-crypto/

It should be noted that it has less issues then the `x/crypto` library and should work in all cases for parsing gpg keys.

bobcallaway (commented):
looking good, a couple thoughts

pkg/types/helm/providence.go
lukehinds: I recall @bobcallaway used the protonmail module as well in a different type, @sabre1041 can you look to using protonmail/gro-crypto instead?

pkg/types/helm/providence.go
sabre1041: I can return the entire string. but if someone is just running `sha256sum`, it will only return the hash without the `sha256`. I took a look at some other examples in the project and at least one other type is not including the prefix.

pkg/types/helm/v0.0.1/entry.go
sabre1041: Resolved

pkg/types/helm/v0.0.1/entry.go
sabre1041: Resolved

pkg/types/helm/providence.go
sabre1041: Resolved

pkg/types/helm/providence.go
sabre1041: Resolved using the `pki/pgp` `Verify` function

pkg/types/helm/providence.go
sabre1041: Resolved

pkg/types/helm/providence.go
sabre1041: @lukehinds Looked into `github.com/ProtonMail/go-crypto/openpgp` and it does provide what would be needed for the Helm type. Are you also looking at replacing other instances of `x/crypto` that is in `pki/pgp` too?

lukehinds: @sabre1041 , yes very much. I will lodge an issue right now. Also I just noticed I approved when I meant to just comment, please hang out for an approval from mr @bobcallaway as he is the types man.

sabre1041: Do we want to separate that work into a separate PR per the issue you will create? I'd be happy to also take a look at that as well

lukehinds: the others are in ./pkg/pki/ssh ./pkg/pki/pgp, I will create an issue now, it would be preferable to get this patch to merge in with this already in good order.

bobcallaway (approved):
LGTM - I wouldn't block on the openpgp port

bacongobbler:
> Is the helm provenance spec "firm"?

Yes. It's worked just as it did back in Helm 2.0.

> 2\. Sadly Helm supports only pgp. However, the approach is to implement/integrate with the current set of capabilities in Helm and then work with the Helm community about the possibilities of incorporating other key types which could then be downstreamed here

Yep. PRs/suggestions are always welcome :)

https://github.com/helm/helm/blob/main/CONTRIBUTING.md#proposing-an-idea

technosophos:
YAY! Glad to see this. As @bacongobbler suggests, we are happy to see proposals for other key types. It's been hard enough to get people to use OpenPGP even though it is broadly supported, compatible with Keybase and GitHub, etc.

lukehinds:
> YAY! Glad to see this. As @bacongobbler suggests, we are happy to see proposals for other key types. It's been hard enough to get people to use OpenPGP even though it is broadly supported, compatible with Keybase and GitHub, etc.

Thanks, it might be worth a prototype using keyless (OpenID based ephemeral keys). The low adoption of PGP in projects is very common. It seems to be typically never more than 10% as folks seem adverse to managing the keys.

sabre1041:
@lukehinds @bobcallaway @dlorenc anything else before integrating this in?

pkg/types/helm/v0.0.1/helm_v0_0_1_schema.json
```diff
@@ -0,0 +1,118 @@
...
+            ]
+        },
+        "extraData": {
+            "description": "Arbitrary content to be included in the verifiable entry in the transparency log",
+            "type": "object",
+            "additionalProperties": true
+        }
+    },
+    "required": [
+        "signature",
```
bobcallaway: should this be `publicKey` not `signature`?

bobcallaway (commented):
LGTM other than the one check in the JSON schema

sabre1041:
> LGTM other than the one check in the JSON schema

@bobcallaway JSON schema changes made

bobcallaway (approved):

PR #356, opened 2021-07-06
Implement git push certificates

Foxboron:
This is an initial attempt at implementing support for git push
certificates in rekor.

Git push certificates allows the committer to sign the commits they
pushed towards some remote. These are important to log as they can work
as a canary if there are malicious commits on the git forge server. An
example of this previously is the git.php.net compromise along with
several attempts at compromising kernel.org in the past.

kernel.org already implements a transparency logs with all git push
operations, but not all of them are signed[0].

There is also a POC implementation of a monitor for this transparency
log[1], but it would be better served to be on shared infrastructure
with other projects instead of being reinvented on a per project basis.

The current implementation is a veryvery naive json schema over the
documented push certificate mostly representing everything as strings
with minimal validation over the schema[2]. Hopefully this can be
better validated with parsing support in something like go-git in the
future[3]

It currently missing some frontend tool for the parsing itself, but the
idea is to hook it up with a post-receive hook from the repository which
would feed rekor[4].

[0]: https://korg.docs.kernel.org/gitolite/transparency-log.html
[1]: https://tlog.linderud.dev/
[2]: https://git-scm.com/docs/pack-protocol/2.29.0#_push_certificate
[3]: https://github.com/go-git/go-git
[4]: https://git-scm.com/docs/git-receive-pack/2.11.4#_post_receive_hook

Signed-off-by: Morten Linderud <morten@linderud.pw>


bobcallaway:
Thanks for the PR @Foxboron - is there a reason why the full public key of the pusher cannot be submitted at the same time with the rest of the fields?

Also, do you have a sample push certificate that you could include as part of the e2e test suite?

Foxboron:
> is there a reason why the full public key of the pusher cannot be submitted at the same time with the rest of the fields?

It's not present on the server. Git only receives the signature it self and there is no way for git to retrieve the public key. I'm not sure if there are ways around this if we where to envision a rekor deployment scenario which would not be gnupg based.

The question is if we should try retrieve the certificate out-of-bounds? Either through `hkp` and some public keyservers or through WKD.

>Also, do you have a sample push certificate that you could include as part of the e2e test suite?

That requires a complete parser implementation. I wasn't sure where to put it. Does it belong in `pkg/types/gitpush` or another place?

It would largely be an adapted version from my project which I wanted to try upstream to `go-git`.

https://github.com/Foxboron/kernel.org-git-verifier/blob/master/object.go
https://github.com/Foxboron/kernel.org-git-verifier/blob/master/object_test.go

PR #391, opened 2021-07-27
makefile: add rule to download and set swagger and make rule to build rekor-cli for cross platform

cpanato:
- this adds a way to automatically download the swagger in the specific version we care and don't mess with go modules for the project.
- also add rule to cross build rekor-cli for other linux/darwin/windows

### DIFF
--- a/scripts/go_install.sh
+++ b/scripts/go_install.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


scripts/go_install.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Kubernetes Authors.
```
bobcallaway: s/Kubernetes/Sigstore

lukehinds (approved):
LGTM (with @bobcallaway review nit fixed)

scripts/go_install.sh
cpanato: done! thanks for the review

bobcallaway:
how does this affect GitHub actions CI to ensure it is building with the same swagger version as defined in the Makefile?

cpanato:
@bobcallaway forgot CI, i will check, but ci will use this one we define in the makefile

cpanato:
@bobcallaway updated CI

bobcallaway (approved):

PR #397, opened 2021-07-28
Add a SignedTimestampNote

asraa:

Adds a new SignedTimestampNote object. A TimestampNote (a "timestamp query") can be signed by Rekor to produce a SignedTimestampNote (a "timestamp response).

I separated this PR out from two other major changes:
* Adding a storage time for a TimestampNote
* Adding the client/server code for generating a timestamp note

Because I'm stuck with OpenApi issues on the latter. Will commit the storage one separately.

I also realized the timestamp_note.go got accidentally committed in an earlier PR! So this updates this and adds tests.

TimestampNotes contain the following info:
```
type TimestampNote struct {
	// Ecosystem is the ecosystem/version string
	Ecosystem string
	// MessageImprint is the hash of the message to timestamp, of the form sha256:<sha>
	MessageImprint string
	// Nonce is a short random  bytes to prove response freshness
	Nonce []byte
	// Time is the timestamp to imprint on the message
	Time time.Time
	// Radius is the time in microseconds used to indicate certainty
	Radius int64
	// CertChainRef is a reference URL to the valid timestamping cert chain used to sign the response
	CertChainRef *url.URL
	// OtherContent is any additional data to be included in the signed payload; each element is assumed to be one line
	OtherContent []string
}
```

A Signed one looks like:
```
Timestamp Note v0
sha256:e4ba5cbd251c98e6cd1c23f126a3b81d8d8328abc95387229850952b3ef9f904
ew==
2021-07-26T00:00:00Z
123
http://localhost:3000/api/v1/timestamp/certchain

\u2014 name pOhM+S/mYjEYtQsOF4lL8o/dR+nbjoz5Cvg/n486KIismpVq0s4wxBaakmryI7zThjWAqRUyECPL3WSEcVDEBQ==
```

### DIFF
--- a/pkg/util/timestamp_note.go
+++ b/pkg/util/timestamp_note.go
@@ -64,7 +64,7 @@ func (t TimestampNote) MarshalText() ([]byte, error) {
 	return []byte(t.String()), nil
 }
 
-// validateSHA256Value ensures that the supplied string matches the following format:
+// validateMessgeImprint ensures that the supplied string matches the following format:
 // [sha256:]<64 hexadecimal characters>
 // where [sha256:] is optional
 func validateMessgeImprint(v string) error {
@@ -94,7 +94,7 @@ func validateMessgeImprint(v string) error {
 // The supplied data is expected to begin with the following 6 lines of text,
 // each followed by a newline:
 // <ecosystem/version string>
-// <base64 representation of message hash>
+// <message hash of the format sha256:$SHA>
 // <base64 representation of the nonce>
 // <RFC 3339 representation of the time>
 // <decimal representation of radius>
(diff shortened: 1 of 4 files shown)


pkg/util/timestamp_note.go
```diff
@@ -69,13 +64,37 @@ func (t TimestampNote) MarshalText() ([]byte, error) {
...
+	split := strings.SplitN(v, ":", 2)
+	switch len(split) {
+	case 1:
+		hash = split[0]
+	case 2:
+		prefix = split[0]
+		hash = split[1]
+	}
+
+	s := struct {
```
dlorenc: this is fine, would it make sense to expose/export this struct somewhere else though?

asraa: Yeah, might not be the best file name but done, and shared code with the same validation in the client

dlorenc (approved):

bobcallaway (approved):
lgtm

PR #402, opened 2021-07-30
Use an in memory timestamping key

asraa:

* Generates an in-memory timestamping key. The rekor signer creates a timestamping certificate for this key.
* If a timestamp_chain is loaded in, then this is used for the timestamping chain. It's leaf must be a certificate for the rekor signer.
* Otherwise, a self-signed in-memory CA is created.

Tested with both in-memory signer (no chain) and a timestamp_chain loaded in authorizing the rekor signer as a CA to issue timestamping certificates


asraa:
@dlorenc

pkg/signer/memory.go
```diff
@@ -39,33 +41,53 @@ type Memory struct {
...
+
+	if len(chain) == 0 {
+		// Generate a self-signed root CA.
+		ca := &x509.Certificate{
+			SerialNumber: big.NewInt(2019),
+			Subject: pkix.Name{
+				Organization:  []string{"in-memory root CA"},
+				Country:       []string{"US"},
+				Province:      []string{""},
+				Locality:      []string{"San Francisco"},
```
dlorenc: Should we fill this in with slightly more real info?

pkg/signer/memory.go
```diff
@@ -97,11 +114,33 @@ func NewTimestampingCertWithSelfSignedCA(pub crypto.PublicKey) ([]*x509.Certific
...
+	root := x509.NewCertPool()
+	root.AddCert(chain[len(chain)-1])
+	intermediates := x509.NewCertPool()
+	for _, intermediate := range chain[:len(chain)-1] {
+		intermediates.AddCert(intermediate)
+	}
+	verifyOptions := x509.VerifyOptions{
+		Roots:         root,
+		Intermediates: intermediates,
+		KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
```
dlorenc: can we constrain to timestamps here?

asraa: Oops, done! I don't know why I missed it in https://cs.opensource.google/go/go/+/go1.16.6:src/crypto/x509/x509.go;l=570

pkg/signer/memory.go
asraa: I just removed the fake US thing and called it a rekor in-memory CA

asraa:
Thanks for the check! Confirmed it work in the cases:
* Load up a gcpkms signer and a timestamping cert chain -> uses the signer to create a TSA cert, and verifies with the chain provided (the cert chain must authorize the signer to be a CA)
* Load up a gcpkms signer and no timestamping cert chain -> creates an in-memory root CA which creates the TSA cert
* Load up an in-memory signer and no timestamping cert chain -> creates an in-memory root CA which creates the TSA cert
* If you load up an in-memory signer and a timestamping cert chain -> you will fail on start up because it won't verify the chain

dlorenc (approved):

PR #410, opened 2021-08-11
Update snippets to use `console` type for snippets

tiziano88:
`shell` or `bash` expects the entire snippet to be a bash script, while `console` highlights only lines prefixed with `$` as bash.

### DIFF
--- a/types.md
+++ b/types.md
@@ -6,7 +6,7 @@ This documentation contains information on how to sign and upload data in differ
 
 Create a keypair with something like:
 
-```shell
+```console
 $ minisign -G
 Please enter a password to protect the secret key.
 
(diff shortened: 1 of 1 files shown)


dlorenc:
Whoa! I had no idea this was a thing!

tiziano88:
Me neither until 5 minutes ago! I was reading the readme and thought that there must be a proper way of highlighting those snippets, and finally ended up on https://github.com/github/linguist/blob/7efa7b3380ad0221c499f9ef1ca84c4dfcc0b299/lib/linguist/languages.yml#L5594-L5605 , which works!

BTW the DCO is complaining that my commit (which I authored from the web UI) is not signed..... I guess I need to do a proper one from my laptop and force push.

I must say I don't really understand what signing commits actually buys you though.

cpanato:
@tiziano88 thanks for the PR
yes, you need to do `git commit --amend --no-edit --signoff` and push again

tiziano88:
Done now.

dlorenc (approved):

PR #426, opened 2021-08-29
Drop 404 errors down to a warning.

dlorenc:
We get flooded with scapers so it makes it hard to find real errors in our logs.

### DIFF
--- a/pkg/generated/restapi/configure_rekor_server.go
+++ b/pkg/generated/restapi/configure_rekor_server.go
@@ -200,7 +200,11 @@ func cacheForever(handler http.Handler) http.Handler {
 }
 
 func logAndServeError(w http.ResponseWriter, r *http.Request, err error) {
-	log.RequestIDLogger(r).Error(err)
+	if apiErr, ok := err.(errors.Error); ok && apiErr.Code() == http.StatusNotFound {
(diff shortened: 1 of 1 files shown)


pkg/generated/restapi/configure_rekor_server.go
```diff
@@ -200,7 +200,11 @@ func cacheForever(handler http.Handler) http.Handler {
 }
 
 func logAndServeError(w http.ResponseWriter, r *http.Request, err error) {
-	log.RequestIDLogger(r).Error(err)
+	if apiErr, ok := err.(errors.Error); ok && apiErr.Code() == http.StatusNotFound {
+		log.RequestIDLogger(r).Warn(err)
+	} else {
+		log.RequestIDLogger(r).Error(err)
```
bobcallaway, line 206: do you want to collapse the debug statement below into the else clause?

otherwise LGTM

bobcallaway (approved):

pkg/generated/restapi/configure_rekor_server.go
dlorenc: I left it that way on purpose, debug logs seem fine all the time, especially if they get more info out of the request.

PR #433, opened 2021-09-14
Cleanup some initialization/flag parsing in rekor-server.

dlorenc:
This is in preparation for supporting multiple logIDs (for sharding).

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	"github.com/google/trillian"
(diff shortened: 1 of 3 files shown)


dlorenc:
Up next:

- support starting the server with multiple log IDs, instead of just one.
  - this will probably look like a config file containing a map of index ranges to log ids.
  - only one can be "active"
- support parsing the UUID during GET calls to determine which log to query (with defaulting for the 0 case here)

bobcallaway (approved):
lgtm

PR #435, opened 2021-09-14
Add a flag to start specifying log index ranges for virtual indices.

dlorenc:
This is part of the temporal sharding work. The flag is not
hooked up anywhere yet.

### DIFF
--- a/cmd/rekor-server/app/flags_test.go
+++ b/cmd/rekor-server/app/flags_test.go
@@ -1,0 +1,135 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestLogRanges_Set(t *testing.T) {
+	tests := []struct {
+		name string
+		arg  string
+		want []LogRange
+	}{
+		{
+			name: "one, no length",
+			arg:  "1234",
+			want: []LogRange{
+				{
+					TreeID:     1234,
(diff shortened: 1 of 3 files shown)


cmd/rekor-server/app/root.go
```diff
@@ -60,6 +61,8 @@ func init() {
 	rootCmd.PersistentFlags().String("trillian_log_server.address", "127.0.0.1", "Trillian log server address")
 	rootCmd.PersistentFlags().Uint16("trillian_log_server.port", 8090, "Trillian log server port")
 	rootCmd.PersistentFlags().Uint("trillian_log_server.tlog_id", 0, "Trillian tree id")
+	rootCmd.PersistentFlags().Var(&logRangeMap, "trillian_log_server.log_id_ranges", "list of tree ids and ranges")
```
bobcallaway: Is there any need to specify a different log server IP/port for each tree ID here, or is a single trillian instance sufficient for multiple trees?

cmd/rekor-server/app/flags_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+func TestLogRanges_ResolveVirtualIndex(t *testing.T) {
+	lrs := LogRanges{
+		Ranges: []LogRange{
+			{TreeId: 1, TreeLength: 17},
+			{TreeId: 2, TreeLength: 1},
+			{TreeId: 3, TreeLength: 100},
+			{TreeId: 4},
```
bobcallaway: This makes sense here (for all new additions to go into `TreeId: 4`), but should we extend the command line arg to have a trailing value to match this? If the server panics and restarts, we don't want it to create a new tree but if the trillian instance reports back more than one tree that isn't specified in the list we wouldn't be able to disambiguate which one to use.

cmd/rekor-server/app/root.go
```suggestion
	rootCmd.PersistentFlags().Var(&logRangeMap, "trillian_log_server.log_id_ranges", "ordered list of tree ids and ranges")
```

bobcallaway (commented):
makes sense so far

cmd/rekor-server/app/root.go
dlorenc: I could see us needing to eventually, but can't think of a real reason yet. WDYT? One trillian instance can handle many trees.

cmd/rekor-server/app/flags_test.go
dlorenc: +1, I thought about this but skipped it. We should. I'll fix it up here.

lukehinds (approved):
LGTM if Bob's points are resolved.

PR #461, opened 2021-10-19
Fix a bug in x509 certificate handling.

dlorenc:
The CryptoPubKey function only returned the key value, but we should
retrieve it from the cert if set. This fixes the rest of #918.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


mattmoor (approved):

mattmoor:
The test is consistent with the structure here, so if this works then hopefully we're good 🤞 
https://github.com/sigstore/cosign/blob/7d2d51d6153e008607bfd5c4c1b5f0889d9d943e/pkg/cosign/tlog.go#L100

cpanato (commented):
lgtm

bobcallaway (approved):

PR #464, opened 2021-10-23
Reproducible builds with trimpath

naveensrinivasan:
Included -trimpath flag for able to have reproducible builds.


cpanato (approved):
Thanks!! I thought we already have that

PR #474, opened 2021-10-31
Correct Helm index keys

sabre1041:
Corrections to how Helm index keys are stored

### DIFF
--- a/pkg/types/helm/provenance_test.go
+++ b/pkg/types/helm/provenance_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/pki"
 	"github.com/sigstore/rekor/pkg/pki/pgp"
 )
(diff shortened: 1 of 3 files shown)


pkg/types/helm/provenance_test.go
```diff
@@ -37,12 +38,16 @@ func TestProvenance(t *testing.T) {
...
 	}
 
-	checksum, err := provenance.GetChartHash()
+	algorithm, checksum, err := provenance.GetChartAlgorithmHash()
 
 	if err != nil {
 		t.Fatalf("Error retrieving chart hash: %v", err)
 	}
 
+	if models.AlpineV001SchemaPackageHashAlgorithmSha256 != algorithm {
```
dlorenc: You should be able to find this in helm somewhere instead of alpine...

dlorenc (commented):
Just the one nit about which constant to use.

pkg/types/helm/provenance_test.go
sabre1041: @dlorenc whoops. good catch. updated

pkg/types/helm/provenance_test.go
```diff
@@ -37,12 +38,16 @@ func TestProvenance(t *testing.T) {
...
 
-	checksum, err := provenance.GetChartHash()
+	algorithm, checksum, err := provenance.GetChartAlgorithmHash()
 
 	if err != nil {
 		t.Fatalf("Error retrieving chart hash: %v", err)
 	}
 
+	if models.HelmV001SchemaChartHashAlgorithmSha256 != algorithm {
+		t.Fatalf("Unexpected checksum algorithm. Expected %s, found %s", models.AlpineV001SchemaPackageHashAlgorithmSha256, algorithm)
```
bobcallaway:
```suggestion
		t.Fatalf("Unexpected checksum algorithm. Expected %s, found %s", models. HelmV001SchemaChartHashAlgorithmSha256, algorithm)
```

sabre1041: sigh. updated. good old copy/paste similar to the the prior issue

bobcallaway (approved):

PR #485, opened 2021-11-09
Included pprof for profiling the application.

naveensrinivasan:
Included pprof for profiling the application.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -62,7 +62,7 @@ func init() {
 
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.rekor-server.yaml)")
 	rootCmd.PersistentFlags().StringVar(&logType, "log_type", "dev", "logger type to use (dev/prod)")
-	rootCmd.PersistentFlags().BoolVar(&enablePprof, "enable_pprof", false, "enable pprof for profiling")
+	rootCmd.PersistentFlags().BoolVar(&enablePprof, "enable_pprof", false, "enable pprof for profiling on port 6060")
 
 	rootCmd.PersistentFlags().String("trillian_log_server.address", "127.0.0.1", "Trillian log server address")
 	rootCmd.PersistentFlags().Uint16("trillian_log_server.port", 8090, "Trillian log server port")
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -104,7 +104,7 @@ func init() {
 			mux.HandleFunc("/debug/pprof/{action}", pprof.Index)
 			mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
 
-			if err := http.ListenAndServe(":6060", mux); err != nil {
+			if err := http.ListenAndServe(":6060", mux); err != nil && err != http.ErrServerClosed {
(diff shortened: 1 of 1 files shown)


dlorenc:
Is this safe to run in production? Would it make more sense to manually register the handlers via a flag?

naveensrinivasan:
> Is this safe to run in production? Would it make more sense to manually register the handlers via a flag?

Looks like we can still run it in Prod. https://stackoverflow.com/a/64057856/19407

I agree, I have included a flag that will enable pprof.  Would this work?

cmd/rekor-server/app/root.go
```diff
@@ -88,6 +94,21 @@ func init() {
...
+	// Enable pprof
+	if enablePprof {
+		go func() {
+			mux := http.NewServeMux()
+
+			mux.HandleFunc("/debug/pprof/", pprof.Index)
+			mux.HandleFunc("/debug/pprof/{action}", pprof.Index)
+			mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+
+			if err := http.ListenAndServe(":6060", mux); err != nil {
```
bobcallaway: Not that I want to have an abundance of command line arguments, but should we make the IP and port configurable?

naveensrinivasan: I am on the fence. The command-line flags are growing a lot. It shouldn't become like this https://cortexmetrics.io/docs/configuration/arguments/ which is a project by itself IMO.

cmd/rekor-server/app/root.go
```diff
@@ -57,6 +62,7 @@ func init() {
 
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.rekor-server.yaml)")
 	rootCmd.PersistentFlags().StringVar(&logType, "log_type", "dev", "logger type to use (dev/prod)")
+	rootCmd.PersistentFlags().BoolVar(&enablePprof, "enable_pprof", false, "enable pprof for profiling")
```
```suggestion
	rootCmd.PersistentFlags().BoolVar(&enablePprof, "enable_pprof", false, "enable pprof for profiling on port 6060")
```

cmd/rekor-server/app/root.go
bobcallaway: we can always add it later if folks request it

cmd/rekor-server/app/root.go
```diff
@@ -88,6 +94,21 @@ func init() {
 	if GitVersion != "devel" {
 		return
 	}
+	log.Logger.Debugf("pprof enabled %v", enablePprof)
+	// Enable pprof
+	if enablePprof {
```
bobcallaway, line 99: do we need to check if the user has requested port 6060 for the main rekor endpoint and give a helpful error message?

cmd/rekor-server/app/root.go
```suggestion
			if err := http.ListenAndServe(":6060", mux); err != nil && err != http.ErrServerClosed {
```

bobcallaway (commented):
LGTM minus small nits

cmd/rekor-server/app/root.go
naveensrinivasan: For now, I would pass on this because this flag has to be turned on and the default rekor endpoint port has to be set to `6060`. The odds of that happening are less IMO. Also, we have included the docs in the command line it could help someone debug. 

Thoughts?

dlorenc (approved):

PR #490, opened 2021-11-15
Refactor Trillian - Decoupling With Interface

naveensrinivasan:
Refactoring trillian to be interface based which is easier to mock for testing. 
With the present implementation, it is impossible to test other than running a trillian server.
It is hard to get code coverage metrics with an external server.


naveensrinivasan:
-

PR #509, opened 2021-11-23
Fix fuzz go.sum issue

naveensrinivasan:


dlorenc (approved):

dlorenc:
Doesn't seem to have worked...

naveensrinivasan:
> Doesn't seem to have worked...

It is this go bug which is causing this issue https://github.com/golang/go/issues/44129

naveensrinivasan:
@dlorenc Hopefully this should have fixed the problem. It is a hack until we move to go 1.17.

PR #514, opened 2021-11-30
types/intoto: decode attestation before returning

### LINKED ISSUE #511: In-toto Attestations are double base64 encoded
Currently the in-toto rekor type returns the base64 encoded payload from the DSSE payload.  The attestation rekor then stores to the bucket is that same base64 encoded payload.  This attestation is then base64 encoded again in responses from the rekor API.

I currently have a patch that alters the in-toto type to base64 decode the attestation prior to storing it which fixes this, but causes backward compatibility concerns about users who may currently expect the double-encoded behavior.  I can open a PR with this patch but would like if the maintainers would weigh in.  Perhaps making the change in a 0.0.2 version of the in-toto rekor type is the right answer?

mikhailswift:
#### Summary

In-toto attestations returned from the API are currently base64 encoded
twice.  This is due to the attestation being stored on disk encoded as
base64 and the API encoding responses as base64.  This commit decodes
the attestation prior to being saved to disk.

Signed-off-by: Mikhail Swift <mswift@mswift.dev>

#### Ticket Link

Fixes #511 

#### Release Note

```release-note
in-toto attestations are base64 decoded before being stored to disk
in-toto attestations are only base64 encoded once in api responses
```

### DIFF
--- a/pkg/types/intoto/v0.0.1/entry.go
+++ b/pkg/types/intoto/v0.0.1/entry.go
@@ -199,11 +199,19 @@ func (v *V001Entry) validate() error {
 }
 
 func (v *V001Entry) Attestation() (string, []byte) {
-	if len(v.env.Payload) > viper.GetInt("max_attestation_size") {
+	decodedLen := base64.StdEncoding.DecodedLen(len(v.env.Payload))
+	if decodedLen > viper.GetInt("max_attestation_size") {
 		log.Logger.Infof("Skipping attestation storage, size %d is greater than max %d", len(v.env.Payload), viper.GetInt("max_attestation_size"))
 		return "", nil
 	}
-	return v.env.PayloadType, []byte(v.env.Payload)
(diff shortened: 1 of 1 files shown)


mikhailswift:
Since this changes existing API behavior there are some backward compatibility considerations to take into account.  Perhaps it is best to make an in-toto v0.0.2 type with this change? Maintainer input would definitely be appreciated on the best course forward.

bobcallaway:
This patch looks fine, but we need to update these two things as well:

1. The end-to-end test that is failing in CI
2. the `rekor-cli get` command that prints the entry should pretty-print the attestation JSON (as is done with the body) vs the base64 blob

mikhailswift:
I was mistaken about this, it doesn't seem to be doubly base64 encoded after working on this again this evening.  Apologies for that, not sure what happened on my end.

PR #515, opened 2021-11-30
UUID parsing to include pre-pended shard ID

### LINKED ISSUE #487: Support UUID --> shard ID parsing
Goal: support parsing the UUID during GET calls to determine which log to query (with defaulting for the 0 case here). This would require two changes:

- [x] During upload of an entry, the UUID created should include a parseable log ID. For example, if the active log has log ID 163, the UUID might look like: `163-d2f305428d7c222d7b77f56453dd4b6e6851752ecacc78e5992779c8f9b61dd9`, where the number before the `-` separator is the log ID.

- [ ] Then during handling of a GET request for a particular UUID, the log ID can be easily determined and the request routed accordingly.

@dlorenc @bobcallaway Does this sound right to you based on the design?

lkatalin:
#### Summary
This is a draft! Feedback welcome, including overall design considerations. I've got some to-dos that I am still sorting out as listed here and in code comments. It's not many lines of code, but I did a fair amount of experimentation to get to this point as it's my first "real" Rekor PR. : )

Here is how it's working currently:

Screenshot from 2021-11-30 13-01-40

#### Ticket Link

Fixes #487 

#### Release Note

```release-note

This would change the format of the UUID that people use on the CLI and pre-pend the 64-digit hex UUID with 
a 6-digit human-readable decimal shard ID, like so: 
`ssssss-uuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu`
where `s` is a digit of the shard ID and `u` is a digit of the old UUID / hex-encoded trillian merkle leaf 
hash.
```

- [x] TODO: Allow for backwards-compatibility by letting people search for the old 64-digit UUIDs and internally prepending those with the current shard ID. Is there a way to allow two different valid lengths for a parameter in `pflags.go` and `openapi.yaml`? I'm looking into this.

### DIFF
--- a/cmd/rekor-cli/app/get.go
+++ b/cmd/rekor-cli/app/get.go
@@ -30,6 +30,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/sigstore/rekor/cmd/rekor-cli/app/format"
+	"github.com/sigstore/rekor/cmd/rekor-cli/app/sharding"
 	"github.com/sigstore/rekor/pkg/client"
 	"github.com/sigstore/rekor/pkg/generated/client/entries"
 	"github.com/sigstore/rekor/pkg/generated/models"
@@ -84,6 +85,12 @@ var getCmd = &cobra.Command{
 		}
 
 		logIndex := viper.GetString("log-index")
+		uuid := viper.GetString("uuid")
+
+		if logIndex == "" && uuid == "" {
+			return nil, errors.New("either --uuid or --log-index must be specified")
+		}
+
 		if logIndex != "" {
 			params := entries.NewGetLogEntryByIndexParams()
 			params.SetTimeout(viper.GetDuration("timeout"))
@@ -106,11 +113,17 @@ var getCmd = &cobra.Command{
 			}
 		}
 
-		uuid := viper.GetString("uuid")
 		if uuid != "" {
 			params := entries.NewGetLogEntryByUUIDParams()
 			params.SetTimeout(viper.GetDuration("timeout"))
-			params.EntryUUID = uuid
+
+			if len(uuid) == sharding.UUIDLen {
+				params.EntryUUID = uuid
+			} else if len(uuid) == sharding.FullIDLen {
+				params.EntryUUID = uuid[sharding.FullIDLen-sharding.UUIDLen:]
+			} else {
+				return nil, fmt.Errorf("invalid UUID length: %v", len(uuid))
+			}
 
 			resp, err := rekorClient.Entries.GetLogEntryByUUID(params)
 			if err != nil {
@@ -118,10 +131,7 @@ var getCmd = &cobra.Command{
 			}
 
 			for k, entry := range resp.Payload {
-				if k != uuid {
-					continue
-				}
-
+				// TODO: understand why we don't need k to match params.EntryUUID
 				if verified, err := verifyLogEntry(context.Background(), rekorClient, entry); err != nil || !verified {
 					return nil, fmt.Errorf("unable to verify entry was added to log %w", err)
 				}
@@ -130,7 +140,7 @@ var getCmd = &cobra.Command{
 			}
 		}
 
-		return nil, errors.New("either --uuid or --log-index must be specified")
+		return nil, errors.New("entry not found")
 	}),
 }
 
--- a/cmd/rekor-cli/app/pflags.go
+++ b/cmd/rekor-cli/app/pflags.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sigstore/rekor/cmd/rekor-cli/app/sharding"
 	"github.com/sigstore/rekor/pkg/pki"
 	"github.com/sigstore/rekor/pkg/util"
 
@@ -56,8 +57,8 @@ var pflagValueFuncMap map[FlagType]newPFlagValueFunc
 func initializePFlagMap() {
 	pflagValueFuncMap = map[FlagType]newPFlagValueFunc{
 		uuidFlag: func() pflag.Value {
-			// this corresponds to the merkle leaf hash of entries, which is represented by a 64 character hexadecimal string
-			return valueFactory(uuidFlag, validateString("required,len=64,hexadecimal"), "")
+			// this corresponds to the merkle leaf hash of entries, which is represented by a 64 character hexadecimal string, prepended by the 6-digit shard ID and '-'  separator
+			return valueFactory(uuidFlag, validateFullID, "")
 		},
 		shaFlag: func() pflag.Value {
 			// this validates a valid sha256 checksum which is optionally prefixed with 'sha256:'
@@ -188,6 +189,53 @@ func validateFileOrURL(v string) error {
 	}
 	valGen = pflagValueFuncMap[urlFlag]
 	return valGen().Set(v)
+}
+
+// validateFullID ensures the FullID is composed of a valid shardID, separator character, and UUID
+func validateFullID(v string) error {
+	if len(v) == sharding.FullIDLen {
+
+		// validate separator
+		if string(v[sharding.ShardIDLen]) != sharding.FullIDSeparator {
+			return fmt.Errorf("unexpected separator char in fullID: %v", v[sharding.ShardIDLen])
+		}
+
+		// validate the ShardID
+		shardID := v[:sharding.ShardIDLen]
+		shardIDTag := "required,len=" + fmt.Sprintf("%v", sharding.ShardIDLen) + ",number"
+		shardIDStringValidatorFunc := validateString(shardIDTag)
+		shardIDErr := shardIDStringValidatorFunc(shardID)
+
+		if shardIDErr != nil {
+			return fmt.Errorf("invalid shardID: %v", shardID)
+		}
+
+		// validate the UUID
+		UUID := v[sharding.FullIDLen-sharding.UUIDLen:]
+		UUIDTag := "required,len=" + fmt.Sprintf("%v", sharding.UUIDLen) + ",hexadecimal"
+		UUIDStringValidatorFunc := validateString(UUIDTag)
+		UUIDErr := UUIDStringValidatorFunc(UUID)
+
+		if UUIDErr != nil {
+			return fmt.Errorf("invalid uuid: %v", UUID)
+		}
+
+		return nil
+
+	} else if len(v) == sharding.UUIDLen {
+		// indicates older UUID format is being used without pre-pended shardID
+		UUIDStringValidatorFunc := validateString("required,len=64,hexadecimal")
+		UUIDErr := UUIDStringValidatorFunc(v)
+
+		if UUIDErr != nil {
+			return fmt.Errorf("invalid uuid: %v", v)
+		}
+
+		return nil
+
+	}
+
+	return fmt.Errorf("fullID len error, expected %v but got %v", sharding.FullIDLen, len(v))
 }
 
 // validateLogIndex ensures that the supplied string is a valid log index (integer >= 0)
--- a/cmd/rekor-cli/app/sharding/sharding.go
+++ b/cmd/rekor-cli/app/sharding/sharding.go
@@ -1,0 +1,58 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sharding
+
+import (
+	"fmt"
+)
+
+// A FullID refers to a specific artifact's ID and is made of two components,
+// the ShardID and the UUID, separated by a character, the FullIDSeparator.
+// The ShardID is in human-readable decimal and refers to the specific log or
+// shard where the artifact can be found. The UUID refers to the hex-encoded
+// merkle leaf hash from trillian for the specific artifact.
+
+const ShardIDMax = 999999
+const FullIDSeparator = "-"
+
+const ShardIDLen = 6
+const UUIDLen = 64
+const SeparatorLen = len(FullIDSeparator)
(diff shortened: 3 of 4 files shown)


lkatalin:
@lukehinds @asraa @bobcallaway @dlorenc 
This is what I've got so far... I'm actively working on the remaining TODOs. If you have any thoughts I'm happy to incorporate them. Thanks!

cmd/rekor-cli/app/pflags.go
```diff
@@ -56,8 +57,10 @@ var pflagValueFuncMap map[FlagType]newPFlagValueFunc
 func initializePFlagMap() {
 	pflagValueFuncMap = map[FlagType]newPFlagValueFunc{
 		uuidFlag: func() pflag.Value {
-			// this corresponds to the merkle leaf hash of entries, which is represented by a 64 character hexadecimal string
-			return valueFactory(uuidFlag, validateString("required,len=64,hexadecimal"), "")
+			// this corresponds to the merkle leaf hash of entries, which is represented by a 64 character hexadecimal string, prepended by the 6-digit shard ID and '-'  separator
+			// TODO: does this have to be a literal? can we not use the const to define the len?
+			// TODO: is there a way to validate only the UUID part of this string as hexadecimal? or validate different parts of the string differently? can / should they be split into different flags?
```
asraa: you could make a custom validator! instead of `validateString` you could make `validateShardedUUID` with a custom impl like the ones found here https://github.com/sigstore/rekor/blob/5025a24e2861d1876d67a09cf6bfad0ee76d64c8/cmd/rekor-cli/app/pflags.go#L170

lkatalin: Thank you @asraa ! That's exactly what I want to do. I appreciate the example.

pkg/api/entries.go
```diff
@@ -183,7 +184,9 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	metricNewEntries.Inc()
 
 	queuedLeaf := resp.getAddResult.QueuedLeaf.Leaf
-	uuid := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+	shardID := "000000"
```
bobcallaway: is the thought that the `shardID` would be a base10 integer or base16 (hex) like the rest of the UUID?

lkatalin: @bobcallaway I was going to have it as a base10 integer so that it's more easily human-readable, but I'm definitely open to suggestions here. Is there an advantage to having it in hex?

lkatalin:
I've got an updated version here that
- Fixed the linter errors
- Created a custom validator for the new `fullID`
- Added some custom types / structs in `sharding.go` so the `shardID` is no longer hardcoded in `entries.go`

Outstanding todos:
- [ ] Make sure tests pass
- [x] Don't hardcode the `CurrentShardID` as `0` but instead grab it from / increment it in... somewhere the state is kept? Maybe in a separate PR? I don't understand this part well yet.
- [x] Update custom validator function to allow the old 64-digit UUIDs
- [x] Maybe make the `shardID` hex? Do we want to do this?
- [ ] Add some (possibly dummy at first) routing in `get.go` to route the request to the proper shard. Maybe in a separate PR?

lkatalin:
Also, I am not sure whether it is normal with go code to have `go.mod` and `go.sum` update as part of a PR. I'm not doing this deliberately so if that's not meant to happen I can reset them.

lukehinds:
> Also, I am not sure whether it is normal with go code to have `go.mod` and `go.sum` update as part of a PR. I'm not doing this deliberately so if that's not meant to happen I can reset them.

That's fine, its different to cargo where you tend to only push the toml file.

lkatalin:
I think this is almost ready for review, but I'm not sure how to reproduce the failing e2e test locally (and therefore figure out what's going on with it). It seems to pass:
```
[rekor@fedora rekor]$ sudo ./tests/e2e-test.sh 
starting services
rekor_redis-server_1 is up-to-date
rekor_mysql_1 is up-to-date
rekor_trillian-log-server_1 is up-to-date
rekor_trillian-log-signer_1 is up-to-date
rekor_rekor-server_1 is up-to-date
building CLI and server
waiting up to 60 sec for system to start
running tests
ok  	github.com/sigstore/rekor/tests	13.996s

```

dlorenc:
Could be a flake if it passes locally trying to re-run it now.

The other thing to try might be making sure you're fully rebased.

lkatalin:
> Could be a flake if it passes locally trying to re-run it now.
> 
> The other thing to try might be making sure you're fully rebased.

Hmm, I had rebased on main last night before running the tests and I just did so again. Puzzling.

cmd/rekor-cli/app/pflags.go
```diff
@@ -190,6 +191,53 @@ func validateFileOrURL(v string) error {
...
+	if len(v) == sharding.FullIDLen {
+
+		// validate separator
+		if string(v[sharding.ShardIDLen]) != sharding.FullIDSeparator {
+			return fmt.Errorf("unexpected separator char in fullID: %v", v[sharding.ShardIDLen])
+		}
+
+		// validate the ShardID
+		shardID := v[:sharding.ShardIDLen]
+		shardIDTag := "required," + "len=" + fmt.Sprintf("%v", sharding.ShardIDLen) + ",number"
```
jeffvance: nit: is there a reason this is not `shardIDTag := "required, len=" + fmt.Sprintf("%v",`?
Ditto for L215

lkatalin: @jeffvance No, there isn't. : ) Thanks for the catch, I will update this.

lkatalin:
It does look like it's an empty attestation on the failing test:
```
--- FAIL: TestIntoto (0.23s)
    e2e_test.go:439: g.Attesation:  []
    e2e_test.go:443: unexpected end of JSON input
```

dlorenc:
I've managed to reproduce this locally! Sorry it took a bit longer than expected, I'm on a new M1 Macbook and it took a bit to get the dev environment running there.

```shell
--- FAIL: TestIntoto (0.52s)
    e2e_test.go:439: g.Attesation:  []
    e2e_test.go:443: unexpected end of JSON input
FAIL
FAIL	github.com/sigstore/rekor/tests	0.710s
FAIL
```

I think I have a guess at the issue...

pkg/api/entries.go
```diff
@@ -183,7 +184,8 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	metricNewEntries.Inc()
 
 	queuedLeaf := resp.getAddResult.QueuedLeaf.Leaf
-	uuid := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+	shardID := sharding.NewCurrent()
+	uuid := shardID.ShardIDString + sharding.FullIDSeparator + hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
```
dlorenc, line 189: I think we need to make this a helper function, and call it anywhere we go from a MerkleLeafHash to a UUID.

To fix the e2e test, we also need it right here: https://github.com/sigstore/rekor/blob/main/pkg/api/entries.go#L95

dlorenc:
OK, got a passing test over here! https://github.com/sigstore/rekor/pull/548

I think I see the issue now. Let me know if you want to chat a bit more this week, I have a few ideas.

pkg/api/entries.go
lkatalin: @dlorenc I can change it to a helper function, that's a great idea.
Thank you for figuring out what's going on with the e2e test!

lkatalin:
> OK, got a passing test over here! #548
> 
> I think I see the issue now. Let me know if you want to chat a bit more this week, I have a few ideas.

Thanks @dlorenc , I really appreciate you looking into it. I still need to understand how the changes you suggested are related to the e2e test. I'm also happy to incorporate those changes in whatever way you think is best to get this to pass CI and merge it asap.

dlorenc:
> I still need to understand how the changes you suggested are related to the e2e test

At a high level it's just a case of confusion between `MerkleLeafHash` and `UUID` - the code was storing something using `MerkleTreeHash` (an In-Toto Attestation), and then trying to retrieve it later by `UUID`. 

The In-Toto Attestations are stored separately from the Transparency Log itself, to make them redactable if we ever need to. The failing test tests that behavior. The change in [entries.go](https://github.com/sigstore/rekor/pull/548/commits/f5b26b386013cd983ac77b6d33ee001df961e2f8#diff-d20fa2d682d00a31fa8535c7b3607012adab9f6fa9151768090a6dfab1a5b9b0)  fixes that up so we retrieve it by the UUID instead of the `MerkleTreeHash`, so we're consistent.

The change in [get.go](https://github.com/sigstore/rekor/pull/548/commits/f5b26b386013cd983ac77b6d33ee001df961e2f8#diff-0b8353c3890b62ea649c203d4b638c5f1aa823cdd63fa3146dc42cae1ebe1fd4) is unnecessary, but helped me when debugging and is otherwise safe. You can drop it if you prefer and we can handle that one separately.

cmd/rekor-cli/app/sharding/sharding.go
```diff
@@ -0,0 +1,58 @@
...
+func NewCurrent() ShardID {
+	return ShardID{
+		ShardIDInt:    CurrentShardID,
+		ShardIDString: IntToString(CurrentShardID)}
+}
+
+// TODO: The shardID will be part of the state and will be updated. Store it in state.go?
+// TODO: Add a check that the CurrentShardID cannot be updated beyond 999,999
+
+var CurrentShardID uint32 = 0
```
dlorenc, line 58: I think this will need to be passed or parsed out of the log index range flag, it would basically be the last one in the list: https://github.com/sigstore/rekor/blob/main/cmd/rekor-server/app/flags.go#L29

The client shouldn't need to know the active shard id - it's purely controlled server side by that flag.

cmd/rekor-cli/app/sharding/sharding.go
```diff
@@ -0,0 +1,58 @@
...
+
+// A ShardID is a number with specific properties, including
+// its representation as a six-digit string.
+type ShardID struct {
+	ShardIDInt    uint32
+	ShardIDString string
+}
+
+// Create a 6-digit string from shardID
+func IntToString(i uint32) string {
```
dlorenc, line 44: You can use strconv.ParseInt here, it'll also handle the error cases.

pkg/api/entries.go
```diff
@@ -183,7 +184,8 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	metricNewEntries.Inc()
 
 	queuedLeaf := resp.getAddResult.QueuedLeaf.Leaf
-	uuid := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
```
dlorenc, line 186: Actually, I'm not 100% sure this change is safe as is - I don't think it would be backwards compatible (we need a test for this). It means the server would start returning the shard-prefixed UUID to clients who create one - but older clients wouldn't have the change you added in here to `get.go`.

Older clients would then have a UUID that's invalid to them. I think we might need to drop this part for a release or two, to make sure clients can handle the new format.

cmd/rekor-cli/app/sharding/sharding.go
lkatalin: Thanks, I will use the `ActiveIndex()` from here https://github.com/sigstore/rekor/pull/549

pkg/api/entries.go
lkatalin: Interesting, I had not considered that older clients might still be used. I can write a test to make sure this will work when we include it.

lkatalin:
> > I still need to understand how the changes you suggested are related to the e2e test
> 
> At a high level it's just a case of confusion between `MerkleLeafHash` and `UUID` - the code was storing something using `MerkleTreeHash` (an In-Toto Attestation), and then trying to retrieve it later by `UUID`.
> 
> The In-Toto Attestations are stored separately from the Transparency Log itself, to make them redactable if we ever need to. The failing test tests that behavior. The change in [entries.go](https://github.com/sigstore/rekor/pull/548/commits/f5b26b386013cd983ac77b6d33ee001df961e2f8#diff-d20fa2d682d00a31fa8535c7b3607012adab9f6fa9151768090a6dfab1a5b9b0) fixes that up so we retrieve it by the UUID instead of the `MerkleTreeHash`, so we're consistent.
> 
> The change in [get.go](https://github.com/sigstore/rekor/pull/548/commits/f5b26b386013cd983ac77b6d33ee001df961e2f8#diff-0b8353c3890b62ea649c203d4b638c5f1aa823cdd63fa3146dc42cae1ebe1fd4) is unnecessary, but helped me when debugging and is otherwise safe. You can drop it if you prefer and we can handle that one separately.

@dlorenc Thank you for this explanation - I had totally missed that there is another part of the code using the `UUID`. I'm glad the CI caught this.

lkatalin:
I'm still making changes to this.

cmd/rekor-cli/app/get.go
```diff
@@ -106,22 +113,25 @@ var getCmd = &cobra.Command{
...
 			params := entries.NewGetLogEntryByUUIDParams()
 			params.SetTimeout(viper.GetDuration("timeout"))
-			params.EntryUUID = uuid
+
+			if len(uuid) == sharding.UUIDLen {
+				params.EntryUUID = uuid
+			} else if len(uuid) == sharding.FullIDLen {
+				params.EntryUUID = uuid[sharding.FullIDLen-sharding.UUIDLen:]
+			} else {
+				return nil, fmt.Errorf("invalid UUID length: %v", len(uuid))
```
bobcallaway, line 125: should we augment this error message to tell the user what the valid length/format should be?

cmd/rekor-cli/app/get.go
```diff
@@ -106,22 +113,25 @@ var getCmd = &cobra.Command{
...
 			if err != nil {
 				return nil, err
 			}
 
 			for k, entry := range resp.Payload {
-				if k != uuid {
-					continue
-				}
-
+				// TODO: understand why we don't need k to match params.EntryUUID
```
bobcallaway, line 134: `k` should match the UUID returned but as Dan mentioned in his other review comment there would be a concern about backward compatibility.

cmd/rekor-cli/app/get.go
bobcallaway, line 125: also, if you end up writing a helper method for creating this within the sharding package then would it make sense to move the validation logic in there, as well as add a new helper method to parse the input?

lkatalin:
Closed in favor of #583 #587 and #623 which together resolve #487

PR #535, opened 2021-12-14
sign using keyless for ci containers

cpanato:
#### Summary
- update key flag and add workflow-dispacth to trigger the build manually
- sing using keyless approach 

#### Ticket Link
n/a

#### Release Note

```release-note
update key flag and add workflow-dispacth to trigger the build manually
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,8 +119,8 @@ ko:
 		github.com/sigstore/rekor/cmd/rekor-cli
 
 sign-container: ko
-	cosign sign -key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-server:$(GIT_HASH)
-	cosign sign -key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-cli:$(GIT_HASH)
+	cosign sign --key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-server:$(GIT_HASH)
+	cosign sign --key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-cli:$(GIT_HASH)
 
 .PHONY: ko-local
 ko-local:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -119,8 +119,8 @@ ko:
 		github.com/sigstore/rekor/cmd/rekor-cli
 
 sign-container: ko
-	cosign sign -key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-server:$(GIT_HASH)
-	cosign sign -key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-cli:$(GIT_HASH)
+	cosign sign --force -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-server:$(GIT_HASH)
+	cosign sign --force -a GIT_HASH=$(GIT_HASH) ${KO_DOCKER_REPO}/rekor-cli:$(GIT_HASH)
```
lukehinds, line 128: We will need to revert these lines , as keyless won't work outside of a github action with `id-token: write` or having someone to manage a browser session. 

I guess the logic here is that ko builds rekor and then pushes to gcr.io. If we could do that outside of ko in a GH action then we can get rid of stashing private keys into the repo (which I swear one day is going to turn around a bite us).

cpanato: sounds good we can use the target that I've added here https://github.com/sigstore/rekor/pull/533/files#diff-1fd8c247611fff773b4bbd9d20b2e30557185c2893bb6613404996c3c2290816R35 to sign using keyless

dlorenc:
Is this one all set or do we need to delete the lines?

Makefile
cpanato: reverted

cpanato: added a target for keyless/ci let me know what you think about that

cpanato:
@lukehinds @dlorenc made some changes, let me know your thoughts

dlorenc (approved):

dlorenc (approved):

cpanato:
will watch the job

cpanato:
green now https://github.com/sigstore/rekor/runs/4549547999?check_suite_focus=true

PR #564, opened 2021-12-23
Refactor the shard map parsing so we can pass it down into the API ob…

dlorenc:
…ject.

Right now the type itself is defined in the cli package, which means we can't
use it without an import cycle.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


dlorenc:
This should also help with #515.

The next step is to build a map of trillian log clients, one per log id: https://github.com/sigstore/rekor/blob/main/pkg/api/api.go#L57

bobcallaway (approved):
lgtm

PR #565, opened 2021-12-23
Remove the attestation media type field.

dlorenc:
This was never actually correct - these are technically "payloadTypes", which are
not actually mediaTypes. Some implementations mistakenly sent incorrect media types, so
it appeared to work. The GCS storage layer rejected correct implementations that sent the
payloadType, because these are not valid mediaTypes.

We never used this field anyway, so let's drop it. I verified that the API correctly ignores
unknown fields, so removing this will not break clients that send it.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/rekor-cli/app/get.go
+++ b/cmd/rekor-cli/app/get.go
@@ -159,7 +159,6 @@ func parseEntry(uuid string, e models.LogEntryAnon) (interface{}, error) {
 
 	if e.Attestation != nil {
 		obj.Attestation = string(e.Attestation.Data)
-		obj.AttestationType = e.Attestation.MediaType
 	}
 
 	return &obj, nil
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -468,9 +468,7 @@ definitions:
           type: object
           properties:
             data:
-              format: byte
-            mediaType:
-              format: string
+              format: byte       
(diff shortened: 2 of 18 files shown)


dlorenc:
cc @mlieberman85 for catching this!

dlorenc:
We might want to add this back at some point - if we do then we probably want to call it "payloadType"

mlieberman85:
> We might want to add this back at some point - if we do then we probably want to call it "payloadType"

There also still seems to be some confusion over how to accurately and unambiguously communicate that the payload type is.

dlorenc:
> There also still seems to be some confusion over how to accurately and unambiguously communicate that the payload type is.

Yup - I think we got confused on which field it was supposed to represent.

There are at least 3 important ones.

- A DSSE envelope itself uses a DSSE specific media type when stored on disk or in a registry.
- The envelope is supposed to use this media type: https://github.com/in-toto/attestation/tree/main/spec#envelope inside to identify itself and to describe how to parse the body.
- The body then has another type (SLSA, cosign, etc.)

Since rekor always stores "attestations", they're always DSSE envelopes which are always of the in-toto payloadType.

lukehinds (approved):

PR #569, opened 2021-12-27
Implement /api/v1/version for rekor-server

Foxboron:
#### Summary
This implements `/api/v1/version` and allows us to properly check what version `rekor-server` is currently running on a deployed instance.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


dlorenc:
Nice! Let's see if GitHub is working again...

Foxboron:
The only discussion point I believe is if the different LDFLAGS values really belong under `pkg/api`, or if we want `pkg/version` to contain them?

dlorenc:
> The only discussion point I believe is if the different LDFLAGS values really belong under `pkg/api`, or if we want `pkg/version` to contain them?

I don't mind much either way. The new `debug/buildinfo` stuff in go 1.18 will help out here in a few months when we can upgrade to it.

Foxboron:
Fixed the missing license header. I can squash the commit with the generated swagger stuff if it's needed.

dlorenc (approved):

PR #583, opened 2022-01-05
Add sharding package and update validators

lkatalin:
#### Summary
Commit 1: Add sharding package. I moved this to `pkg/` so it can be accessed by both the server and cli as needed, but if this is no the right place, I can move it.
Commit 2: Update UUID validators to accept both current 64-char UUID and longer 80-char FullID defined in sharding package.

#### Ticket Link

Related to #487 
(More work to be done before the issue is closed)

### DIFF
--- a/cmd/rekor-cli/app/pflags.go
+++ b/cmd/rekor-cli/app/pflags.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/sigstore/rekor/pkg/pki"
+	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/util"
 
 	"github.com/spf13/pflag"
@@ -56,8 +57,10 @@ var pflagValueFuncMap map[FlagType]newPFlagValueFunc
 func initializePFlagMap() {
 	pflagValueFuncMap = map[FlagType]newPFlagValueFunc{
 		uuidFlag: func() pflag.Value {
-			// this corresponds to the merkle leaf hash of entries, which is represented by a 64 character hexadecimal string
-			return valueFactory(uuidFlag, validateString("required,len=64,hexadecimal"), "")
+			// this validates a UUID with or without a prepended TreeID;
+			// the UUID corresponds to the merkle leaf hash of entries,
+			// which is represented by a 64 character hexadecimal string
+			return valueFactory(uuidFlag, validateID, "")
 		},
 		shaFlag: func() pflag.Value {
 			// this validates a valid sha256 checksum which is optionally prefixed with 'sha256:'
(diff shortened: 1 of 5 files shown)


cpanato (commented):
lgtm

pkg/sharding/sharding.go
```diff
@@ -0,0 +1,78 @@
...
+		err := fmt.Errorf("invalid uuid len: %v", len(uuid))
+		return CreateEmptyFullID(), err
+	}
+
+	return FullID{
+		TreeID: treeid,
+		UUID:   uuid}, nil
+}
+
+func CreateEmptyFullID() FullID {
```
bobcallaway: is this intent for this to be public or just a internal helper function?

pkg/sharding/sharding.go
```diff
@@ -0,0 +1,78 @@
...
+var dummy = api.LogRanges{
+	Ranges: []api.LogRange{},
+}
+
+type FullID struct {
+	TreeID string
+	UUID   string
+}
+
+func CreateFullID(treeid string, uuid string) (FullID, error) {
```
bobcallaway, line 53: do you need any sort of validation in this method to ensure the arguments are valid hex strings?

cmd/rekor-cli/app/pflags.go
```diff
@@ -190,6 +191,22 @@ func validateFileOrURL(v string) error {
...
+// validateID ensures the ID is either a FullID (TreeID + UUID) or a UUID
+func validateID(v string) error {
+	if len(v) != sharding.FullIDHexStringLen && len(v) != sharding.UUIDHexStringLen {
+		return fmt.Errorf("ID len error, expected %v (FullID) or %v (UUID) but got len %v for ID %v", sharding.FullIDHexStringLen, sharding.UUIDHexStringLen, len(v), v)
+	}
+
+	IDStringValidatorFunc := validateString("required,hexadecimal")
+	IDErr := IDStringValidatorFunc(v)
+
+	if IDErr != nil {
```
bobcallaway:
```suggestion
	if err := validateString("required,hexadecimal")(v); err != nil {
```

lkatalin: Nice, thank you! I'm making this change.

pkg/sharding/sharding.go
lkatalin: Good catch, I'm still getting used to go's use of caps to mean public. I suppose for now it should be internal so I'll update it.

pkg/sharding/sharding.go
lkatalin: Thanks, added!

lkatalin:
Thanks for the helpful comments, @bobcallaway . I've made some updates. Also I'm not sure what we should call the `TreeID + UUID`. I've been using `FullID`, but maybe there is something more descriptive?

bobcallaway (approved):
lgtm, thanks!

bobcallaway:
> Thanks for the helpful comments, @bobcallaway . I've made some updates. Also I'm not sure what we should call the `TreeID + UUID`. I've been using `FullID`, but maybe there is something more descriptive?

`EntryID`?

dlorenc (approved):

dlorenc:
Nice job!

PR #587, opened 2022-01-07
Add logic to GET artifacts via old or new UUID

lkatalin:
#### Summary
This PR has 5 commits:
1. Moves `ranges.go` and its testing into the sharding package because it is related to sharding and I was getting import cycle errors otherwise.
2. Updates the name FullID to EntryID
3. Adds unit tests and a few more helper functions to sharding package
4. Adds logic to handle both new and old UUIDs during a GET request
5. Adds an e2e test to make sure a GET request with a longer EntryID is still handled properly

- [x] TODO: I think this would benefit from an integration test before being merged, but I'm still looking into how to add that. I assume it should have its own function is `tests/e2e_test.go`?

#### Ticket Link
Related to #487 
(needs another PR before that can be closed)

### DIFF
--- a/cmd/rekor-cli/app/get.go
+++ b/cmd/rekor-cli/app/get.go
@@ -34,6 +34,7 @@ import (
 	"github.com/sigstore/rekor/pkg/generated/client/entries"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/types"
 )
 
@@ -84,6 +85,11 @@ var getCmd = &cobra.Command{
 		}
 
 		logIndex := viper.GetString("log-index")
+		uuid := viper.GetString("uuid")
+		if logIndex == "" && uuid == "" {
+			return nil, errors.New("either --uuid or --log-index must be specified")
+		}
+
 		if logIndex != "" {
 			params := entries.NewGetLogEntryByIndexParams()
 			params.SetTimeout(viper.GetDuration("timeout"))
@@ -106,11 +112,14 @@ var getCmd = &cobra.Command{
 			}
 		}
 
-		uuid := viper.GetString("uuid")
 		if uuid != "" {
 			params := entries.NewGetLogEntryByUUIDParams()
 			params.SetTimeout(viper.GetDuration("timeout"))
-			params.EntryUUID = uuid
+
+			params.EntryUUID, err = sharding.GetUUIDFromIDString(uuid)
+			if err != nil {
+				return nil, fmt.Errorf("unable to parse uuid: %w", err)
+			}
 
 			resp, err := rekorClient.Entries.GetLogEntryByUUID(params)
 			if err != nil {
@@ -118,7 +127,7 @@ var getCmd = &cobra.Command{
 			}
 
 			for k, entry := range resp.Payload {
-				if k != uuid {
+				if k != params.EntryUUID {
 					continue
 				}
 
@@ -130,7 +139,7 @@ var getCmd = &cobra.Command{
 			}
 		}
 
-		return nil, errors.New("either --uuid or --log-index must be specified")
+		return nil, errors.New("entry not found")
 	}),
 }
 
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -39,6 +39,7 @@ import (
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/generated/restapi/operations/entries"
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/types"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
@@ -266,7 +267,12 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
 // GetLogEntryByUUIDHandler gets log entry and inclusion proof for specified UUID aka merkle leaf hash
 func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware.Responder {
 	ctx := params.HTTPRequest.Context()
-	hashValue, _ := hex.DecodeString(params.EntryUUID)
+
+	entryUUID, err := sharding.GetUUIDFromIDString(params.EntryUUID)
+	if err != nil {
+		return handleRekorAPIError(params, http.StatusBadRequest, err, "")
+	}
+	hashValue, _ := hex.DecodeString(entryUUID)
 	tc := NewTrillianClient(params.HTTPRequest.Context())
 
 	resp := tc.getLeafAndProofByHash(hashValue)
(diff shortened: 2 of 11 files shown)


pkg/sharding/sharding.go
```diff
@@ -87,3 +85,8 @@ func PrependActiveTreeID(uuid string) (FullID, error) {
 	active := dummy.ActiveIndex()
 	return CreateFullID(strconv.FormatUint(active, 10), uuid)
 }
+
+// Returns UUID (with no prepended TreeID) from a UUID or FullID string
+func GetUUIDFromIDString(id string) string {
+	return id[len(id)-UUIDHexStringLen:]
```
bobcallaway: what happens here if someone passes in a string that isn't a valid ID? you could end up panicking if the lengths are wrong. I think we need some sort of a defensive check here (with potentially an error return value as well)

lkatalin: You are right @bobcallaway , sorry I missed it.

lkatalin:
I'm still going to try to add an e2e test in here.

cmd/rekor-cli/app/get.go
```diff
@@ -110,15 +111,20 @@ var getCmd = &cobra.Command{
 		if uuid != "" {
 			params := entries.NewGetLogEntryByUUIDParams()
 			params.SetTimeout(viper.GetDuration("timeout"))
-			params.EntryUUID = uuid
+
+			entryUUID, err := sharding.GetUUIDFromIDString(uuid)
+			if err != nil {
+				return nil, fmt.Errorf("unable to parse uuid: %w", err)
+			}
+			params.EntryUUID = entryUUID
```
bobcallaway:
```suggestion
			params.EntryUUID, err = sharding.GetUUIDFromIDString(uuid)
			if err != nil {
				return nil, fmt.Errorf("unable to parse UUID: %w", err)
			}
```

cmd/rekor-cli/app/get.go
```diff
@@ -110,15 +111,20 @@ var getCmd = &cobra.Command{
...
 
 			resp, err := rekorClient.Entries.GetLogEntryByUUID(params)
 			if err != nil {
 				return nil, err
 			}
 
 			for k, entry := range resp.Payload {
-				if k != uuid {
+				if k != entryUUID {
```
bobcallaway:
```suggestion
				if k != params.entryUUID {
```

pkg/sharding/sharding.go
```diff
@@ -19,71 +19,81 @@ import (
...
-	return CreateFullID(strconv.FormatUint(active, 10), uuid)
+	return CreateEntryID(strconv.FormatUint(active, 10), uuid)
+}
+
+func (e EntryID) ReturnEntryIDString() string {
+	return e.TreeID + e.UUID
+}
+
+// Returns UUID (with no prepended TreeID) from a UUID or EntryID string
+func GetUUIDFromIDString(id string) (string, error) {
```
bobcallaway: Is there a way to move `validateID` logic from CLI into this package and just re-use it here? As written this doesn't check that `id` is a valid hex string. Its a tad paranoid given that the API layer will check the input values, but since its a exported function I think it would be good to be consistent.

pkg/sharding/sharding.go
```diff
@@ -19,71 +19,81 @@ import (
...
 }
 
-type FullID struct {
+type EntryID struct {
 	TreeID string
 	UUID   string
 }
 
-func CreateFullID(treeid string, uuid string) (FullID, error) {
+func CreateEntryID(treeid string, uuid string) (EntryID, error) {
```
bobcallaway: can you please add unit tests for this package?

pkg/sharding/sharding.go
lkatalin: Yeah, this is a good idea. I'm just not sure yet how to do it without tearing out the entire `pflags.go` logic and moving it into sharding. `validateID` relies on `validateString`, which uses the `validationFunc`, `baseValue`, etc. and I can't import those into `sharding` because `rekor-cli` has `sharding` as a dependency. Writing a duplicate version of `validateString` in `sharding` also seems suboptimal. Possibly would make sense to move `pflags.go` into a package accessible to both `sharding` and `rekor-cli`? What do you think?

pkg/sharding/sharding.go
lkatalin: Definitely!

pkg/sharding/sharding.go
lkatalin: I guess this function could also just move into `pflags`.

lkatalin: I think in order to do this, `pflags.go` and related files would have to move "up" and out of `rekor-cli` into a new `pkg/validation` or `cmd/validation` because that way `GetUUIDFromIDString` could be used in both `cmd/rekor-cli/app/get.go` and `pkg/api/entries.go` without import cycles. Still experimenting with this.

pkg/sharding/sharding.go
lkatalin: Done: the unit tests have their own commit.

pkg/sharding/sharding.go
lkatalin: I ended up using `hex.DecodeString` instead of moving the `pflags` logic around.

pkg/api/entries.go
```diff
@@ -266,7 +267,12 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
 // GetLogEntryByUUIDHandler gets log entry and inclusion proof for specified UUID aka merkle leaf hash
 func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware.Responder {
 	ctx := params.HTTPRequest.Context()
-	hashValue, _ := hex.DecodeString(params.EntryUUID)
+
+	entryUUID, err := sharding.GetUUIDFromIDString(params.EntryUUID)
+	if err != nil {
+		return handleRekorAPIError(params, http.StatusInternalServerError, err, "")
```
bobcallaway: If the UUID isn't valid, this should return an `http.StatusBadRequestError` since the problem is invalid input from the client.

pkg/sharding/sharding.go
```diff
@@ -19,71 +19,109 @@ import (
...
+
+// Returns UUID (with no prepended TreeID) from a UUID or EntryID string
+func GetUUIDFromIDString(id string) (string, error) {
+	if len(id) != UUIDHexStringLen && len(id) != EntryIDHexStringLen {
+		return "", fmt.Errorf("invalid ID len %v for %v", len(id), id)
+	}
+
+	if _, err := hex.DecodeString(id); err != nil {
+		err := fmt.Errorf("id %v is not a valid hex string: %v", id, err)
+		return "", err
```
bobcallaway:
```suggestion
		return "", fmt.Errorf("id %v is not a valid hex string: %v", id, err)
```

bobcallaway (commented):
2 minor nits. otherwise thanks for addressing all of the other comments and i think this should be good to merge once these two are fixed.

pkg/api/entries.go
lkatalin: Makes sense!

lkatalin:
> 2 minor nits. otherwise thanks for addressing all of the other comments and i think this should be good to merge once these two are fixed.

@bobcallaway Thanks for catching these!

bobcallaway (approved):
i'd like @dlorenc to take one final look but LGTM

dlorenc (approved):

PR #596, opened 2022-01-13
dsse type

### LINKED ISSUE #582: in-toto records don't contain signatures
Currently the in-toto type does not contain any signatures.  This prevents users of in-toto records from verifying attestations that are stored in rekor's attestation stores.

Additionally, the IntotoObj.content.hash refers to the hash of the entire DSSE envelope where the Attestation in the stores only contains the payload of the DSSE.  This prevents us from ensuring the attestation returned to us from Rekor's stores is what the IntotoObj on the merkle tree is describing unless we can reassemble the entire DSSE envelope byte-for-byte, which includes the signature.

@SantiagoTorres proposes storing the entire DSSE envelope as the Attestation

mikhailswift:
#### Summary

Adds a DSSE type that validated each signature on the envelope.  If the payload is an in-toto statement all in-toto subjects will be indexed.  The hash of the payload, PAE encoded payload, and public keys are also indexed.

I've tested this locally and it works but I need to write tests and add it to the e2e test script yet.  I'm opening this draft for discussion and input.

#### Ticket Link

Fixes #582

#### Release Note

```release-note

```


pkg/types/dsse/dsse.go
```diff
@@ -0,0 +1,73 @@
...
+var VersionMap = types.NewSemVerEntryFactoryMap()
+
+func (bdt BaseDsseType) UnmarshalEntry(pe models.ProposedEntry) (types.EntryImpl, error) {
+	if pe == nil {
+		return nil, errors.New("proposed entry cannot be nil")
+	}
+
+	in, ok := pe.(*models.Dsse)
+	if !ok {
+		return nil, errors.New("cannot unmarshal non-Rekord types")
```
bobcallaway:
```suggestion
		return nil, errors.New("cannot unmarshal non-DSSE types")
```

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,326 @@
...
+		})
+
+		allPubKeyBytes = append(allPubKeyBytes, sig.PublicKey)
+	}
+
+	_, err := verifyEnvelope(allPubKeyBytes, env)
+	if err != nil {
+		return fmt.Errorf("could not verify envelope: %w", err)
+	}
+
```
bobcallaway:
```suggestion
	if _, err := verifyEnvelope(allPubKeyBytes, env); err != nil {
		return fmt.Errorf("could not verify envelope: %w", err)
	}

```

bobcallaway (commented):
looking good so far, a couple minor things.

Does it make sense to move any of the `verifier` logic into [sigstore/sigstore](https://github.com/sigstore/sigstore/tree/main/pkg/signature/dsse) for reuse?

dlorenc:
Thanks for the PR and much needed cleanup!

mikhailswift:
Sorry for the delay in getting this finished up -- lost power over the weekend.  I'm going to try and get this cleaned up and a full PR up today.

pxp928:
@mikhailswift I just ran into this issue recently. Do you need any help pushing this over the finish line?

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
```diff
@@ -0,0 +1,61 @@
...
+    "type": "object",
+    "properties": {
+        "payload": {
+            "description": "payload of the envelope",
+            "type": "string",
+            "format": "byte",
+            "writeOnly": true
+        },
+        "payloadType": {
+            "description": "type descriping the payload",
```
dlorenc: Typo, describing

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
```diff
@@ -0,0 +1,61 @@
...
+            "type": "string"
+        },
+        "payloadHash": {
+            "description": "hash of the envelope's payload after being PAE encoded",
+            "readOnly": true,
+            "type": "object",
+            "properties": {
+                "algorithm": {
+                    "description": "The hasing function used to compue the hash value",
```
dlorenc: Typo, hashing

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,337 @@
...
+		return errors.New("nil verifier")
+	}
+	return v.v.VerifySignature(bytes.NewReader(sig), bytes.NewReader(data))
+}
+
+func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.ArtifactProperties) (models.ProposedEntry, error) {
+	returnVal := models.Dsse{}
+	re := V001Entry{}
+
+	var err error
```
dlorenc: Any reason to hoist this up here? It looks like it can be left inside the scope lower down.

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,337 @@
...
+	if err := json.Unmarshal(artifactBytes, &env); err != nil {
+		return nil, fmt.Errorf("payload must be a valid dsse envelope: %w", err)
+	}
+
+	allPubKeyBytes := make([][]byte, 0)
+	if props.PublicKeyBytes != nil {
+		allPubKeyBytes = append(allPubKeyBytes, props.PublicKeyBytes)
+	}
+
+	allPubKeyBytes = append(allPubKeyBytes, props.PublicKeysBytes...)
```
dlorenc, line 251: I might be misunderstanding this, but it looks like the props.PublicKeyBytes will get appended twice. Is that intentional?

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,337 @@
...
+		return nil, err
+	}
+
+	decodedPayload, err := base64.StdEncoding.DecodeString(env.Payload)
+	if err != nil {
+		return nil, fmt.Errorf("could not decode envelope payload: %w", err)
+	}
+
+	paeEncodedPayload := dsse.PAE(env.PayloadType, decodedPayload)
+	fmt.Println(decodedPayload)
```
dlorenc: Remove?

dlorenc (commented):
A few comments.

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
```diff
@@ -0,0 +1,61 @@
...
+                        "description": "public key that corresponds to this signature",
+                        "type": "string",
+                        "format": "byte",
+                        "readOnly": true
+                    }
+                }
+            }
+        }
+    },
+    "required": ["payloadType", "payloadHash", "signatures"]
```
bobcallaway: I think `payload` should be required here, but not `payloadHash`?

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func (v V001Entry) IndexKeys() ([]string, error) {
+	var result []string
+	payloadBytes := v.DsseObj.Payload
+	payloadType := *v.DsseObj.PayloadType
+	h := sha256.Sum256(payloadBytes)
+	payloadKey := "sha256:" + hex.EncodeToString(h[:])
+	result = append(result, payloadKey)
+	result = append(result, v.DsseObj.PayloadHash.Algorithm+":"+v.DsseObj.PayloadHash.Value)
```
bobcallaway: if `v.DsseObj.PayloadHash.Algorithm` is `sha256`, there's no need to recompute and add this twice, correct?

pkg/types/dsse/v0.0.1/entry.go
mikhailswift: PublicKeyBytes vs PublicKeysBytes

Part of the discussion on the issue was how to handle passing multiple public keys for cases where we would want to validate multiple signatures on the envelope.  To handle this a new prop called PublicKeysBytes was added to handle that case.  It's not a very nice solution though, so definitely open to alternatives here.

pkg/types/dsse/dsse.go
```diff
@@ -0,0 +1,75 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
bobcallaway:
```suggestion
// Copyright 2022 The Sigstore Authors.
```

bobcallaway: and same throughout

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	paeEncodedPayload := dsse.PAE(env.PayloadType, decodedPayload)
+	h := sha256.Sum256(paeEncodedPayload)
+	re.DsseObj.Payload = strfmt.Base64(env.Payload)
+	re.DsseObj.PayloadType = &env.PayloadType
+	re.DsseObj.PayloadHash = &models.DsseV001SchemaPayloadHash{
+		Algorithm: models.DsseV001SchemaPayloadHashAlgorithmSha256,
+		Value:     hex.EncodeToString(h[:]),
+	}
```
bobcallaway: this should be recomputed on the server side so no need to compute or set it here.

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,351 @@
...
+		allSigs[sig.Sig] = struct{}{}
+	}
+
+	for _, pubKeyBytes := range allPubKeyBytes {
+		key, err := x509.NewPublicKey(bytes.NewReader(pubKeyBytes))
+		if err != nil {
+			return nil, fmt.Errorf("could not parse public key as x509: %w", err)
+		}
+
+		vfr, err := signature.LoadVerifier(key.CryptoPubKey(), crypto.SHA256)
```
bobcallaway, line 318: is the hash algorithm always going to be SHA256 for this?

bobcallaway (commented):
A few comments/questions inline; I think we should also update `pkg/types/README.md` with some info about this type and how it explicitly differs from `intoto`.

pkg/types/dsse/v0.0.1/entry.go
mikhailswift: While not part of the DSSE spec it seems that all current users of DSSE default to using SHA256 during signing.

dlorenc:
This can get closed out now!

jmeth:
Sorry to comment on a closed PR but it seems like this was closed but never merged in to main. Was that the intention here?

pxp928:
> Sorry to comment on a closed PR but it seems like this was closed but never merged in to main. Was that the intention here?

@jmeth All the changes here were incorporated in another PR https://github.com/sigstore/rekor/pull/973

PR #623, opened 2022-01-26
Create EntryID for new artifacts and return EntryID to user

### LINKED ISSUE #487: Support UUID --> shard ID parsing

lkatalin:
#### Summary
This will complete the UUID parsing component of the sharding work. It creates and returns a longer UUID (EntryID) to the user, so this change should go into a new release.

Edit:
As far as I understand, we should wait to merge this PR until we're ready for a new Rekor release - are there other changes that would go into that release? Do we have a timeline?

#### Ticket Link

Fixes #487 

#### Release Note

```release-note
    TODO
```

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -187,7 +187,15 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	metricNewEntries.Inc()
 
 	queuedLeaf := resp.getAddResult.QueuedLeaf.Leaf
+
 	uuid := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+	activeTree := fmt.Sprintf("%x", tc.logID)
+	entryIDstruct, err := sharding.CreateEntryIDFromParts(activeTree, uuid)
+	if err != nil {
+		err := fmt.Errorf("error creating EntryID from active treeID %v and uuid %v: %w", activeTree, uuid, err)
+		return nil, handleRekorAPIError(params, http.StatusInternalServerError, err, fmt.Sprintf(validationError, err))
+	}
+	entryID := entryIDstruct.ReturnEntryIDString()
 
 	// The log index should be the virtual log index across all shards
 	virtualIndex := sharding.VirtualLogIndex(queuedLeaf.LeafIndex, api.logRanges.ActiveTreeID(), api.logRanges)
@@ -206,7 +214,7 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 				return
 			}
 			for _, key := range keys {
-				if err := addToIndex(context.Background(), key, uuid); err != nil {
+				if err := addToIndex(context.Background(), key, entryID); err != nil {
 					log.RequestIDLogger(params.HTTPRequest).Error(err)
 				}
 			}
@@ -218,11 +226,11 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 		go func() {
 			attestation := entry.Attestation()
 			if attestation == nil {
-				log.RequestIDLogger(params.HTTPRequest).Infof("no attestation for %s", uuid)
+				log.RequestIDLogger(params.HTTPRequest).Infof("no attestation for %s", entryID)
 				return
 			}
 			// TODO stop using uuid and use attestation hash
-			if err := storeAttestation(context.Background(), uuid, attestation); err != nil {
+			if err := storeAttestation(context.Background(), entryIDstruct.UUID, attestation); err != nil {
 				log.RequestIDLogger(params.HTTPRequest).Errorf("error storing attestation: %s", err)
 			}
 		}()
--- a/tests/e2e_test.go
+++ b/tests/e2e_test.go
@@ -223,7 +223,10 @@ func TestGet(t *testing.T) {
 	out := runCli(t, "upload", "--artifact", artifactPath, "--signature", sigPath, "--public-key", pubPath)
 	outputContains(t, out, "Created entry at")
 
-	uuid := getUUIDFromUploadOutput(t, out)
+	uuid, err := sharding.GetUUIDFromIDString(getUUIDFromUploadOutput(t, out))
+	if err != nil {
+		t.Error(err)
+	}
 
 	// since we at least have 1 valid entry, check the log at index 0
 	runCli(t, "get", "--log-index", "0")


pkg/api/entries.go
```diff
@@ -183,7 +183,14 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	metricNewEntries.Inc()
 
 	queuedLeaf := resp.getAddResult.QueuedLeaf.Leaf
-	uuid := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+
+	hexEncodedHash := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+	entryID, err := sharding.CreateEntryIDWithActiveTreeID(hexEncodedHash)
+	if err != nil {
+		err := fmt.Errorf("error creating EntryID from active treeID and uuid %v: %v", hexEncodedHash, err)
+		return nil, handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf(validationError, err))
```
```suggestion
		err := fmt.Errorf("error creating EntryID from active treeID and uuid %v: %w", hexEncodedHash, err)
		return nil, handleRekorAPIError(params, http.StatusInternalServerError, err, fmt.Sprintf(validationError, err))
```

bobcallaway: at this point we've already committed the change to the log... so while I expect this code path to never actually get executed, the burden wouldn't be on the client to resolve anything so we should return a 500 error.

dlorenc:
> Edit: As far as I understand, we should wait to merge this PR until we're ready for a new Rekor release - are there other changes that would go into that release? Do we have a timeline?

This is correct! We should release a version of the client that supports longer versions ASAP - then release this some time later.

@cpanato any idea if we can kick off another Rekor CLI release?

dlorenc:
Just to confirm my understanding:

* If you build rekor-cli from HEAD, it should work with rekor-server running this change AND before this change
* The currently released rekor-cli will not work with this change

cpanato:
> > Edit: As far as I understand, we should wait to merge this PR until we're ready for a new Rekor release - are there other changes that would go into that release? Do we have a timeline?
> 
> This is correct! We should release a version of the client that supports longer versions ASAP - then release this some time later.
> 
> @cpanato any idea if we can kick off another Rekor CLI release?

we can cut a v0.5.0 release, I don't think we are ready for a v1.0.0

I will check the changelog and maybe we can cut that today

dlorenc:
+1 to v0.5!

lkatalin:
> Just to confirm my understanding:
> 
>     * If you build rekor-cli from HEAD, it should work with rekor-server running this change AND before this change
> 
>     * The currently released rekor-cli will not work with this change

@dlorenc This is also my understanding. There is an e2e test right now that includes a rekor-cli GET command for the new EntryID [here](https://github.com/sigstore/rekor/blob/main/tests/e2e_test.go#L260).

cpanato:
> +1 to v0.5!

https://github.com/sigstore/rekor/releases/tag/v0.5.0

lukehinds (approved):
code looks good to me, once the cli is up to speed.

pkg/api/entries.go
```diff
@@ -183,7 +183,14 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	metricNewEntries.Inc()
 
 	queuedLeaf := resp.getAddResult.QueuedLeaf.Leaf
-	uuid := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+
+	hexEncodedHash := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+	entryID, err := sharding.CreateEntryIDWithActiveTreeID(hexEncodedHash)
+	if err != nil {
+		err := fmt.Errorf("error creating EntryID from active treeID and uuid %v: %v", hexEncodedHash, err)
```
bobcallaway: nit if you end up re-spinning this:
```suggestion
		err := fmt.Errorf("error creating EntryID from active treeID and uuid %v: %w", hexEncodedHash, err)
```

bobcallaway (approved):
lgtm other than minor error wrapping nit

lkatalin:
Note to self: after #671 is merged, we will have to use `CreateEntryIDFromParts(api.logRanges.ActiveIndex(), uuid)` instead of `CreateEntryIDWithActiveTreeID(uuid)`. This moves some logic out of the sharding package.

lkatalin:
I think this will require #671 to be merged first to pass CI.

bobcallaway (approved):

dlorenc:
What's the plan for merging and rolling this out?

lkatalin:
> What's the plan for merging and rolling this out?

@dlorenc Once we merge #727 , that should take care of the last known bug. @priyawadhwa and I are double checking on anything else that needs to happen from the [roadmap](https://github.com/sigstore/rekor/issues/711) before merging this, but I think most/all of the outstanding work should be optional.

tests/e2e_test.go
```diff
@@ -223,7 +223,10 @@ func TestGet(t *testing.T) {
 	out := runCli(t, "upload", "--artifact", artifactPath, "--signature", sigPath, "--public-key", pubPath)
 	outputContains(t, out, "Created entry at")
 
-	uuid := getUUIDFromUploadOutput(t, out)
+	uuid, err := sharding.GetUUIDFromIDString(getUUIDFromUploadOutput(t, out))
```
priyawadhwa, line 226: should `getUUIDFromUploadOutput` be renamed to `getEntryIDFromUploadOutput`?

pkg/api/entries.go
```diff
@@ -187,7 +187,14 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	metricNewEntries.Inc()
 
 	queuedLeaf := resp.getAddResult.QueuedLeaf.Leaf
-	uuid := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
+
+	hexEncodedHash := hex.EncodeToString(queuedLeaf.GetMerkleLeafHash())
```
priyawadhwa: `hexEncodedHash` is the same as `uuid` right? might be clearer to keep the name as `uuid` since `entry id = tree id + uuid`

priyawadhwa (approved):
small nits around naming, otherwise LGTM!

tests/e2e_test.go
lkatalin: Agree, though this is a larger problem all throughout the code. #625 is tracking this. Pulling on that thread will definitely create enough changes for another PR.

priyawadhwa: we can save it for later then!

priyawadhwa (approved):
lgtm

PR #626, opened 2022-01-27
 Add container simple signing type

### LINKED ISSUE #610: New Type based on Container signing specification
**Description**

This would be a rekor type specifically for container signatures (https://github.com/containers/image/blob/main/docs/containers-signature.5.md). This came up because cosign upload the hash of the entire signed payload, and rekor does not understand that payload.

For container signatures (not blobs), we may want to use a specific "containerType" so that rekor users can search for the sha256 of the image signed.

c/f https://sigstore.slack.com/archives/C01DGF0G8U9/p1642619114104300

asraa:

Adds a container type which understands simple signing payloads. This allows upload where the container signed can be indexed in the rekor entry. 
Adds search by reference for container types.

```
LogID: d54a19f9cf7ae9f46ca60224ec1dd0236c3ed0f70f2b6eb20469f1ef836fa735
Index: 23
IntegratedTime: 2022-02-04T01:00:30Z
UUID: d566e10b2ab155ebf92b8478d783c40495f509b296e258f1e8d0b992795ae797
Body: {
  "ContainerObj": {
    "data": {
      "content": {
        "critical": {
          "identity": {
            "docker-reference": "gcr.io/asra-ali/busybox/demo"
          },
          "image": {
            "docker-manifest-digest": "sha256:37e5287945774f27b418ce567cd77f4bbc9ef44a1bcd1a2312369f31f9cce567"
          },
          "type": "cosign container image signature"
        },
        "optional": null
      },
      "hash": {
        "algorithm": "sha256",
        "value": "05b434de7cebc2f8ca6245aaba3789d83a9c62f21f1e99b03b0362bb43f28dae"
      }
    },
    "signature": {
      "content": "MEYCIQDfqRSIaIWks7sGYsli7rmzckd1RbHeHIWbKRr4JW83fQIhAJRVWXA9nYgWl2hQK0i4ViC+hQ3ngAQxevuREDAcDCy2",
      "publicKey": {
        "content": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUNJVENDQWFlZ0F3SUJBZ0lVQUs1cm9HMEF4azhPYmI3MnRkSkV0cG90K0ZNd0NnWUlLb1pJemowRUF3TXcKS2pFVk1CTUdBMVVFQ2hNTWMybG5jM1J2Y21VdVpHVjJNUkV3RHdZRFZRUURFd2h6YVdkemRHOXlaVEFlRncweQpNakF5TURNeE5qRTRNREJhRncweU1qQXlNRE14TmpJM05UbGFNQk14RVRBUEJnTlZCQW9UQ0hOcFozTjBiM0psCk1Ga3dFd1lIS29aSXpqMENBUVlJS29aSXpqMERBUWNEUWdBRUdURkN4RUg0MEUzdnpJbWw5ZTA3ZDJZVktOZW8KMjh4Y1h5Q0Q3c3p4VzVFL211THA0N0U1d1IzU1NhR0pMdTJnb2thOVRiTnVqUDhQN3dIZXpKVzV4Nk9Cd1RDQgp2akFPQmdOVkhROEJBZjhFQkFNQ0I0QXdFd1lEVlIwbEJBd3dDZ1lJS3dZQkJRVUhBd013REFZRFZSMFRBUUgvCkJBSXdBREFkQmdOVkhRNEVGZ1FVaEppclc1eVVHaGlVenpXL3p5aGp5blJRelVzd0h3WURWUjBqQkJnd0ZvQVUKV01BZVg1RkZwV2FwZXN5UW9aTWkwQ3JGeGZvd0d3WURWUjBSQkJRd0VvRVFZWE55WVdGQVoyOXZaMnhsTG1OdgpiVEFzQmdvckJnRUVBWU8vTUFFQkJCNW9kSFJ3Y3pvdkwyZHBkR2gxWWk1amIyMHZiRzluYVc0dmIyRjFkR2d3CkNnWUlLb1pJemowRUF3TURhQUF3WlFJd0FZR3hwVnhRcW03SWhpVXRrR2o1NkFtTExhM0ZiTWJpdlZxQU4rWEMKblc2dDNzVzhTR3dKK2lCdWtndWhtNHFFQWpFQTFlTkxCQjZ5R0ZKdUY3S1ZzRlJ1cTZJemxselJCdjFXT01KSQpzOW5NdVNvYlMxazQrUXFFSVRKTzZjT09LM1J3Ci0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K"
      }
    }
  }
}

```

#### Ticket Link

Fixes https://github.com/sigstore/rekor/issues/610

#### Release Note

```release-note
* Adds a new container type that supports Simple Signing
```


pkg/types/container/container.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/pkg/errors"
+	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/sigstore/rekor/pkg/types"
+)
+
+const (
+	KIND = "container"
+)
+
+type BaseRekordType struct {
```
bobcallaway:
```suggestion
type BaseContainerType struct {
```

pkg/types/container/v0.0.1/container_v0_0_1_schema.json
```diff
@@ -0,0 +1,79 @@
...
+    "type": "object",
+    "properties": {
+        "signature": {
+            "description": "Information about the signature associated with the entry",
+            "type": "object",
+            "properties": {
+                "format": {
+                    "description": "Specifies the format of the signature",
+                    "type": "string",
+                    "enum": [ "pgp", "minisign", "x509", "ssh" ]
```
bobcallaway: do we need this level of flexibility for container signing?

pkg/types/container/v0.0.1/container_v0_0_1_schema.json
```diff
@@ -0,0 +1,79 @@
...
+                    "description": "Specifies the location of the payload",
+                    "type": "string",
+                    "format": "uri",
+                    "writeOnly": true
+                },
+                "content": {
+                    "description": "Specifies the JSON payload inline within the document",
+                    "type": "object",
+                    "additionalProperties": true,
+                    "writeOnly": true
```
bobcallaway:
```suggestion
```
since you're persisting things this will be both read & write

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,392 @@
...
+		b, err := json.Marshal(v.ContainerObj.Data.Content)
+		if err != nil {
+			log.Logger.Error(err)
+		} else {
+			img := &payload.Cosign{}
+			if err := json.Unmarshal(b, img); err == nil {
+				result = append(result, img.Image.DigestStr())
+			}
+		}
+		// TODO: Should we include the image reference as well?
```
bobcallaway: +1 I would include this, though we'd probably need to update the search interface for it

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,392 @@
...
+		img := &payload.Cosign{}
+		if err := json.Unmarshal(artifactBytes, img); err != nil {
+			return nil, err
+		}
+		re.ContainerObj.Data.Content = img
+	}
+
+	re.ContainerObj.Signature = &models.ContainerV001SchemaSignature{}
+	switch props.PKIFormat {
+	case "pgp":
```
bobcallaway: same question as above for desirable PKI formats

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+	dataReadCloser, err := util.FileOrURLReadCloser(ctx, v.ContainerObj.Data.URL.String(), contentBytes)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer dataReadCloser.Close()
+
+	hasher := sha256.New()
+	b := &bytes.Buffer{}
+	c := &bytes.Buffer{}
```
bobcallaway: where is this used?

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		log.Logger.Error(err)
+	} else {
+		keyHash := sha256.Sum256(key)
+		result = append(result, strings.ToLower(hex.EncodeToString(keyHash[:])))
+	}
+
+	result = append(result, keyObj.EmailAddresses()...)
+
+	// This is the hash of the signed payload.
```
bobcallaway, line 89: this is the hash over the marshalled JSON envelope, correct? do we have issues if the marshal/unmarshal of the signing envelope isn't consistently ordered?

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,392 @@
...
+func (v *V001Entry) fetchExternalEntities(ctx context.Context) (pki.PublicKey, pki.Signature, error) {
+	oldSHA := ""
+	if v.ContainerObj.Data.Hash != nil && v.ContainerObj.Data.Hash.Value != nil {
+		oldSHA = swag.StringValue(v.ContainerObj.Data.Hash.Value)
+	}
+
+	var contentBytes []byte
+	if v.ContainerObj.Data.Content != nil {
+		var err error
+		contentBytes, err = json.Marshal(v.ContainerObj.Data.Content)
```
bobcallaway: what happens if this marshalling ends up with the elements in a different order than when the SHA value (also in the same input doc) was computed?

bobcallaway (commented):
great stuff, some comments inline

bobcallaway (commented):
great stuff, some comments inline

dlorenc:
This looks great to me overall! Ping me when it's ready for a full review!

pkg/types/container/v0.0.1/container_v0_0_1_schema.json
asraa: I was trying to decide on that -- probably not, definitely not pgp, will restrict to x509

pkg/types/container/v0.0.1/entry.go
asraa: Hmmmm this is a good point. Ideally I think this should compare the SHA of the original `props.ArtifactBytes` that's originally passed in to avoid the unmarshal/marshal cycle.

pkg/types/container/v0.0.1/entry.go
asraa: If I change it like below to be the SHA of the payload bytes I think we're still sort of good -- hoping that it is the SHA of the raw bytes returned by something like https://github.com/sigstore/cosign/blob/a7bd67c0a7314200ed72e5d9870911e6149d76bf/pkg/oci/internal/signature/layer.go#L59

(which is currently what hashed rekord does -- I don't know if this totally "matches spec")

pkg/types/container/v0.0.1/entry.go
asraa: done

pkg/types/container/v0.0.1/entry.go
asraa: leftover debuggin!

asraa:
comments addressed!

* specifically decided not to add URL references for the container data
* i take the sha of the bytes passed in, which will be easy for cosign to do, and match the payload of `cosign generate`
* JSON remains in the entry stored (and the hash, to allow verification)
* Added search by reference (and both the sha of the signed payload and the containers sha)

cmd/rekor-cli/app/search.go
```diff
@@ -56,6 +56,8 @@ func addSearchPFlags(cmd *cobra.Command) error {
 
 	cmd.Flags().Var(NewFlagValue(shaFlag, ""), "sha", "the SHA256 or SHA1 sum of the artifact")
 
+	cmd.Flags().String("reference", "", "reference used to refer to or download the image ")
```
bobcallaway, line 59: could we use `dockerRef` or something a bit more specific here?

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+		publicKeyBytes, err = ioutil.ReadFile(filepath.Clean(props.PublicKeyPath.Path))
+		if err != nil {
+			return nil, fmt.Errorf("error reading public key file: %w", err)
+		}
+	}
+	re.ContainerObj.Signature.PublicKey.Content = strfmt.Base64(publicKeyBytes)
+
+	// validates and sets the data hash
+	if err := re.fetchExternalEntities(bytes.TrimSpace(artifactBytes)); err != nil {
```
bobcallaway, line 317: is there a reason we are editing the byte slice here? If i write use a local file as input here, this changes what was signed against which fails signature validation.

pkg/types/container/container.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func init() {
+	types.TypeMap.Store(KIND, New)
+}
+
+func New() types.TypeImpl {
+	brt := BaseContainerType{}
+	brt.Kind = KIND
+	brt.VersionMap = VersionMap
```
bobcallaway, line 42:
```suggestion
	bct := BaseContainerType{}
	bct.Kind = KIND
	bct.VersionMap = VersionMap
```

pkg/types/container/container.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil, errors.New("proposed entry cannot be nil")
+	}
+
+	container, ok := pe.(*models.Container)
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("%s, %s", "cannot unmarshal non-container types", pe.Kind()))
+	}
+
+	return rt.VersionedUnmarshal(container, *container.APIVersion)
+}
```
bobcallaway, line 59:
```suggestion
func (bct BaseContainerType) UnmarshalEntry(pe models.ProposedEntry) (types.EntryImpl, error) {
	if pe == nil {
		return nil, errors.New("proposed entry cannot be nil")
	}

	container, ok := pe.(*models.Container)
	if !ok {
		return nil, errors.New(fmt.Sprintf("%s, %s", "cannot unmarshal non-container types", pe.Kind()))
	}

	return bct.VersionedUnmarshal(container, *container.APIVersion)
}
```

pkg/types/container/container.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (rt *BaseContainerType) CreateProposedEntry(ctx context.Context, version string, props types.ArtifactProperties) (models.ProposedEntry, error) {
+	if version == "" {
+		version = rt.DefaultVersion()
+	}
+	ei, err := rt.VersionedUnmarshal(nil, version)
```
bobcallaway, line 65:
```suggestion
func (bct *BaseContainerType) CreateProposedEntry(ctx context.Context, version string, props types.ArtifactProperties) (models.ProposedEntry, error) {
	if version == "" {
		version = bct.DefaultVersion()
	}
	ei, err := bct.VersionedUnmarshal(nil, version)
```

pkg/types/container/container.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching container version implementation")
+	}
+
+	return ei.CreateFromArtifactProperties(ctx, props)
+}
+
+func (rt BaseContainerType) DefaultVersion() string {
```
bobcallaway, line 73:
```suggestion
func (bct BaseContainerType) DefaultVersion() string {
```

pkg/types/container/v0.0.1/container_v0_0_1_schema.json
```diff
@@ -0,0 +1,60 @@
...
+                    }
+                }
+            }
+        },
+        "data": {
+            "description": "Information about the JSON payload associated with the entry",
+            "type": "object",
+            "properties": {
+                "content": {
+                    "description": "Specifies the JSON payload inline within the document",
```
bobcallaway, line 35: just to confirm, you're leaving this generic to support OCI Descriptors or the simple signing format within the same rekor type?

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	hash := data.Hash
+	if hash == nil {
+		return nil, nil, errors.New("'hash' must be specified for data")
+	}
+	decoded, err := hex.DecodeString(*hash.Value)
+	if err != nil {
+		return nil, nil, err
+	}
```
bobcallaway, line 253: These aren't marked as required in the OpenAPI spec so either the spec should be fixed or they should be optional here

pkg/types/container/v0.0.1/entry.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+	}
+	re.ContainerObj.Signature.PublicKey.Content = strfmt.Base64(publicKeyBytes)
+
+	// validates and sets the data hash
+	if err := re.fetchExternalEntities(bytes.TrimSpace(artifactBytes)); err != nil {
+		return nil, fmt.Errorf("error validating entry: %w", err)
+	}
+
+	if _, _, err := re.validate(); err != nil {
```
bobcallaway, line 321: both the call to `fetchExternalEntities` and `validate` both verify the digital signature, we could probably get rid of one of them

pkg/types/container/v0.0.1/entry_test.go
```diff
@@ -0,0 +1,269 @@
...
+	keyBytes, _ := ioutil.ReadFile("../../../../tests/test_container_public_key.key")
+	dataBytes, _ := ioutil.ReadFile("../../../../tests/test_container.txt")
+
+	h := sha256.Sum256(bytes.TrimSpace(dataBytes))
+	dataSHA := hex.EncodeToString(h[:])
+
+	dataContent := &payload.Cosign{}
+	_ = json.Unmarshal(dataBytes, dataContent)
+
+	testCases := []TestCase{
```
bobcallaway, line 66: fyi: 

```
bcallaway@bcallaway01:~/devconf$ od -c junk2
0000000   n   u   l   l
0000004
bcallaway@bcallaway01:~/devconf$ /tmp/rekor/rekor-cli upload --artifact junk2 --signature junk2.sig --public-key ec_public.pem --type container --rekor_server=http://localhost:3000
Created entry at index 0, available at: http://localhost:3000/api/v1/log/entries/6e3840b58e13be114a8c83e1dace1731503ffaff357a079b7f70d28186933bc5
```

due to https://github.com/sigstore/sigstore/blob/10daa6daa06d88ce963939db227228764a336595/pkg/signature/payload/payload.go#L83

bobcallaway (commented):
I'm wondering if it is better to have the input JSON be a base64-encoded byte stream under `data`, and just verify the signature over that to avoid some of the marshal/unmarshal/marshal dancing in the code. I am in agreement on persisting the unmarshalled JSON in the actual rekor entry though - what do you think?

dlorenc:
Note that when we roll this out, we'll want to do it in a few phases:

* add support to the server
* make cosign and rekor-cli and any other clients understand the return type
* start using this type in writes

right now clients would typically explode if they receive an entry they don't understand, so it probably makes sense to cut a rekor-cli v0.6.0 after this goes in, but before cosign starts using it.

pkg/types/container/v0.0.1/entry.go
asraa: you're right -- this was to deal with cosign generate outputting with a trailing new line, but you should probably just feed it in without the trailing new line

pkg/types/container/v0.0.1/container_v0_0_1_schema.json
asraa: Yep -- clarified in the description that this is the payload associated with the signed entry.

pkg/types/container/v0.0.1/entry.go
asraa: On second thought I don't think we need it at all -- originally it was there for fetching the content by a URL ref, and then I thought to keep it in case you didn't want to upload the signed content. but at that point just use hashedrekord.

EDIT: jk, i want to keep the hash before I marshal it into the JSON for the entry.

asraa: Hmmm @bobcallaway I was just running the tests and finishing up addressing these comments but I have a major problem:

without a canonicalized hash of the content, there's no good way of verifying the entry. someone could send proposed entry with a a valid sig over a sha256 hash and then put whatever JSON payload they want. so entry verification really has to match the JSON payload with the sha256hash. 

i feel like there should be a way of canonicalization. but since signing w/ cosign is a signature over the payload bytes of the signed manifest, maybe I should stick the marshalled artifact bytes into the entry, rather than the unmarshalled interface

lukehinds:
still intending to work on this @asraa ?

bobcallaway:
closing due to inactivity

PR #657, opened 2022-02-07
use upstream k8s version lib

n3wscott:
#### Summary

Use the upsteam k8s version lib. See https://github.com/kubernetes-sigs/release-utils/pull/31 https://github.com/kubernetes-sigs/release-utils/pull/29 https://github.com/kubernetes-sigs/release-utils/pull/28

Note: there is a change around the pproc flag, if the flag is provided, the server will host the pproc endpoints. This use to be guarded by the lack of ldflags setting gitVersion. (though, would have had a bug if someone installed the server via `go install` or `go get`)

#### Ticket Link

See https://github.com/kubernetes-sigs/release-utils/issues/23

#### Release Note
```release-note
pproc flag use to be ignored for production builds, no longer ignored.
```


lukehinds:
What's the advantage of using the upstream version , is the current implementation lacking somehow?

edit: my main concern here is we are pulling in quite a few new deps, for what is more of a cosmetic change (iirc?)

n3wscott:
Deleting code and using a common lib to do the same thing. The main issue is this code is duplicated around the cloud native landscape about 300 times so we are working to centralize it. See also https://github.com/sigstore/cosign/pull/1413

edit: the new deps in .sum file are not added to the binary. There is only two new deps: the ascii text and the upstream k8s lib. The rest are just in the module at other paths.

bobcallaway (approved):

lukehinds (approved):

PR #671, opened 2022-02-15
Use logRangesFlag in API, route reads based on TreeID

### LINKED ISSUE #629: Prepare log-index lookup for sharding
The log-index lookup should resolve the virtual index to the treeID + index for a specific shard.

### LINKED ISSUE #632: Add routing logic to UUID retrieval for sharding
Currently UUID parsing prepends a TreeID (of zero) to a UUID to create an EntryID, and returns that to the user. UUID lookup can still be performed as usual using this EntryID, because the TreeID is actually discarded during the query. We should instead make sure the treeID is correct when creating the entry (perhaps in #630 ), and use the TreeID to query the correct tree during UUID-based artifact retrieval.

### LINKED ISSUE #670: Use logRangesMap flag in the actual API
The flag is passed in, but not used.

lkatalin:
#### Summary
This PR hooks up the `logRangeMap` flag to be used in API construction. The map can then be accessed to do routing for uuid-based and log-index-based reads.

I made the `logRangeMap` flag into a string because the `Set()` method can be used for construction and it was easier than trying to cram the data into a struct from the flag. But if there's a clean way to do this that doesn't involve this change, I'm happy to do it.

WIP because I still need to add some tests, ~~fix the e2e test~~, and do some cleanup, ex. maybe around the int types. And I need to think through 1) making sure this will not break old clients, 2) do we need to use this logic in the writes too before merging (I think this will actually be a change made to the pending PR that emits longer UUIDs upon artifact upload).

Questions:
- [x] How can the current TreeID be accessed during e2e testing?
  - Instead of this, I added a condition that if the `treeID` is zero, we use the `api.logID` instead 
- [x] In the case that both a `tlog_id` and `logRangeMap` are passed in, which takes precedence? The `logRangeMap`'s `ActiveIndex()` **should be** the `tlog_id` (currently using `ActiveIndex()`). 
  - Should we add a check to make sure these are the same? Edit: not doing this
  - Should we base the `tlog_id` in the API on the `ActiveIndex`? Edit: no, I'm appending the `tlog_id` to the `rangeMap`
  - Or should the `tlog_id` function as a flag for signaling the need for a new shard? This seems to be how it is used currently via `createAndInitTree`. Edit: yes, still doing this

Note: Currently it looks like the `tlog_id` in the API is used in only [one place](https://github.com/sigstore/rekor/blob/adab5c53055aa020e0d70da1ad25a394376a9aa6/pkg/api/api.go#L82) in `api.go`, to check whether it is zero and determine if the program should `createAndInitTree`.

Edit: So I think my answer to the above is to add logic that will append any nonzero `tlog_id` passed in at server startup to the API's `logRange`, see code from this PR [here](https://github.com/sigstore/rekor/blob/25b2ac8c0bb61a9bf059e0fad6b011900ef404c8/pkg/api/api.go#L92). If this isn't the right way, let me know or I guess we will change it later. 

#### Ticket Link
Fixes #670
Fixes #632 
Fixes #629

### DIFF
--- a/cmd/rekor-server/app/serve.go
+++ b/cmd/rekor-server/app/serve.go
@@ -103,6 +103,14 @@ var serveCmd = &cobra.Command{
 		server.Port = int(viper.GetUint("port"))
 		server.EnabledListeners = []string{"http"}
 
+		// Update logRangeMap if flag was passed in
+		rangeMap := viper.GetString("trillian_log_server.log_id_ranges")
+		if rangeMap != "" {
+			if err := logRangeMap.Set(rangeMap); err != nil {
+				log.Logger.Fatal("unable to set logRangeMap from flag: %v", err)
+			}
+		}
+
 		api.ConfigureAPI(logRangeMap.Ranges)
 		server.ConfigureAPI()
 
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strconv"
 
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
 	"github.com/go-openapi/runtime/middleware"
@@ -117,9 +118,11 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 // GetLogEntryAndProofByIndexHandler returns the entry and inclusion proof for a specified log index
 func GetLogEntryByIndexHandler(params entries.GetLogEntryByIndexParams) middleware.Responder {
 	ctx := params.HTTPRequest.Context()
-	tc := NewTrillianClient(ctx)
+	tid, resolvedIndex := api.logRanges.ResolveVirtualIndex(int(params.LogIndex))
+	tc := NewTrillianClientFromTreeID(ctx, tid)
+	log.RequestIDLogger(params.HTTPRequest).Debugf("Retrieving resolved index %v from TreeID %v", resolvedIndex, tid)
 
-	resp := tc.getLeafAndProofByIndex(params.LogIndex)
+	resp := tc.getLeafAndProofByIndex(resolvedIndex)
 	switch resp.status {
 	case codes.OK:
 	case codes.NotFound, codes.OutOfRange, codes.InvalidArgument:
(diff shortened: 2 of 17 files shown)


lkatalin:
TODO:
- [x] there's a bug when passing in a `--trillian_log_server.tlog_id` other than zero, because it doesn't get updated to the current one since it doesn't go through `createAndInitTree()` Update: FIXED

lukehinds (approved):
LGTM

cmd/rekor-server/app/serve.go
```diff
@@ -103,6 +103,14 @@ var serveCmd = &cobra.Command{
 		server.Port = int(viper.GetUint("port"))
 		server.EnabledListeners = []string{"http"}
 
+		// Update logRangeMap if flag was passed in
+		rangeMap := viper.GetString("trillian_log_server.log_id_ranges")
+		if rangeMap != "" {
+			if err := logRangeMap.Set(rangeMap); err != nil {
+				log.Logger.Errorf("unable to set logRangeMap from flag: %v", err)
```
bobcallaway: if we can't set the `logRangeMap` should this error be fatal?

pkg/api/entries.go
```diff
@@ -117,9 +117,11 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 // GetLogEntryAndProofByIndexHandler returns the entry and inclusion proof for a specified log index
 func GetLogEntryByIndexHandler(params entries.GetLogEntryByIndexParams) middleware.Responder {
 	ctx := params.HTTPRequest.Context()
-	tc := NewTrillianClient(ctx)
+	tid, resolvedIndex := api.logRanges.ResolveVirtualIndex(int(params.LogIndex))
+	tc := NewTrillianClientFromTreeID(ctx, int64(tid))
+	log.Logger.Debugf("Retrieving resolved index %v from TreeID %v", resolvedIndex, tid)
```
```suggestion
	log.RequestIDLogger(params.HTTPRequest).Debugf("Retrieving resolved index %v from TreeID %v", resolvedIndex, tid)
```

pkg/api/entries.go
```diff
@@ -272,8 +274,23 @@ func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware
...
+		}
+		// If tree ID is not known, fall back on using tlog_id from API
+	} /*else if tid == 0 {
+		tid = api.logID
+	}*/
 	hashValue, _ := hex.DecodeString(entryUUID)
-	tc := NewTrillianClient(params.HTTPRequest.Context())
+
+	tc := NewTrillianClientFromTreeID(params.HTTPRequest.Context(), tid)
+	log.Logger.Debugf("Retrieving UUID %v from TreeID %v", entryUUID, tid)
```
```suggestion
	tc := NewTrillianClientFromTreeID(ctx, tid)
	log.RequestIDLogger(params.HTTPRequest).Debugf("Retrieving UUID %v from TreeID %v", entryUUID, tid)
```

pkg/sharding/sharding.go
```diff
@@ -124,3 +124,27 @@ func GetUUIDFromIDString(id string) (string, error) {
 
 	return id[len(id)-UUIDHexStringLen:], nil
 }
+
+// Returns TreeID from an EntryID string
+// TODO: Add some tests
+func GetTreeIDFromEntryIDString(id string) (int64, error) {
+	if len(id) == UUIDHexStringLen {
```
bobcallaway: style: rewrite this as a switch statement on `len(id)`?

lkatalin:
I made the requested changes but also ended up adding some testing for `GetTreeIDFromIDString()` that led to a bunch of refactoring in the sharding package that now has its own commit.

Also ended up changing the type of the ints in `logRange` to `int64` because that's what trillian uses.

The handling of a `treeID` of `0` is currently a bit awkward as I'm torn between flagging it as invalid vs. allowing it to be a special value that signifies we should just use the API's current `tlog_id` and/or the API's `logRange`'s `ActiveIndex()`.

Edit: Unless there are objections, I've made a `treeID` of `0` invalid. I think this makes the most sense as that can't actually be routed and doesn't add information to a plain UUID.

lkatalin:
I added the `treeID` as output to `rekor-cli loginfo` and it does allow the `e2e` test to pass. However, I'm not sure how often the Rekor public server is updated with new code - there is an error when a `loginfo` with `treeID` request is sent to the currently running public Rekor server since it doesn't return this data.

Would it make more sense to return the `treeID` as a separate API endpoint after all so that people can still query the current server with `loginfo`? Or is this not an issue? @dlorenc @bobcallaway what do you think?

priyawadhwa (approved):

lkatalin:
@lukehinds @bobcallaway @dlorenc Is this okay to merge?

lukehinds:
Looks good to me. I will merge by 'lazy consensus' late Monday if nothing more from @bobcallaway / @dlorenc is posted before then.

dlorenc:
Lgtm!

PR #684, opened 2022-02-18
Set rekor-cli User-Agent header on requests

bobcallaway:
This sets the User-Agent header to match the format used in cosign of

`rekor-cli/vX.Y.Z (OS) (ARCH)`

#### Release Note
User-Agent HTTP request header set with specific version information rather than generic `Go/http-client` value


cpanato (approved):
cool

dlorenc (approved):

PR #700, opened 2022-03-02
Consistent parenthesis use in Makefile

k4leung4:
Signed-off-by: Kenny Leung <kleung@chainguard.dev>

#### Summary

Parenthesis and braces in Makefile are treated the same; https://web.mit.edu/gnu/doc/html/make_6.html#SEC56

We should be consistent and use one or the other, not both.
Since there seems to be more existing use of parenthesis, switching over to that.

#### Ticket Link

Fixes

#### Release Note

```release-note
None
```


dlorenc (approved):

cpanato (commented):
thanks
lgtm

PR #712, opened 2022-03-07
Change TreeID to be of type `string` instead of `int64`

priyawadhwa:
When printing the TreeID with rekor-cli loginfo, if the output is parsed through jq
then the TreeID gets rounded down as an int because it is bigger than JSON allows Numbers to be.

This is how jq works and is mentioned in the FAQ: https://github.com/stedolan/jq/wiki/FAQ#numbers

Switching this to a string will preserve the actual Tree ID.

How this used to work:
```
% go run ./cmd/rekor-cli loginfo --rekor_server http://localhost:3000 --format json
Persisted log state matches the current state of the log
{"TreeSize":0,"RootHash":"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","TimestampNanos":1646669010959148926,"TreeID":1593213237895900003}

# With jq, note TreeID is rounded down
% go run ./cmd/rekor-cli loginfo --rekor_server http://localhost:3000 --format json | jq
Persisted log state matches the current state of the log
{
  "TreeSize": 0,
  "RootHash": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
  "TimestampNanos": 1646669014825668000,
  "TreeID": 1593213237895900000
}
```

cc @lkatalin 

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/cmd/rekor-cli/app/log_info.go
+++ b/cmd/rekor-cli/app/log_info.go
@@ -43,7 +43,7 @@ type logInfoCmdOutput struct {
 	TreeSize       int64
 	RootHash       string
 	TimestampNanos uint64
-	TreeID         int64
+	TreeID         string
 }
 
 func (l *logInfoCmdOutput) String() string {
@@ -53,7 +53,7 @@ func (l *logInfoCmdOutput) String() string {
 Tree Size: %v
 Root Hash: %s
 Timestamp: %s
-TreeID:    %v
+TreeID:    %s
 `, l.TreeSize, l.RootHash, ts, l.TreeID)
 }
 
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -582,8 +582,9 @@ definitions:
         format: signedCheckpoint
         description: The current signed tree head
       treeID:
-        type: integer
+        type: string
         description: The current treeID
+        pattern: '^[0-9]+$'
     required:
       - rootHash
       - treeSize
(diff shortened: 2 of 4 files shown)


lkatalin (approved):
Great catch, @priyawadhwa , thank you!!

openapi.yaml
```diff
@@ -582,7 +582,7 @@ definitions:
         format: signedCheckpoint
         description: The current signed tree head
       treeID:
-        type: integer
+        type: string
```
bobcallaway, line 585: should we add a pattern here to help protect against bad input, e.g:

`pattern: '^[0-9]+'`

bobcallaway (commented):
one minor comment but LGTM

openapi.yaml
priyawadhwa: sgtm, updated to '^[0-9]+$'

dlorenc (approved):

cmd/rekor-cli/app/log_info.go
```diff
@@ -43,7 +43,7 @@ type logInfoCmdOutput struct {
 	TreeSize       int64
 	RootHash       string
 	TimestampNanos uint64
-	TreeID         int64
+	TreeID         string
```
mattmoor, line 46: Maybe use `json.Number`? https://pkg.go.dev/encoding/json#Number

priyawadhwa: Weirdly I'm still seeing the same issue with this even though it should be a string...

PR #718, opened 2022-03-08
Make the loginfo command a bit more future/backwards proof.

dlorenc:
The logid field comes back as nil (which is expected), but we break trying
to dereference that for formatting.

Looks like this was introduced in https://github.com/sigstore/rekor/pull/671

#### Ticket Link

Fixes

#### Release Note

```release-note

```


bobcallaway:
LGTM - just FYI there is a helper library in a dep we have that already implements these sorts of patterns:

https://github.com/go-openapi/swag

bobcallaway (approved):

dlorenc:
> LGTM - just FYI there is a helper library in a dep we have that already implements these sorts of patterns:

I couldn't find anything there that went this direction - it's really useful though for going from string to *string and stuff though.

bobcallaway:
https://github.com/go-openapi/swag/blob/f55deed90c3b563ee8dd3b7f68fed3e0104fc023/convert_types.go#L250

PR #734, opened 2022-03-15
Refactor rekor-cli loginfo

priyawadhwa:
This refactors the loginfo file in preparation for also getting info for inactive shards and verifying them as well.

It just breaks up the logic into a few smaller functions:
- `verifyTree` -- currently verifies only the active log, but it'll be easy to pass in treeID later to verify a specific log
- `loadVerifier` -- currently just gets the public key, but separating this out will make it easier to get the public key for a specific log later
- `proveConsistency` -- proves consistency between stored state and current state of the log

### DIFF
--- a/cmd/rekor-cli/app/log_info.go
+++ b/cmd/rekor-cli/app/log_info.go
@@ -28,6 +28,7 @@ import (
 	"github.com/go-openapi/swag"
 	"github.com/google/trillian/merkle/logverifier"
 	"github.com/google/trillian/merkle/rfc6962"
+	rclient "github.com/sigstore/rekor/pkg/generated/client"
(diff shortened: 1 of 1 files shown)


lkatalin:
Thanks, this is a good idea. If I'm reading this correctly, we will also be able to use this to verify each shard/tree in a `logRangeMap` passed in by the user as a next step?

priyawadhwa:
> If I'm reading this correctly, we will also be able to use this to verify each shard/tree in a logRangeMap passed in by the user as a next step?

yup, exactly!

lkatalin (approved):

dlorenc (approved):

PR #735, opened 2022-03-15
remove URL fetch of keys/artifacts server-side

bobcallaway:
We originally added the ability for Rekor API callers to specify URLs to fetch public keys, artifacts, & signatures to simplify fetching larger objects rather than forcing clients to download them locally and upload them to Rekor for verification. 

Now that we have the `hashedRekord` type, this patch removes the ability for the Rekor server to fetch arbitrary URLs before making entries into the log, largely due to the risk of [SSRF](https://owasp.org/www-community/attacks/Server_Side_Request_Forgery).

Note that the option for `rekor-cli` to specify URLs on the command line remains, as the CLI tool will download the artifacts locally and then insert the relevant contents in the appropriate position within the payload submitted to the Rekor server. 


dlorenc (approved):

dlorenc:
When and how do you want to roll this out?

bobcallaway:
It depends on how many releases we have before we go 1.0; given that it does deprecate function, it would be good to announce it in the next release (perhaps just the sharding stuff) and then we have a follow-on release that addresses the API (i.e. move to gRPC) and this fix. 

Thoughts?

dlorenc:
SGTM, so hold this until we cut the next release?

bobcallaway:
yup, sounds like a plan.

@cpanato we should include a note in the next rekor release noting that the "server side remote fetching of resources will be removed in the next release" (feel free to re-word)

bobcallaway:
https://github.com/sigstore/rekor/blob/12d1a47c7ac986932a2734cb855c642ac01ffde4/pkg/api/index.go also has another instance of this pattern we should remove.

Hayden-IO:
Also there's options to read sig and public key for PGP using `http.Client` instead of `util.FileOrURLReadCloser`

https://github.com/sigstore/rekor/blob/5025a24e2861d1876d67a09cf6bfad0ee76d64c8/pkg/pki/pgp/pgp.go#L84

https://github.com/sigstore/rekor/blob/5025a24e2861d1876d67a09cf6bfad0ee76d64c8/pkg/pki/pgp/pgp.go#L225

Hayden-IO:
Did we want to include something in 0.6 around the deprecation of this? 
@cpanato fyi

bobcallaway:
it was in the changelog

Hayden-IO:
Ah missed it, all good!

dlorenc:
Should be safe to merge after that goes out!

PR #738, opened 2022-03-17
Update loginfo API endpoint to return information about inactive shards

### LINKED ISSUE #713: Need to update state.json for new shards
Otherwise, the output for `loginfo` is:

```
[rekor@fedora rekor]$ rekor-cli loginfo --rekor_server http://127.0.0.1:3000
current size of tree reported from server 1 is less than previously persisted state 19
```

priyawadhwa:
I updated the `LogInfo` object in the openapi spec to also return information about inactive shards. 

This PR also updates `rekor-cli loginfo` to verify inactive shards if they exist. `loginfo` will now first attempt to store state in `state.json` by TreeID instead of URL (that's because the URL for different shards is the same so this will prevent shards overwriting each other). If no TreeID is present then it will fall back to using the URL, so this change is backwards compatible.

This PR also updates the sharding integration test to run loginfo on inactive shards to ensure this works as expected.

cc @lkatalin @bobcallaway 

fixes https://github.com/sigstore/rekor/issues/713

```release-note
Return loginfo data for inactive shards if they exist when calling `/api/v1/log`.
```


pkg/api/tlog.go
```diff
@@ -136,3 +152,47 @@ func GetLogProofHandler(params tlog.GetLogProofParams) middleware.Responder {
 
 	return tlog.NewGetLogProofOK().WithPayload(&consistencyProof)
 }
+
+func inactiveShardLogInfo(ctx context.Context, tid int64) (*models.InactiveShardLogInfo, error) {
+	tc := NewTrillianClientFromTreeID(ctx, tid)
+	resp := tc.getLatest(0)
+	if resp.status != codes.OK {
+		return nil, fmt.Errorf("resp code is not OK")
```
dlorenc: nit: maybe log the actual response code?

dlorenc (commented):
One tiny not, otherwise LGTM!

pkg/api/tlog.go
```diff
@@ -136,3 +152,47 @@ func GetLogProofHandler(params tlog.GetLogProofParams) middleware.Responder {
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	sth.SetTimestamp(uint64(time.Now().UnixNano()))
+
+	// sign the log root ourselves to get the log root signature
+	if _, err := sth.Sign(viper.GetString("rekor_server.hostname"), api.signer, options.WithContext(ctx)); err != nil {
+		return nil, err
+	}
```
lkatalin, line 185: How does this get the log root signature?

cmd/rekor-cli/app/log_info.go
```diff
@@ -115,19 +142,24 @@ func verifyTree(rekorClient *rclient.Rekor, signedTreeHead, serverURL string) er
...
 		return err
 	}
 
 	if viper.GetBool("store_tree_state") {
+		if treeID != "" {
+			if err := state.Dump(treeID, &sth); err != nil {
+				log.CliLogger.Infof("Unable to store previous state: %v", err)
+			}
+		}
 		if err := state.Dump(serverURL, &sth); err != nil {
```
lkatalin, line 155: If there is no `treeID` this will just dump from the URL, but if there is both a `treeID` and a URL, what happens? Does the URL only cover the active tree and only in the case of an older client? Or does some state get stored twice?

lkatalin (approved):
Awesome!
I had a couple of informational questions that would help me understand some of the code.

cmd/rekor-cli/app/log_info.go
priyawadhwa: Some state will get stored twice! But when reading from the file the client will always prefer the `treeID` over the URL.

lkatalin: Okay cool, yeah I see that in `oldState`. Just trying to understand the mechanism here - the URL covers all trees, right? What does dumping by `treeID` (if available) add?

pkg/api/tlog.go
priyawadhwa: So this calls the [Sign]([pkg/api/tlog.go](https://github.com/sigstore/rekor/pull/738/files/e8d9d0e7e6e310fc247e1dbaed34207588c4d55c#diff-3f7114b69ea9d0ea2ade2f57bd77a2920a3b8337495e35de0e6b26588309abe0)) function, which stores the signature in the struct:

https://github.com/sigstore/rekor/blob/ebc1d7ae9a87a88e7b823f0f6921e74899cda914/pkg/util/signed_note.go#L69

To access it you would call `sth.Signatures`. Sorry if that wasn't clear, let me know if that doesn't answer your question!

cmd/rekor-cli/app/log_info.go
priyawadhwa: So if we have two shards, they'll be accessed at the same URL (e.g. rekor.sigstore.dev). We can't store state for both of them, because the key by URL would be the same. The TreeID is unique for each shard, so we can store state for all shards at once if we use `treeID` as the key.

pkg/api/tlog.go
lkatalin: Ah okay, I think I get it, this enables us to get the signature later (not in this line of code) with `sth.Signatures` - the lack of clarity is just from me being less familiar here, I appreciate the explanation.

cmd/rekor-cli/app/log_info.go
lkatalin: I think this makes sense, I see it is a hash map so we're giving it differentiated keys. I will try to get some output to see it for myself. Thanks for the explanation.

dlorenc (approved):

PR #739, opened 2022-03-18
Require tlog_id when log_id_ranges is passed in

lkatalin:
#### Summary
Based on the conversation @priyawadhwa in #727 , this separates concerns more cleanly between the `log_id_ranges` flag and `tlog_id` flag. `tlog_id` now specifies the active tree (shard) and can be used by itself. `log_id_ranges` specifies inactive trees (shards) and must be used in conjunction with `tlog_id`.

### DIFF
--- a/cmd/rekor-server/app/serve.go
+++ b/cmd/rekor-server/app/serve.go
@@ -106,7 +106,7 @@ var serveCmd = &cobra.Command{
 
 		// Update logRangeMap if flag was passed in
 		shardingConfig := viper.GetString("trillian_log_server.sharding_config")
-		treeID := viper.GetString("trillian_log_server.tlog_id")
+		treeID := viper.GetUint("trillian_log_server.tlog_id")
 
 		ranges, err := sharding.NewLogRanges(shardingConfig, treeID)
 		if err != nil {
--- a/pkg/sharding/log_index.go
+++ b/pkg/sharding/log_index.go
@@ -16,18 +16,28 @@ package sharding
 
 // VirtualLogIndex returns the virtual log index for a given leaf index
 func VirtualLogIndex(leafIndex int64, tid int64, ranges LogRanges) int64 {
-	// if we have no ranges, we have just one log! return the leafIndex as is
-	if ranges.Empty() {
-		return leafIndex
+	// if we have no inactive ranges, we have just one log! return the leafIndex as is
+	// as long as it matches the active tree ID
+	if ranges.NoInactive() {
+		if ranges.GetActive() == tid {
+			return leafIndex
+		}
+		return -1
 	}
 
 	var virtualIndex int64
-	for _, r := range ranges.GetRanges() {
+	for _, r := range ranges.GetInactive() {
 		if r.TreeID == tid {
 			return virtualIndex + leafIndex
 		}
 		virtualIndex += r.TreeLength
 	}
-	// this should never happen
+
+	// If no TreeID in Inactive matches the tid, the virtual index should be the active tree
+	if ranges.GetActive() == tid {
+		return virtualIndex + leafIndex
+	}
+
+	// Otherwise, the tid is invalid
 	return -1
 }
--- a/pkg/sharding/ranges.go
+++ b/pkg/sharding/ranges.go
@@ -18,7 +18,6 @@ package sharding
 import (
 	"fmt"
 	"io/ioutil"
-	"strconv"
 	"strings"
 
 	"github.com/ghodss/yaml"
@@ -26,7 +25,8 @@ import (
 )
 
 type LogRanges struct {
-	ranges Ranges
+	inactive Ranges
+	active   int64
 }
 
 type Ranges []LogRange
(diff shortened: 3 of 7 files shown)


cmd/rekor-server/app/flags.go
```diff
@@ -27,41 +27,43 @@ type LogRangesFlag struct {
...
-	lastTreeID, err := strconv.ParseInt(lastRangeStr, 10, 64)
-	if err != nil {
-		return err
+			inputRanges = append(inputRanges, lr)
+		}
 	}
 
+	// The last entry is special and should not have a terminating range, because this is active.
+	// It is specified by the tlog_id.
 	inputRanges = append(inputRanges, sharding.LogRange{
```
priyawadhwa: Instead of appending the active ID, we can add a `ActiveTreeID int` field to the `LogRangesFlag` struct that'll make it easier to access it

cmd/rekor-server/app/serve.go
```diff
@@ -103,11 +103,15 @@ var serveCmd = &cobra.Command{
...
 		server.EnabledListeners = []string{"http"}
 
-		// Update logRangeMap if flag was passed in
+		// Build logRangeMap if related flags were passed in
 		rangeMap := viper.GetString("trillian_log_server.log_id_ranges")
+		tlogID := viper.GetInt64("trillian_log_server.tlog_id")
 		if rangeMap != "" {
-			if err := logRangeMap.Set(rangeMap); err != nil {
-				log.Logger.Fatal("unable to set logRangeMap from flag: %v", err)
+			if tlogID == 0 {
```
priyawadhwa: we don't need this check because we check for it later in `logRangeMap.Set` again

cmd/rekor-server/app/flags.go
lkatalin: @priyawadhwa That's a good idea. This also implies we may have to move methods like `func (l *LogRanges) ResolveVirtualIndex(index int)` and `func (l *LogRanges) TotalLength() int64` to be methods on `logRangesFlag` instead of `LogRanges` in order to work across all shards (active + inactive), does that seem reasonable?

lkatalin: I've actually got a better way of doing it, I'm turning the `LogRanges` struct into this:
```
type LogRanges struct {
        inactive []LogRange
        active   int64
}
```
That should be the simplest way, stay tuned!

priyawadhwa: awesome!

cmd/rekor-server/app/root.go
```diff
@@ -63,7 +63,7 @@ func init() {
 
 	rootCmd.PersistentFlags().String("trillian_log_server.address", "127.0.0.1", "Trillian log server address")
 	rootCmd.PersistentFlags().Uint16("trillian_log_server.port", 8090, "Trillian log server port")
-	rootCmd.PersistentFlags().Uint("trillian_log_server.tlog_id", 0, "Trillian tree id")
+	rootCmd.PersistentFlags().String("trillian_log_server.tlog_id", "0", "Trillian tree id")
```
priyawadhwa: sorry I wasn't clear when we chatted offline, I think this should actually remain a `Uint` for backwards compatibility.

pkg/sharding/log_index.go
```diff
@@ -16,18 +16,29 @@ package sharding
...
+	if ranges.NoInactive() {
+		if ranges.active == tid {
+			return leafIndex
+		}
+		return -1
 	}
 
 	var virtualIndex int64
-	for _, r := range ranges.GetRanges() {
+	// TODO: Are the tree IDs guaranteed to be ordered deterministically in the Ranges?
```
priyawadhwa: yup, we can assume the user correctly set up the config so that the shards exist in order

pkg/sharding/ranges.go
```diff
@@ -44,6 +45,9 @@ func NewLogRanges(path string, treeID string) (LogRanges, error) {
 	if err != nil {
 		return LogRanges{}, errors.Wrapf(err, "%s is not a valid int64", treeID)
 	}
+	if id == 0 {
+		return LogRanges{}, errors.New("non-zero tlog_id required when passing in shard config filepath")
```
priyawadhwa: We could add the fix to this error message to make it easier for users:

```
please set the active tree ID via the `--trillian_log_server.tlog_id` flag
```

pkg/sharding/ranges.go
```diff
@@ -53,59 +57,76 @@ func NewLogRanges(path string, treeID string) (LogRanges, error) {
...
+
+func (l *LogRanges) NoActive() bool {
+	return l.active == 0
+}
+
+func (l *LogRanges) NoInactive() bool {
+	return l.inactive == nil
 }
 
 func (l *LogRanges) Empty() bool {
```
priyawadhwa: i think this might be unused at this point and can be removed?

priyawadhwa (commented):
Looks great!

pkg/sharding/ranges.go
lkatalin: Sounds good.

pkg/sharding/log_index.go
lkatalin: Perfect. Thank you!!

cmd/rekor-server/app/root.go
lkatalin: Aaahhh, got it. No worries, easy to fix.

priyawadhwa (approved):
LGTM!

bobcallaway (approved):

PR #742, opened 2022-03-21
Replace `trillian_log_server.log_id_ranges` flag with a config file

priyawadhwa:
Each shard requires the following info:
- treeID
- treeLength
- publicKey for verification

It's not easy to pass all of this in via CLI flags, so I proposed replacing `trillian_log_server.log_id_ranges` with `trillian_log_server.sharding_config` which point to a config file in YAML for inactive shards.

The active shard will still be specified by the current flags.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


priyawadhwa:
cc  @lkatalin !

dlorenc:
You did it!!

dlorenc (approved):

PR #746, opened 2022-03-29
Specify public key for inactive shards in shard config

priyawadhwa:
This is pretty much the last item we need for sharding! Specify a public key for inactive shards in the sharding config in case we are rotating signers; this way users can request the public key for a specific shard from the API

cc @lkatalin and @asraa 

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -93,6 +93,12 @@ paths:
       operationId: getPublicKey
       tags:
         - pubkey
+      parameters:
+        - in: query
+          name: treeID
+          type: string
+          pattern: '^[0-9]+$'
+          description: The tree ID of the tree that you wish to prove consistency for
       produces:
         - application/x-pem-file
       responses:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -98,7 +98,7 @@ paths:
           name: treeID
           type: string
           pattern: '^[0-9]+$'
-          description: The tree ID of the tree that you wish to prove consistency for
+          description: The tree ID of the tree you wish to get a public key for
       produces:
         - application/x-pem-file
       responses:
--- a/pkg/sharding/ranges.go
+++ b/pkg/sharding/ranges.go
@@ -16,6 +16,7 @@
 package sharding
 
 import (
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"strings"
@@ -25,18 +26,21 @@ import (
 )
 
 type LogRanges struct {
-	inactive Ranges
-	active   int64
+	inactive        Ranges
+	activePublicKey string
+	active          int64
 }
 
 type Ranges []LogRange
 
 type LogRange struct {
-	TreeID     int64 `yaml:"treeID"`
-	TreeLength int64 `yaml:"treeLength"`
+	TreeID           int64  `yaml:"treeID"`
+	TreeLength       int64  `yaml:"treeLength"`
+	EncodedPublicKey string `yaml:"encodedPublicKey"`
+	decodedPublicKey string
 }
 
-func NewLogRanges(path string, treeID uint) (LogRanges, error) {
+func NewLogRanges(path string, treeID uint, activePublicKey string) (LogRanges, error) {
 	if path == "" {
 		return LogRanges{}, nil
 	}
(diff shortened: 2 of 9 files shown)


pkg/sharding/ranges_test.go
```diff
@@ -94,3 +97,62 @@ func TestLogRanges_ResolveVirtualIndex(t *testing.T) {
...
+		description    string
+		treeID         string
+		expectedPubKey string
+		shouldErr      bool
+	}{
+		{
+			description:    "empty tree ID",
+			expectedPubKey: "activekey",
+		}, {
+			description:    "tree id with encoded public key",
```
lkatalin:
```suggestion
			description:    "tree id with decoded public key",
```
Since we're skipping the encoding/decoding and that tree has decoded pubkey only it seems better to describe that way.

pkg/sharding/ranges.go
```diff
@@ -119,3 +131,30 @@ func (l *LogRanges) String() string {
 	ranges = append(ranges, fmt.Sprintf("active=%d", l.active))
 	return strings.Join(ranges, ",")
 }
+
+// PublicKey returns the associated public key for the given Tree ID
+// and returns the active public key by default
+func (l *LogRanges) PublicKey(activePublicKey, treeID string) (string, error) {
```
lkatalin, line 137: Is the `activePublicKey` going to be different each time this is called? How will this normally be specified?

openapi.yaml
```diff
@@ -93,6 +93,12 @@ paths:
       operationId: getPublicKey
       tags:
         - pubkey
+      parameters:
+        - in: query
+          name: treeID
+          type: string
+          pattern: '^[0-9]+$'
+          description: The tree ID of the tree you wish to get a public key for
```
lkatalin, line 101: Is this endpoint external for the user like `loginfo` or just internal?

lkatalin (commented):
This looks good! I had a question around the pubkey API endpoint because I don't see any documentation but maybe it's only internal?

lkatalin:
Also I love the e2e tests! :100:

openapi.yaml
priyawadhwa: it's external! you can query it like this:

```
curl "http://localhost:3000/api/v1/log/publicKey?treeID=383271703219633232"
```

pkg/sharding/ranges.go
priyawadhwa: it should be the same every time, we get it from the signer passed in via the `--rekor_server.signer` flag

lkatalin: Would it make sense to make it global then instead of an argument? Just thinking out loud / nitpicking, it is also fine as-is.

openapi.yaml
lkatalin: Is that documented anywhere? Maybe I missed it.

lkatalin (approved):

dlorenc (approved):

openapi.yaml
priyawadhwa: I had to look at the openAPI docs [here](https://swagger.io/docs/specification/describing-parameters/#query-parameters) to figure out the different types of params they accept. I went with query because that's what we do for a bunch of other api calls as well (`/api/v1/log/retrieve`, `/api/v1/log/proof` etc)

https://github.com/sigstore/rekor/blob/035b262939ce595a5d86d2cbcb9a886812479703/openapi.yaml#L113-L132

pkg/sharding/ranges.go
priyawadhwa: Yah that would be better! I tried to do that initially but it required a bigger refactor since right now the the log ranges are passed in to `NewAPI` and that's where we get the public key. We'd need to move the public key logic out for this to work & I thought about doing it in another PR so this one wouldn't get too big.

lkatalin: That's totally reasonable. Thanks.

openapi.yaml
lkatalin: Oh okay, thank you, I meant documented in Rekor for the user to be able to use some of these commands, but we can do that later if we want to. The link is helpful to me as I'm also trying to figure it out!

priyawadhwa: ah gotcha, yah docs would be useful for sure!

Issue #711, opened 2022-03-07
Sharding open questions and todos remaining


lkatalin:
@priyawadhwa and I wanted a place to record some open questions, bugs, and todo items remaining for the sharding work.

### Questions
- [x] Is it expected that `rekor-cli search` will return artifacts from all available shards in that Rekor instance? (We think yes)
- [x] How will clients interact with sharded rekor instances when verifying? (Each shard will have its own signing key(s) according to the initial design doc) (Issue: #721) What is the impact of this? (@asraa has been working on this from the cosign side, but currently it seems there are not todo items relating to this in Rekor)
- [x] How can we "remove" an entry from a log? We don't currently have logs pointing at other logs in a linked list format. (deferred until after GA)

### Todos
- [x] Get Public Key via the API per Tree ID for loginfo/verification in general (PR: #734 and PR: #733)
- [x] Implement way to pass in inactive shard map as a config file (PR: #742)

### Bugs
- [x] Need to return universal index over shards to user upon artifact upload (PR: #725 )
- [x] Need to correctly update `state.json` when new shard is created (Issue: #713 )
- [x] Return TreeID as string in loginfo so that it doesn't get rounded down in jq (PR: #712 )
- [x] Leaving `tlog_id` unspecified or as `0` seems to default to old (inactive) shard on server startup. (Issue: #726, PR: #727)

### Documentation To Dos
- [x] Neglecting to populate a shard config file and pass it in at server startup breaks index resolution when manually restarting the local server. This is because state is saved in the mariadb backend and a new shard is created at each startup, so re-starting multiple times creates multiple shards.  (#749 ) --> this would also be fixed by building up the `logRangeMap` internally. UPDATE: This usage is actually an antipattern, so we will cover this in documentation.
- [x] Add docs on sharding, including how to populate the sharding config file (PR https://github.com/sigstore/sigstore-website/pull/145)

### Expected bugs
- Using the `v0.5` client with a sever that returns `entryIDs` (aka server code after #623 is merged) will work correctly for upload and GET requests for the active tree, but will not work correctly with GET requests for archived trees. This is because the `v0.5` client does not yet have correct tree routing to pass to the server. (Fixed by #750 )

### Lower-priority todos (can be deferred to after Sigstore GA)
- [ ] Update terminology throughout Rekor to refer to `treeIDs` and `entryIDs` #625 #672 
- [x] Update `createAndInitTree()` debug output to reflect use of pre-existing tree when appropriate (PR: #724 )
[...]

PR #750, opened 2022-04-04
Search inactive trees for GET by UUID requests

lkatalin:
UPDATED:
This fix is useful for client v0.5 when there is more than one shard. When a GET by UUID request is made with a UUID (as opposed to the EntryID that has the prepended TreeID), this change will first check the active tree, then iterate through inactive trees to see if the UUID can be found in any of them.

Fixes the "expected bug" mentioned in #711 

~~Add temporary fix to check old prod treeID for entry if UUID is not
found in active tree. This can be removed once client v0.5 is phased
out.~~

~~Related to #711 (the "expected bug") - this fix is only needed if we plan to shard the production server before the client `v0.5` is phased out.~~

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -272,54 +272,43 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
 
 // GetLogEntryByUUIDHandler gets log entry and inclusion proof for specified UUID aka merkle leaf hash
 func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware.Responder {
-	ctx := params.HTTPRequest.Context()
-
 	uuid, err := sharding.GetUUIDFromIDString(params.EntryUUID)
 	if err != nil {
-		return handleRekorAPIError(params, http.StatusBadRequest, err, "")
+		return handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf("could not get UUID from ID string %v", params.EntryUUID))
 	}
-	var tid int64
 	tidString, err := sharding.GetTreeIDFromIDString(params.EntryUUID)
-	if err != nil {
-		// If EntryID is plain UUID, assume no sharding and use ActiveTreeID. The ActiveTreeID
-		// will == the tlog_id if a tlog_id is passed in at server startup.
-		if err.Error() == "cannot get treeID from plain UUID" {
-			tid = api.logRanges.ActiveTreeID()
-		} else {
-			return handleRekorAPIError(params, http.StatusBadRequest, err, "")
-		}
-	} else {
-		tid, err = strconv.ParseInt(tidString, 16, 64)
+
+	// If treeID is found in EntryID, route to correct tree
+	if err == nil {
+		tid, err := strconv.ParseInt(tidString, 16, 64)
 		if err != nil {
-			return handleRekorAPIError(params, http.StatusBadRequest, err, "")
+			return handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf("could not convert treeID %v to int", tidString))
 		}
-	}
-	hashValue, _ := hex.DecodeString(uuid)
-
-	tc := NewTrillianClientFromTreeID(params.HTTPRequest.Context(), tid)
-	log.RequestIDLogger(params.HTTPRequest).Debugf("Retrieving UUID %v from TreeID %v", uuid, tid)
-
-	resp := tc.getLeafAndProofByHash(hashValue)
-	switch resp.status {
-	case codes.OK:
-	case codes.NotFound:
-		return handleRekorAPIError(params, http.StatusNotFound, fmt.Errorf("grpc error: %w", resp.err), "")
-	default:
-		return handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("grpc error: %w", resp.err), trillianUnexpectedResult)
+		logEntry, err := RetrieveUUID(params, uuid, tid)
+		if err != nil {
+			if errors.Is(err, ErrNotFound) {
+				return handleRekorAPIError(params, http.StatusNotFound, err, "")
+			}
+			return handleRekorAPIError(params, http.StatusInternalServerError, err, "")
+		}
+		return entries.NewGetLogEntryByUUIDOK().WithPayload(logEntry)
 	}
 
-	result := resp.getLeafAndProofResult
-	leaf := result.Leaf
-	if leaf == nil {
-		return handleRekorAPIError(params, http.StatusNotFound, errors.New("grpc returned 0 leaves with success code"), "")
-	}
+	// If EntryID is plain UUID (ex. from client v0.5), check all trees
+	if errors.Is(err, sharding.ErrPlainUUID) {
+		trees := []sharding.LogRange{{TreeID: api.logRanges.ActiveTreeID()}}
+		trees = append(trees, api.logRanges.GetInactive()...)
 
-	logEntry, err := logEntryFromLeaf(ctx, api.signer, tc, leaf, result.SignedLogRoot, result.Proof, tid, api.logRanges)
-	if err != nil {
-		return handleRekorAPIError(params, http.StatusInternalServerError, err, "")
+		for _, t := range trees {
+			logEntry, err := RetrieveUUID(params, uuid, t.TreeID)
+			if err != nil {
+				continue
+			}
+			return entries.NewGetLogEntryByUUIDOK().WithPayload(logEntry)
+		}
+		return handleRekorAPIError(params, http.StatusNotFound, err, "UUID not found in any known trees")
 	}
-
-	return entries.NewGetLogEntryByUUIDOK().WithPayload(logEntry)
+	return handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf("could not get treeID from ID string %v", params.EntryUUID))
 }
 
 // SearchLogQueryHandler searches log by index, UUID, or proposed entry and returns array of entries found with inclusion proofs
@@ -437,4 +426,39 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 	}
 
 	return entries.NewSearchLogQueryOK().WithPayload(resultPayload)
+}
+
+var ErrNotFound = errors.New("grpc returned 0 leaves with success code")
+
+// Attempt to retrieve a UUID from a backend tree
+func RetrieveUUID(params entries.GetLogEntryByUUIDParams, uuid string, tid int64) (models.LogEntry, error) {
+	ctx := params.HTTPRequest.Context()
+	hashValue, err := hex.DecodeString(uuid)
+	if err != nil {
+		return models.LogEntry{}, err
+	}
+
+	tc := NewTrillianClientFromTreeID(params.HTTPRequest.Context(), tid)
+	log.RequestIDLogger(params.HTTPRequest).Debugf("Attempting to retrieve UUID %v from TreeID %v", uuid, tid)
+
+	resp := tc.getLeafAndProofByHash(hashValue)
+	switch resp.status {
+	case codes.OK:
+		result := resp.getLeafAndProofResult
+		leaf := result.Leaf
+		if leaf == nil {
+			return models.LogEntry{}, ErrNotFound
+		}
+
+		logEntry, err := logEntryFromLeaf(ctx, api.signer, tc, leaf, result.SignedLogRoot, result.Proof, tid, api.logRanges)
+		if err != nil {
+			return models.LogEntry{}, errors.New("could not create log entry from leaf")
+		}
+		return logEntry, nil
+
+	case codes.NotFound:
+		return models.LogEntry{}, ErrNotFound
+	default:
+		return models.LogEntry{}, errors.New("unexpected error")
+	}
 }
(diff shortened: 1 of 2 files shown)


pkg/api/entries.go
```diff
@@ -303,7 +303,24 @@ func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware
 	switch resp.status {
 	case codes.OK:
 	case codes.NotFound:
-		return handleRekorAPIError(params, http.StatusNotFound, fmt.Errorf("grpc error: %w", resp.err), "")
+		// TODO: remove later
+		// temporary workaround while users update to client v0.6, which has correct routing based on
+		// tree IDs. if entry is not found in currently active tree, tries again to retrieve entry
+		// from previously active tree.
+		formerTID, err := strconv.ParseInt("3904496407287907110", 10, 64)
```
priyawadhwa: Instead of hardcoding the tree ID, could we go through all inactive shards in `ranges` and see if the UUID exists in them?

pkg/api/entries.go
```diff
@@ -303,7 +303,24 @@ func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware
 	switch resp.status {
 	case codes.OK:
 	case codes.NotFound:
-		return handleRekorAPIError(params, http.StatusNotFound, fmt.Errorf("grpc error: %w", resp.err), "")
+		// TODO: remove later
```
priyawadhwa: Instead of doing this logic here, I think it would make more sense in this conditional where we don't have a tree ID:

https://github.com/sigstore/rekor/blob/main/pkg/api/entries.go#L283-L290

so basically, if we don't have a tree id:
1. Check the active tree ID
2. If that fails, start going through each inactive shard & check those.
3. If none of those work, return an error

pkg/api/entries.go
lkatalin: Great idea! Done in updated version.

lkatalin:
Some notes on the new version:
- I made a new function called `GetTrees()` in `api.go` that can be used in `entries.go` to avoid doing a bunch of imports in that file.
- I made a new function `GetLogServerAndClients()` for setting up the `logClient` and `logAdminClient` since this is used in both `GetTrees()` and `NewAPI()`
- Pulled out logic to check a tree for a UUID into its own function `AttemptRetrieveUUID()` since this is now called multiple times in `GetLogEntryByUUIDHandler()`
- Did some refactoring of `GetLogEntryByUUIDHandler()` because it had a lot of if/else statements. I think the switch makes it more readable but maybe it can be improved further.

I don't like returning the error from `AttemptRetrieveUUID()` as a `bool`, but could not quickly find a way to check a `middleware.Responder` to see if it is an error type - any ideas here?

pkg/api/entries.go
```diff
@@ -438,3 +439,35 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 
 	return entries.NewSearchLogQueryOK().WithPayload(resultPayload)
 }
+
+// Attempt to retrieve a UUID from a backend tree
+// TODO: don't return a bool that specifies whether it is an error, but instead check
+// the middleware.Responder to see if it is an error type (how?)
+func AttemptRetrieveUUID(params entries.GetLogEntryByUUIDParams, uuid string, tid int64) (middleware.Responder, bool) {
```
lkatalin: I would love to not return the `bool` here (it's just a marker for whether this is an error state) - is there some way to check a `middleware.Responder` to see if it's an error type? Or should we implement something? I wanted to throw this up quickly so didn't go down this path of investigation. Could also return a separate `error` but it would be great if even that is not necessary.

priyawadhwa: nit: I would just call this `RetrieveUUID`,  it's implied that we're attempting to do so

pkg/api/entries.go
```diff
@@ -272,54 +272,55 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
...
+		return resp
 
-	resp := tc.getLeafAndProofByHash(hashValue)
-	switch resp.status {
-	case codes.OK:
-	case codes.NotFound:
-		return handleRekorAPIError(params, http.StatusNotFound, fmt.Errorf("grpc error: %w", resp.err), "")
+	// If there is an error, see if it is because the EntryID is a UUID and if so,
+	// handle it
 	default:
```
priyawadhwa: The switch statement within the switch statement is a little hard to read. How about a structure like this:

```golang
tidString, err := sharding.GetTreeIDFromIDString(params.EntryUUID)
if err == nil {
   // Get the UUID from the specified TreeID
}
// This should get the active tree ID and the inactive IDs
trees, err := GetTrees()
if err != nil {
  return handleRekorAPIError(params, http.StatusBadRequest, err, "")
}
for _, t := range trees {
  logEntry, err := RetrieveUUID(params, uuid, t)
  if err != nil {
    continue
  }
  return entries.NewGetLogEntryByUUIDOK().WithPayload(logEntry)
}
// If we've reached this point, none of the shards have worked out, so return an error
return handleRekorAPIError(params, http.StatusBadRequest, fmt.Errorf("unable to find UUID"), "")
```

pkg/api/entries.go
```diff
@@ -272,54 +272,55 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
...
-		return handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("grpc error: %w", resp.err), trillianUnexpectedResult)
-	}
+		switch err.Error() {
+		// If EntryID is plain UUID (ex. from client v0.5), first try the active tree
+		case "cannot get treeID from plain UUID":
+			tid := api.logRanges.ActiveTreeID()
```
priyawadhwa: Won't `GetTrees` include the active Tree ID?

pkg/api/api.go
```diff
@@ -66,18 +66,41 @@ type API struct {
 	certChainPem string              // PEM encoded timestamping cert chain
 }
 
-func NewAPI(ranges sharding.LogRanges, treeID uint) (*API, error) {
+func GetTrees() (*trillian.ListTreesResponse, error) {
```
priyawadhwa: I think we should be getting this list of trees from the config and the `tlog_id` flag, rather than from the server. Right now, this will return all trees, but it doesn't make sense to check them if the user hasn't specified them in the config.

pkg/api/entries.go
lkatalin: That's fair :)

lkatalin:
Thanks for the good comments @priyawadhwa , will incorporate these and get a new one up.

pkg/api/entries.go
lkatalin: I changed it to be more like this, let me know what you think about the readability.

lkatalin:
@priyawadhwa It's ready!

pkg/api/entries.go
```diff
@@ -272,54 +272,37 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
...
-	log.RequestIDLogger(params.HTTPRequest).Debugf("Retrieving UUID %v from TreeID %v", uuid, tid)
-
-	resp := tc.getLeafAndProofByHash(hashValue)
-	switch resp.status {
-	case codes.OK:
-	case codes.NotFound:
-		return handleRekorAPIError(params, http.StatusNotFound, fmt.Errorf("grpc error: %w", resp.err), "")
-	default:
-		return handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("grpc error: %w", resp.err), trillianUnexpectedResult)
+		resp, _ := RetrieveUUID(params, uuid, tid)
```
priyawadhwa: We probably shouldn't hide this error, might be useful to return it for the user if it's non-nil

pkg/api/entries.go
```diff
@@ -272,54 +272,37 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
...
+		return resp
 	}
 
-	result := resp.getLeafAndProofResult
-	leaf := result.Leaf
-	if leaf == nil {
-		return handleRekorAPIError(params, http.StatusNotFound, errors.New("grpc returned 0 leaves with success code"), "")
-	}
+	// If EntryID is plain UUID (ex. from client v0.5), check all trees
+	if err.Error() == "cannot get treeID from plain UUID" {
```
priyawadhwa: Best practice would be not to hardcode this string in case someone changes the wording on the error later on. You could either:

1. Create an error type to check for ([example here](https://earthly.dev/blog/golang-errors/#defining-expected-errors) which uses `errors.Is`)
2. Change the `GetUUIDFromIDString` function to return `"", nil` in the case that there is no treeID. That way the error is nil, but we know if the treeID is empty that we have to go through all the shards

pkg/api/entries.go
```diff
@@ -438,3 +421,36 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 
 	return entries.NewSearchLogQueryOK().WithPayload(resultPayload)
 }
+
+// Attempt to retrieve a UUID from a backend tree
+func RetrieveUUID(params entries.GetLogEntryByUUIDParams, uuid string, tid int64) (middleware.Responder, error) {
+	ctx := params.HTTPRequest.Context()
+	hashValue, _ := hex.DecodeString(uuid)
```
priyawadhwa: We should check this error in case someone passes in a malformed UUID?

pkg/api/entries.go
```diff
@@ -438,3 +421,36 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 
 	return entries.NewSearchLogQueryOK().WithPayload(resultPayload)
 }
+
+// Attempt to retrieve a UUID from a backend tree
+func RetrieveUUID(params entries.GetLogEntryByUUIDParams, uuid string, tid int64) (middleware.Responder, error) {
```
priyawadhwa: I think it would still make more sense to return `(models.LogEntry,, error)` here and create the response in `GetLogEntryByIndexHandler`

priyawadhwa (commented):
Thanks for the speedy changes @lkatalin ! Just a couple more comments and we should be good to go :)

pkg/api/entries.go
lkatalin: Looks like creating a `models.LogEntry` will require turning a [`trillian.GetEntryAndProofResponse`](https://pkg.go.dev/github.com/google/trillian#GetEntryAndProofResponse) into a `models.LogEntryAnon`, is that right? Is there a straightforward way to do that? I'm looking at [`logEntryFromLeaf`](https://github.com/sigstore/rekor/blob/6cef9e450975125b9e372a4d7488c63bf4bfd8ad/pkg/api/entries.go#L66) as an example but not really seeing how the fields match up.

pkg/api/entries.go
lkatalin: Yes, these options make more sense.

pkg/api/entries.go
lkatalin: I just moved this line of code from where it was earlier, but checking it seems like a good thing to do.

pkg/api/entries.go
lkatalin: Make a `LogEntryAnon` and unmarshal into it?

priyawadhwa: I think `logEntryFromLeaf` returns `models.LogEntry`, so after line [443](https://github.com/sigstore/rekor/pull/750/files#diff-d20fa2d682d00a31fa8535c7b3607012adab9f6fa9151768090a6dfab1a5b9b0R443) you could just `return logEntry, nil`?

pkg/api/entries.go
priyawadhwa: thanks ! :)

pkg/api/entries.go
lkatalin: Nice! You're so right. Thanks!

pkg/api/entries.go
lkatalin: This makes sense now when returning `(models.LogEntry, error)` from `RetrieveUUID`

pkg/api/entries.go
```diff
@@ -438,3 +427,39 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 
 	return entries.NewSearchLogQueryOK().WithPayload(resultPayload)
 }
+
+var ErrNotFound = errors.New("grpc returned 0 leaves with success code")
+var ErrUnexpected = errors.New("unexpected error")
```
priyawadhwa: I don't think we need this `ErrUnexpected` since we don't check for it anywhere

pkg/api/entries.go
```diff
@@ -438,3 +427,39 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
...
+	case codes.OK:
+		result := resp.getLeafAndProofResult
+		leaf := result.Leaf
+		if leaf == nil {
+			return models.LogEntry{}, ErrNotFound
+		}
+
+		logEntry, err := logEntryFromLeaf(ctx, api.signer, tc, leaf, result.SignedLogRoot, result.Proof, tid, api.logRanges)
+		if err != nil {
+			return models.LogEntry{}, ErrUnexpected
```
priyawadhwa: as mentioned above we can just return `err` here

priyawadhwa (approved):
Couple small comments!

pkg/api/entries.go
lkatalin: Okay cool, I took it out!

priyawadhwa (approved):
LGTM! 🎉

lkatalin:
Screenshot from 2022-04-08 13-01-49

dlorenc (approved):

Issue #749, opened 2022-04-01
Enforce shard config file requirement in multi-shard environment


lkatalin:
Otherwise, index resolution is broken (the server thinks all indexes are absolute on the current active tree).

~~Passing in no `tlog_id` or shard config file (ex. for a local instance) should still be fine - although as of server 0.6, server startup with no `tlog_id` flag will automatically create a new tree, which breaks lookup in the inactive tree with client 0.5. But I don't think use of client 0.5 with server 0.6 locally should be supported anyway.~~

lkatalin:
Update: shard config file should be required in multi-shard environment. This should be enforced, rather than just letting GET by --log-index break in this scenario.

lkatalin:
Neglecting to populate a shard config file and pass it in at server startup breaks index resolution when manually restarting the local server. This is because state is saved in the database and a new shard is created at each startup, so re-starting multiple times creates multiple shards.

UPDATE: This usage is actually an antipattern, so we will cover this in documentation.

lkatalin:
Fixing this here: https://github.com/sigstore/sigstore-website/pull/146

lkatalin:
Closing since https://github.com/sigstore/sigstore-website/pull/146/ was merged.

PR #755, opened 2022-04-06
Separate create_tree functionality from tlog_id

lkatalin:
Previously, starting a local Rekor server with no tlog_id flag
would automatically create a new tree (shard). The GET by
--log-index breaks if called on a multi-shard server with no
shard config passed in (this problem remains).

This change allows users who are not deliberately using sharding
to (re)start a local Rekor server without automatically sharding,
which in turn prevents GET by --log-index from breaking for these
users. Users who are sharding or working in a multi-shard
environment MUST pass in a shard config file at server startup in
order for GET by --log-index to work.

Signed-off-by: Lily Sturmann <lsturman@redhat.com>

Related: #749 

cc @priyawadhwa


lkatalin:
Closing as discussed with @priyawadhwa . Instead of introducing a new flag to create trees, which will break backwards compatibility, we will document the expected usage of the current flags and what to expect if starting up a local Rekor server multiple times.

PR #759, opened 2022-04-07
set p.Block after parsing in helm provenance type

bobcallaway:
not sure that this is an actual issue, but we should only set `p.Block` after ensuring `parseMessageBlock` returns without error.


asraa (approved):
Nice!

dlorenc (approved):

dlorenc:
Looks great, I'll hold for Andrew to give it a check too.

sabre1041:
@bobcallaway @dlorenc Queued this up to review this evening

sabre1041 (approved):
LGTM

PR #764, opened 2022-04-08
Add the SHA256 digest of the intoto payload into the rekor entry

bobcallaway:
Currently we are storing the digest over the entire intoto envelope in the corresponding Rekor entry. This PR adds the SHA256 digest for the envelope's payload (the base64 decoded bytes) into the log entry as well. 

This also fixes the length check in `Attestation` to measure against the length of bytes after base64 decoding the payload, instead of the payload itself.

@asraa @haydentherapper @SantiagoTorres

### DIFF
--- a/pkg/types/intoto/v0.0.1/entry.go
+++ b/pkg/types/intoto/v0.0.1/entry.go
@@ -150,6 +150,14 @@ func (v *V001Entry) Canonicalize(ctx context.Context) ([]byte, error) {
 			},
 		},
 	}
+	attestation := v.Attestation()
+	if attestation != nil {
+		attH := sha256.Sum256(attestation)
+		canonicalEntry.Content.AttestationHash = &models.IntotoV001SchemaContentAttestationHash{
+			Algorithm: swag.String(models.IntotoV001SchemaContentHashAlgorithmSha256),
+			Value:     swag.String(hex.EncodeToString(attH[:])),
+		}
+	}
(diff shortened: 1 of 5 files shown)


pkg/types/intoto/v0.0.1/entry.go
```diff
@@ -150,6 +151,18 @@ func (v *V001Entry) Canonicalize(ctx context.Context) ([]byte, error) {
 			},
 		},
 	}
+	attestation := v.Attestation()
+	if attestation != nil {
+		decodedAttestation, err := base64.StdEncoding.DecodeString(string(attestation))
+		if err != nil {
+			return nil, errors.Wrap(err, "decoding attestation")
+		}
+		attH := sha256.Sum256(decodedAttestation)
```
Hayden-IO, line 160: Should you either return an error if a non sha256 algorithm enum is specified?

Hayden-IO (approved):

pkg/types/intoto/v0.0.1/entry.go
bobcallaway: we're setting the value, its not user selectable

Hayden-IO: Ah, looks good then!

dlorenc (approved):

dlorenc:
Nice!

PR #767, opened 2022-04-09
Fix search without sha prefix

### LINKED ISSUE #699: Rekor API: searching for an artifact using SHA256 fails if not prepended with `sha256:`
The Rekor API returned no results for an artifact I'd just uploaded when I searched for it using its SHA. This is despite the record existing  and having the correct SHA (verified by searching using its pubkey).

@bobcallaway helped diagnose this; seems like the SHA needs to be prepended with `sha256:` to work, even though the [API spec](https://github.com/sigstore/rekor/blob/main/openapi.yaml#L546) does not explicitly require that.

eddiezane:
#### Summary

Rekor API searches without a SHA prefix (sha1:, sha256:) do not  currently work as the hash is stored in Redis with the prefix. rekor-cli infers the prefix when using --sha.

This PR reuses that logic when searching with `/index/retrieve`.

#### Ticket Link

Fixes #699 

#### Release Note

```release-note
Fixed searching the index by SHA without a prefix.
```

### DIFF
--- a/cmd/rekor-cli/app/search.go
+++ b/cmd/rekor-cli/app/search.go
@@ -37,6 +37,7 @@ import (
 	"github.com/sigstore/rekor/pkg/generated/client/index"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/util"
 )
 
 type searchCmdOutput struct {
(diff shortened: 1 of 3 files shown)


pkg/api/index.go
```diff
@@ -38,8 +38,9 @@ func SearchIndexHandler(params index.SearchIndexParams) middleware.Responder {
 	var result []string
 	if params.Query.Hash != "" {
 		// This must be a valid sha256 hash
```
eddiezane, line 40: SHA1 still seems to be supported?

cmd/rekor-cli/app/search.go
```diff
@@ -108,15 +109,7 @@ var searchCmd = &cobra.Command{
...
-			var prefix string
-			if !strings.HasPrefix(sha, "sha256:") && !strings.HasPrefix(sha, "sha1:") {
-				if len(sha) == 40 {
-					prefix = "sha1:"
-				} else {
-					prefix = "sha256:"
-				}
-			}
-			params.Query.Hash = fmt.Sprintf("%v%v", prefix, sha)
+			params.Query.Hash = util.PrefixSHA(sha)
```
eddiezane, line 112: Is this still needed here? Probably if folks update the cli but not the server?

pkg/api/index.go
dlorenc: Good catch, I think that actually came later because we wanted to support git. So the comment is just out of date.

dlorenc (approved):

cmd/rekor-cli/app/search.go
dlorenc: Right, safe to keep!

cpanato (approved):

Issue #484, opened 2021-11-09
gRPC API Endpoint


naveensrinivasan:
**Description**
Rekor could expose a gRPC endpoint along with REST which will help with performance #481 to publish large binaries

naveensrinivasan:
https://github.com/sigstore/rekor/issues/481#issuecomment-970236333

PR #771, opened 2022-04-12
initial gRPC API & message definition

bobcallaway:
***Note*: this is targeted at a feature branch, not main. I intend to continue to refine the gRPC work in a branch until we get it ready to merge over into main since it will touch most of Rekor in some way or another.**

This starts work on #484; when complete, it also aims to simplify the code generation for `rekor-server`, custom type definition and easier multi-language client code generation. This PR just shares a draft of the external service & message definitions for feedback, and starts to plumb in the golang client & server side code generation.

TODO (likely incomplete but wanted to start a list):

- [ ] figure out how to support current `v1` API support through legacy gRPC interface (similar to Fulcio)
- [ ] figure out version implications of gRPC with regards to parsing older log entries
- [ ] re-envision `types` interfaces (opportunity for leveraging 1.18 generic support?)
- [ ] implement an e2e flow for a single custom
- [ ] implement e2e flows for all custom types
- [ ] update custom type documentation for devs
- [ ] add OpenAPI documentation (to be generated from proto files) instead of a hardcoded `openapi.yaml` file?
- [ ] remove swagger-based implementation


dlorenc:
If you're targeting a feature branch, feel free to just merge away and we can do a larger review after. Feel free to leave them open though if you do want reviews before the merge.

bobcallaway:
I will plan to do that for lots of the functional changes, but I would like some feedback on the external APIs if possible.

rekor_entries.proto
```diff
@@ -0,0 +1,163 @@
...
+     * One or more hash values that can be combined to calculate an inclusion proof
+     */
+    repeated string hashes = 4;
+}
+
+message GetLogEntryRequest {
+    /*
+     * The universally unique identifier for the entry within the log shard
+     */
+    string uuid     = 1;
```
priyawadhwa, line 169: nit: consider renaming uuid to entryID to indicate that we want the universal identifier

priyawadhwa: if we're assuming that the v2 API will only be run with rekor server >=0.6, then we can probably rename all mentions of uuid to entryID?

pkg/types/alpine/alpine.proto
```diff
@@ -0,0 +1,59 @@
...
+     */
+    dev.sigstore.rekor.v2.common.Signature signature = 1;
+    /*
+     * Hash over entire alpine package
+     */
+    dev.sigstore.rekor.v2.common.Hash hash           = 2;
+    /*
+     * Key / value pairs from .PKGINFO file inside alpine package
+     */
+    map<string, string> pkginfo                      = 3;
```
Hayden-IO: Does this let users store arbitrary data?

pkg/types/rfc3161/rfc3161.proto
```diff
@@ -0,0 +1,50 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "RekorRFC3161Proto";
+
+/*
+ * ProposedTimestampRequest represents the structure of the message required to present to Rekor
+ * to make an entry of the TimestampRequest type.
+ *
+ * TimestampRequest uses an identical input and output format;
+ * we create both for potential forward compatibility of separating and changing them;
+ * this also makes the assignment more explicit
```
Hayden-IO, line 32: I'm a little confused by this - So ProposedTImestampRequest is the request, and TimestampRequest is an internal type?

pkg/types/rpm/rpm.proto
```diff
@@ -0,0 +1,59 @@
...
+     */
+    dev.sigstore.rekor.v2.common.Signature signature = 1;
+    /*
+     * Hash over entire RPM package
+     */
+    dev.sigstore.rekor.v2.common.Hash hash           = 2;
+    /*
+     * Key / value pairs from RPM headers
+     */
+    map<string, string> headers                      = 3;
```
Hayden-IO: Same question, does this allow arbitrary data?

rekor_common.proto
```diff
@@ -0,0 +1,70 @@
...
+    SSH                       = 4;
+    PKCS7                     = 5;
+    TUF                       = 6;
+}
+
+message PublicKey {
+    /*
+     * The public key
+     */
+    bytes content = 1;
```
Hayden-IO, line 53: This can contain either a public key, certificate or chain, correct? Should we rename this?

rekor_common.proto
```diff
@@ -0,0 +1,70 @@
...
+package dev.sigstore.rekor.v2.common;
+
+option go_package = "github.com/sigstore/rekor/pkg/generated/protobuf/common";
+option java_package = "dev.sigstore.rekor.v2.common";
+option java_multiple_files = true;
+option java_outer_classname = "RekorCommonProto";
+
+message Signature {
+    /*
+     * The crypto algorithm used to generate the signature
```
Hayden-IO, line 26: nit: This sounds like the signature algorithm, i.e rsa-pss, ecdsa, whereas it's actually the format to hold the signature. Consider rephrasing

rekor_common.proto
```diff
@@ -0,0 +1,70 @@
...
+    PublicKey public_key   = 3;
+}
+
+enum PKIAlgorithm {
+    PKI_ALGORITHM_UNSPECIFIED = 0;
+    PGP                       = 1;
+    MINISIGN                  = 2;
+    X509                      = 3;
+    SSH                       = 4;
+    PKCS7                     = 5;
```
Hayden-IO, line 45: Does this only work with the JAR type, or any type?

rekor_common.proto
```diff
@@ -0,0 +1,70 @@
...
+     * The value of the digital signature
+     */
+    bytes content          = 2;
+    /*
+     * The public key that can be used to verify the signature
+     */
+    PublicKey public_key   = 3;
+}
+
+enum PKIAlgorithm {
```
Hayden-IO: nit: Comments for each enum

rekor_timestamp.proto
```diff
@@ -0,0 +1,78 @@
...
+     * The raw HTTP body is bound to this field; this is an RFC3161 compliant blob.
+     */
+    google.api.HttpBody http_body = 1;
+}
+
+message TimestampTrustBundle {
+    /*
+     * The set of PEM-encoded certificate chains for the timestamp service;
+     * the chain will start with any intermediate certificates (if present),
+     * and finishwith the root certificate.
```
Hayden-IO: finish with

rekor_timestamp.proto
```diff
@@ -0,0 +1,78 @@
...
+          get: "/api/v2/timestamp/certchain"
+        };
+    }
+}
+
+message CreateTimestampRequest {
+    /*
+     * The raw HTTP body is bound to this field; this is an RFC3161 compliant blob.
+     */
+    google.api.HttpBody http_body = 1 [(google.api.field_behavior) = REQUIRED];
```
Hayden-IO, line 53: Can it be a byte array, or does it have to be an HTTP body?

rekor_entries.proto
bobcallaway: yeah, I thought about that as well as normalizing "tree" vs "shard" and trying to make it consistent.

pkg/types/rfc3161/rfc3161.proto
bobcallaway: no, for all types I want to explicitly clarify that they must send `Proposed_______` and they will get `________` as a response. One of the major downsides to the current JSON-schema approach is that it is difficult to interpret what will actually get persisted in the log (by looking at `readOnly` or `writeOnly` tags in JSON schema)... Having two separate types allows for explicit hydration in the code.

pkg/types/rpm/rpm.proto
bobcallaway: same as alpine, the code currently extracts and stores cryptographically-verified metadata from the package headers.

rekor_common.proto
bobcallaway: yeah, this is a carry forward from the current implementation... open to alternative suggestions for sure

rekor_common.proto
bobcallaway: well, it could contain a PGP public key, an X509 certificate (or certificate chain), a TUF metadata root, etc. We could use a oneOf with specific types here if you think its better

rekor_timestamp.proto
bobcallaway: well `http_body` includes a byte array, but in order for rekor to remain an RFC3161 compliant TSA, I think we need to do it this way (could be wrong though, haven't coded it up yet)

pkg/types/rfc3161/rfc3161.proto
bobcallaway: this is a typo, good catch

rekor_common.proto
Hayden-IO: I'd be in support of switching to a oneOf. I think it makes it more explicit and less prone to error.

pkg/types/intoto/v1/intoto.proto
```diff
@@ -0,0 +1,59 @@
...
+option java_multiple_files = true;
+option java_outer_classname = "IntotoV1Proto";
+
+/*
+ * ProposedIntoto represents the structure of the message required to present to Rekor
+ * to make an entry of the Intoto type.
+ */
+message ProposedIntoto {
+    /*
+     * DSSE envelope encompassing the entire Intoto statement
```
asraa, line 33: what about reusing the DSSE proto?
https://github.com/secure-systems-lab/dsse/blob/master/envelope.proto

or too much overhead on clients needing to pull that library?

pkg/types/intoto/v1/intoto.proto
```diff
@@ -0,0 +1,59 @@
...
+     * Public key used to verify signature(s) over envelope
+     */
+    dev.sigstore.rekor.common.PublicKey public_key = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+/*
+ * Intoto represents the structure of the entry persisted in the transparency log for
+ * the Intoto type.
+ */
+message Intoto {
```
asraa, line 46: My biggest issue with this is that you can't take an `Intoto` entry and "verify" it the way that Rekor will verify it.

Can we add the `repeated Signatures signature` in here too? So that clients can verify the entry signature using the Signatures signing the payload_hash?

rekor_common.proto
asraa: FYI this linked issue https://github.com/sigstore/rekor/issues/809 -- would support switching to a oneof

PR #793, opened 2022-04-26
intoto: add index on materials digest of slsa provenance

### LINKED ISSUE #792: Index materials of SLSA provenance statements
https://github.com/sigstore/rekor/pull/764 added support for indexing the payload hash in in-toto entries. It would be nice to do the same for [materials of SLSA provenance attestations](https://slsa.dev/provenance/v0.2#schema), so that it would be possible, for instance, to look up all attestations that shows how a particular git commit (i.e. a material) was built into a binary artifact (i.e. the subject). This would allow for instance to check whether there are issues around reproducibility when done by different builders (e.g. GitHub actions and GCB may end up with a different artifact hash).

cc @bobcallaway @dlorenc @rbehjati @jul-sh @mariaschett @aferr

asraa:

#### Summary

Fixes https://github.com/sigstore/rekor/issues/792

Adds a key to index based on the materials of a SLSA provenance

cc @tiziano88 

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/types/intoto/v0.0.1/entry.go
+++ b/pkg/types/intoto/v0.0.1/entry.go
@@ -92,6 +92,14 @@ func (v V001Entry) IndexKeys() ([]string, error) {
 				result = append(result, alg+":"+ds)
 			}
 		}
+		predicate, err := parseSlsaPredicate(v.env.Payload)
+		if err == nil {
+			for _, s := range predicate.Predicate.Materials {
+				for alg, ds := range s.Digest {
+					result = append(result, alg+":"+ds)
+				}
+			}
+		}
 	default:
 		log.Logger.Infof("Unknown in_toto Statement Type: %s", v.env.PayloadType)
 	}
@@ -108,6 +116,18 @@ func parseStatement(p string) (*in_toto.Statement, error) {
 		return nil, err
 	}
 	return &ps, nil
+}
+
+func parseSlsaPredicate(p string) (*in_toto.ProvenanceStatement, error) {
+	predicate := in_toto.ProvenanceStatement{}
+	payload, err := base64.StdEncoding.DecodeString(p)
+	if err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(payload, &predicate); err != nil {
+		return nil, err
+	}
+	return &predicate, nil
 }
 
 func (v *V001Entry) Unmarshal(pe models.ProposedEntry) error {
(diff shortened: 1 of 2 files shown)


pkg/types/intoto/v0.0.1/entry.go
```diff
@@ -92,6 +92,14 @@ func (v V001Entry) IndexKeys() ([]string, error) {
 				result = append(result, alg+":"+ds)
 			}
 		}
+		predicate, err := parseSlsaPredicate(v.env.Payload)
```
tiziano88: Perhaps add a comment that not all in-toto statements contain SLSA predicates, and provide some supporting links if there are?

pkg/types/intoto/v0.0.1/entry_test.go
```diff
@@ -37,6 +37,8 @@ import (
 	"github.com/go-openapi/swag"
 	"github.com/google/go-cmp/cmp"
 	"github.com/in-toto/in-toto-golang/in_toto"
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+
```
tiziano88: (nit) remove blank line

pkg/types/intoto/v0.0.1/entry.go
```diff
@@ -92,6 +92,14 @@ func (v V001Entry) IndexKeys() ([]string, error) {
...
+		if err == nil {
+			for _, s := range predicate.Predicate.Materials {
```
bobcallaway: should we be a bit more defensive in case `predicate.Predicate` is missing and ensure it is not nil before returning without an error in `parseSlsaPredicate`?

also style nit to use `if predicate, err := parseSlsaPredicate(v.env.Payload); err == nil {` on one line

tiziano88 (dismissed):
awesome, thanks Asra!

pkg/types/intoto/v0.0.1/entry.go
asraa: done re style.


the `in_toto.ProvenanceStatement` always has a `Predicate`. It can be empty. But good catch -- `Materials` could be nil, so added that check.

pkg/types/intoto/v0.0.1/entry.go
asraa: done! i figure we can extend this to something like "getPredicateMaterials" in case the SPDX or other predicates also include materials with digest sets.

dlorenc (approved):

PR #805, opened 2022-05-02
WIP: Add package type

### LINKED ISSUE #804: Add type for generic package
**Description**

As language package managers (PyPI, Ruby Gems, etc) begin to adopt sigstore for signing their packages we may want a generic type that represents a "package.

The initial thought is to leverage https://github.com/package-url/purl-spec in this type.

eddiezane:
#### Summary

This PR adds a new type that represents a generic package from a package manager.

#### Ticket Link

Fixes #804

#### Release Note

```release-note
TODO
```


kommendorkapten:
Is there any progress to this? I'm happy to help if needed.

eddiezane:
> Is there any progress to this? I'm happy to help if needed.

Would love help! Right now we're trying to coordinate with the folks in the [wg-securing-software-repos
](https://github.com/ossf/wg-securing-software-repos) to get this right. There was some initial pushback on using pruls.

cc @znewman01 @di

znewman01:
I would really like to try to do the "do something 3 times before generalizing" workflow here. I wonder if you could have a representative from each of a few ecosystems independently design a package type and then adjust this based on that. I think if you had done that, we could feel pretty confident about the design and get this merged.

Another question—PURLs are  `scheme:type/namespace/name@version?qualifiers#subpath`. Would there be any reason to split those up so they could be searched independently? (Or can our search indexer parse a PURL?)

> There was some initial pushback on using pruls.

I don't recall why that was; do you remember?

di:
(I also don't recall pushback on PURLs)

kommendorkapten:
> I would really like to try to do the "do something 3 times before generalizing" workflow here.

Agree!

From the top of my head, the proposed entries are a good start (signature, digest of package and PURL). I would like to see at least SHA512 being added a supported hash algorithm, as at least npm is using SHA512. 

> Would there be any reason to split those up so they could be searched independently? (Or can our search indexer parse a PURL?)

Yeah, each record type gets to implement an `IndexKeys()` method, so we could parse the PURL there and decide what components to index on. 
An example of this is here for in-toto statements: https://github.com/sigstore/rekor/blob/main/pkg/types/intoto/v0.0.1/entry.go#L73

Issue #806, opened 2022-05-03
Improving Sharding Downtime


priyawadhwa:
Right now, sharding the log requires about 10-20 minutes of downtime.

Sharding process currently looks like this:
1. Create a new trillian tree
2. Mark the current tree as frozen (at this point, the log is readable but not writable)
3. Update rekor config with the following:
    * Change the active tree to the tree created in (1)
    * Update the sharding config with the now-inactive tree ID and length
4. The log is ready to go again!

Downtime occurs between freezing the log & updating the sharding config. We require the length of the inactive shard in the config to properly compute virtual indexes across shards.

**FAQ**

> Why can't we migrate traffic from one shard to the next and then update the sharding config?

In this case, the inactive shard wouldn't be correctly queryable for some portion of time. That's because rekor relies on the sharding config to determine which shards to search and how to determine the "virtual index" across shards. In the period between switching over traffic and updating the sharding config, we could append entries to the log but we wouldn't be able to access entries in the frozen log -- almost all verifications would fail & get by log index would be incorrect.

> What do we use the sharding config for?

The sharding config keeps track of inactive shards and for each tracks:
* ID
* Length
* Associated public key (needed if we rotate keys for new shards)

> Instead of manually creating the sharding config, why don't we automatically search for shards and compute the config in code?

We'd have to guess at how to order the shards (this determines virtual log index computation). We _could_ assume shards should be ordered chronologically, if we're comfortable with that it could reduce downtime.

lkatalin:
Thanks for this recap, @priyawadhwa ! I would be a fan of trying to automate as much of this as possible instead of doing manual shard config / restart. The manual portions introduce the most downtime and also I think there is some possibility for user error down the line if the project continues to grow and scale. Could the issue with trillian trees not necessarily being chronologically ordered be fixed by Rekor code ordering them and keeping track?

lkatalin:
Other ideas floated in the meeting:
- Instead of using the size of the logs to calculate the virt index, have a pointer to the next log
- Create a mapping of log index <--> UUID only during migration-to-new-shard periods

priyawadhwa:
> Could the issue with trillian trees not necessarily being chronologically ordered be fixed by Rekor code ordering them and keeping track?

We can order trees chronologically during runtime, there's a `CreateTime` field on the Tree struct. 

[...]

PR #810, opened 2022-05-04
Retrieve shard tree length if it isn't provided in the config

priyawadhwa:
This should greatly improve the amount of downtime required for sharding 

cc @lkatalin 

cc #806


dlorenc (approved):

lkatalin (approved):
This lgtm! On top of reducing downtime it is good functionality in case the tree lengths are unknown or someone doesn't want to manually specify them.

PR #488, opened 2021-11-12
Fuzzing RequestFromRekor API

naveensrinivasan:
#### Summary

The first steps into fuzzing rekor.
Fuzzing the RequestFromRekor API call using [go-fuzz](https://github.com/dvyukov/go-fuzz) is the standard for fuzzing within `go`. Until go `1.18` is out and until when it is upgraded this could be a good start IMO.

The goal is to integrate this into oss-fuzz using `libfuzzer` https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/ cc @inferno-chromium

```release-note
 Fuzzing API RequestFromRekor 
```

The PR isn't done, I wanted to get some feedback.

PR #813, opened 2022-05-09
Breaking change: Remove timestamping authority

### LINKED ISSUE #812: Remove the timestamping authority before GA
**Description**

TSAs provide an alternative to transparency logs - Witnessing an event at a point in time, but without the ability to audit the signed timestamp issuances. Rekor operates a timestamping authority that:
* Issues signed RFC 3161 timestamps
* Writes the timestamps to the transparency log

The purpose of the TSA is that if you don't want to trust Rekor's internal clock, you can request a signed timestamp, and then provide a signed timestamp to the log. Rekor will include an inclusion time in the entry, but a client can choose to only use the signed timestamp time if they don't want to trust Rekor's clock. However, it adds no value for Rekor to operate both the log and TSA, especially in the same binary - The trust model is the same. You trust Rekor to operate its log without lying about the time, and the timestamping authority to do the same.

If you don't want to trust Rekor's clock, then you should request a signed timestamp from a third party TSA (FreeTSA/Digicert/etc), and include that in the log. Auditing the log can also mitigate concerns around a misconfigured/not honest internal clock.

There's some issues with the current implementation that need to be fixed:
* Doesn't respect the `certReq` bit
* TSA policy ID is invalid, we need to create a policy or use an existing one (and conform to the policy)
* The timestamping signing certificate is generated in memory, meaning after a reboot, there's no way to fetch a certificate chain to validate a previously issue timestamp
* Also because to the above, every instance of Rekor has its own certificate chain. This means it's difficult to fetch the correct certificate chain to validate a timestamp (1/number of instances chance)
* The timestamping certificate is chained off of the Rekor log key - Unexpected reuse of the key in my opinion

According to metrics, this API is barely used - 2/3 requests per minute. If I'm incorrect, please post in the issue if you're actively using the endpoints `/api/v1/timestamp` and `/api/v1/timestamp/certchain`.

We could consider keeping the TSA, but making it optional to run. However, for folks running their own internal instance of Rekor, there's no reason to operate a TSA from a trust perspective, unless you really need RFC 3161 timestamps, though I'd guess at that point you're probably operating an internal TSA already. If there's a use-case, please let me know! We'd still need to make the changes to make it compliant/functional. It's also worth noting that the RFC 3161 is an older RFC. There's newer approaches to trusted timestamps like Roughtime that are more well defined. If there was some need to keep timestamping around, I'd prefer we switch from RFC 3161 to Roughtime.

Hayden-IO:
See #812 for more discussion.

#### Ticket Link

Fixes #812

#### Release Note

```release-note
Removed timestamping authority API. This is a breaking API change. If you are relying on the timestamping authority to issue signed timestamps, create signed timestamps using either OpenSSL or a service such as FreeTSA.
```


dlorenc (dismissed):
I'm happy to remove as is, thanks for clarifying the metrics issue.

Hayden-IO:
Small update to remove unused docker compose env var

dlorenc (dismissed):

dlorenc:
Should we merge this one?

Hayden-IO:
Waiting for @asraa to take a look! Let's wait til EOD tomorrow.

tests/e2e_test.go
```diff
@@ -483,40 +481,6 @@ func TestIntoto(t *testing.T) {
...
-	outputContains(t, out, "Created entry at")
-	uuid := getUUIDFromTimestampOutput(t, out)
-
-	artifactBytes, err := ioutil.ReadFile(tsrPath)
-	if err != nil {
-		t.Error(err)
-	}
-	sha := sha256.Sum256(artifactBytes)
-
-	out = runCli(t, "upload", "--type", "rfc3161", "--artifact", tsrPath)
```
asraa, line 506: We still want to keep the test to upload an rfc 3161 timestamp.

asraa (changes requested):
Thanks! LGTM, minus the e2e test and the timestamp fuzzer needs to be removed.

tests/e2e_test.go
Hayden-IO: Added back!

asraa:
Sorry about the separate ping but I guess the chceks/tools/presubmits for fuzzing need to be removed too
https://github.com/sigstore/rekor/pull/488/files

Hayden-IO:
Let's see if that worked - I made the fuzz target do nothing (Avoiding removing all of the fuzzing additions so it's easy to add fuzz tests in the future)

Hayden-IO:
@asraa @dlorenc - Ready for a last review, I added back an e2e test that tests uploading an rfc3161 timestamp, and removed the timestamp fuzzer (though left all of the fuzzing set up)

CODEOWNERS
```diff
@@ -5,9 +5,6 @@
 
 /pkg/types/		@bobcallaway
 
-/pkg/api/timestamp.go	@asraa @loosebazooka
-/pkg/types/rfc3161/	@asraa @loosebazooka
```
asraa, line 9: keep the rfc type?

asraa (approved):

CODEOWNERS
Hayden-IO: Whoops, I'll create a PR to fix this, sorry!

PR #859, opened 2022-06-02
Allow retrieving entryIDs or UUIDs via `/api/v1/log/entries/retrieve` endpoint

### LINKED ISSUE #855: /api/v1/log/entries/retrieve endpoint does not support the 80 char entryUUID
**Description**

`curl https://rekor.sigstore.dev/api/v1/log/entries/retrieve -H "Content-Type: application/json" -d '{ "entryUUIDs": ["362f8ecba72f432671e977758eba152cd1c11418fd240bf5fb4580fd2d448bb974b0660bebdd4a75"]}'`

`{"code":605,"message":"entryUUIDs.0 in body should match '^[0-9a-fA-F]{64}$'"}`

https://github.com/sigstore/rekor/blob/main/openapi.yaml#L235
Should this have `pattern: '(^[0-9a-fA-F]{64}|[0-9a-fA-F]{80}$)'`

FYI @priyawadhwa 

**Version**

rekor 0.6.0 and 0.7.0

priyawadhwa:
Fixes https://github.com/sigstore/rekor/issues/855 where only 64 char UUIDs were allowed. Also adds in an integration test.

cc @lkatalin could you take a look at this?

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,6 +38,8 @@ services:
       "--appendonly",
       "yes"
     ]
+    ports:
+      - "6379:6379"
     restart: always # keep the redis server running
     healthcheck:
       test: ["CMD", "redis-cli", "ping"]
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -504,7 +504,7 @@ definitions:
         items:
           type: string
           minItems: 1
-          pattern: '^[0-9a-fA-F]{64}$'
+          pattern: '(^[0-9a-fA-F]{64}|[0-9a-fA-F]{80}$)'
       logIndexes:
         type: array
         minItems: 1
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -65,7 +65,7 @@ paths:
             items:
               type: string
               description: Entry UUID in transparency log
-              pattern: '(^[0-9a-fA-F]{64}|[0-9a-fA-F]{80}$)'
+              pattern: '^([0-9a-fA-F]{64}|[0-9a-fA-F]{80})$'
         400:
           $ref: '#/responses/BadContent'
         default:
@@ -214,7 +214,7 @@ paths:
           name: entryUUID
           type: string
           required: true
-          pattern: '(^[0-9a-fA-F]{64}|[0-9a-fA-F]{80}$)'
+          pattern: '^([0-9a-fA-F]{64}|[0-9a-fA-F]{80})$'
           description: the UUID of the entry for which the inclusion proof information should be returned
       responses:
         200:
@@ -504,7 +504,7 @@ definitions:
         items:
           type: string
           minItems: 1
-          pattern: '(^[0-9a-fA-F]{64}|[0-9a-fA-F]{80}$)'
+          pattern: '^([0-9a-fA-F]{64}|[0-9a-fA-F]{80})$'
       logIndexes:
         type: array
         minItems: 1
--- a/pkg/sharding/sharding.go
+++ b/pkg/sharding/sharding.go
@@ -206,7 +206,7 @@ func TreeID(entryID string) (int64, error) {
 	}
 	i, err := strconv.ParseInt(tid, 16, 64)
 	if err != nil {
-		return 0, fmt.Errorf("could not convert treeID %v to int64: %v", tid, err)
+		return 0, fmt.Errorf("could not convert treeID %v to int64: %w", tid, err)
 	}
 	return i, nil
 }
--- a/pkg/sharding/sharding.go
+++ b/pkg/sharding/sharding.go
@@ -198,3 +198,15 @@ func GetTreeIDFromIDString(id string) (string, error) {
 		return "", fmt.Errorf("invalid ID len %v for %v", len(id), id)
 	}
 }
+
+func TreeID(entryID string) (int64, error) {
+	tid, err := GetTreeIDFromIDString(entryID)
+	if err != nil {
+		return 0, err
+	}
+	i, err := strconv.ParseInt(tid, 16, 64)
+	if err != nil {
+		return 0, fmt.Errorf("could not convert treeID %v to int64: %v", tid, err)
+	}
+	return i, nil
+}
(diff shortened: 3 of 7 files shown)


Hayden-IO:
cc @bobcallaway @asraa

docker-compose.yml
```diff
@@ -38,6 +38,8 @@ services:
       "--appendonly",
       "yes"
     ]
+    ports:
```
bobcallaway, line 41: this is probably just a left over change, but just curious as to why it is here.

asraa (dismissed):
Thanks!

dlorenc (dismissed):

openapi.yaml
```diff
@@ -504,7 +504,7 @@ definitions:
         items:
           type: string
           minItems: 1
-          pattern: '^[0-9a-fA-F]{64}$'
+          pattern: '(^[0-9a-fA-F]{64}|[0-9a-fA-F]{80}$)'
```
bobcallaway:
```suggestion
          pattern: '^([0-9a-fA-F]{64}|[0-9a-fA-F]{80})$'
```

tests/sharding-e2e-test.sh
```diff
@@ -217,7 +217,20 @@ fi
...
+UUID2=$($REKOR_CLI get --log-index 3 --rekor_server http://localhost:3000 --format json | jq -r .UUID)
+
+HEX_INITIAL_TREE_ID=$(printf "%x" $INITIAL_TREE_ID)
+HEX_INITIAL_SHARD_ID=$(printf "%x" $SHARD_TREE_ID)
+
+ENTRY_ID_1=$HEX_INITIAL_TREE_ID$UUID1
+ENTRY_ID_2=$HEX_INITIAL_SHARD_ID$UUID2
+
+# -f makes sure we exit on failure
+curl -f http://localhost:3000/api/v1/log/entries/retrieve -H "Content-Type: application/json" -d "{ \"entryUUIDs\": [\"$ENTRY_ID_1\", \"$ENTRY_ID_2\"]}"
```
bobcallaway: this will fail if we get a non 200 response; should we actually pipe to jq and test that we get 2 entries in return? If redis were to be empty we'd return a HTTP 200 with an empty array AFAIK.

bobcallaway (commented):
a couple comments

docker-compose.yml
priyawadhwa: I actually find this really useful for dev/debugging -- i spin up rekor in docker compose, run a `docker stop rekor_server` and then i can run the server on the command line via

```
go run ./cmd/rekor-server serve
```

It still interacts with the redis/trillian containers bc of the port forwarding and it's faster than having to rebuild the rekor server container every time I want to test a change.

I actually left it in here on purpose in case others do the same!

tests/sharding-e2e-test.sh
priyawadhwa: good call, made the change and actually caught a bug! thanks 💯

openapi.yaml
priyawadhwa: thanks! fixed in a couple different spots.

docker-compose.yml
bobcallaway: ok, that makes sense to me.

pkg/sharding/sharding.go
```diff
@@ -198,3 +198,15 @@ func GetTreeIDFromIDString(id string) (string, error) {
...
 }
+
+func TreeID(entryID string) (int64, error) {
+	tid, err := GetTreeIDFromIDString(entryID)
+	if err != nil {
+		return 0, err
+	}
+	i, err := strconv.ParseInt(tid, 16, 64)
+	if err != nil {
+		return 0, fmt.Errorf("could not convert treeID %v to int64: %v", tid, err)
```
bobcallaway:
```suggestion
		return 0, fmt.Errorf("could not convert treeID %v to int64: %w", tid, err)
```

tests/sharding-e2e-test.sh
```diff
@@ -217,7 +217,20 @@ fi
...
+UUID2=$($REKOR_CLI get --log-index 3 --rekor_server http://localhost:3000 --format json | jq -r .UUID)
+
+HEX_INITIAL_TREE_ID=$(printf "%x" $INITIAL_TREE_ID | awk '{printf "%016s\n", $0}')
+HEX_INITIAL_SHARD_ID=$(printf "%x" $SHARD_TREE_ID | awk '{printf "%016s\n", $0}')
+
+ENTRY_ID_1=$HEX_INITIAL_TREE_ID$UUID1
+ENTRY_ID_2=$HEX_INITIAL_SHARD_ID$UUID2
+
+# -f makes sure we exit on failure
+NUM_ELEMENTS=$(curl -f http://localhost:3000/api/v1/log/entries/retrieve -H "Content-Type: application/json" -d "{ \"entryUUIDs\": [\"$ENTRY_ID_1\", \"$ENTRY_ID_2\"]}" | jq '. | length')
```
bobcallaway: You may also want to add `-H "Accept: application/json"` to help protect against yaml being returned (which could theoretically happen, and we've seen it happen based on the unordered map key issue).

priyawadhwa: thanks, done!

dlorenc (approved):

lkatalin (approved):
@priyawadhwa Looks like high level overview is this allows the 80-char entryID for searches in `openapi.yaml` and then in the search code itself makes sure the treeID is appropriate prepended so the search can be routed correctly?

Not sure why but my searches always worked so I didn't catch that this should also be updated. Thanks for the fix.

priyawadhwa:
@lkatalin yup, this is specifically for the `/api/v1/log/entries/retrieve` endpoint -- I don't think `rekor-cli` hits that endpoint except for `rekor-cli verify` which might be why I didn't notice this in testing.

lkatalin (approved):

bobcallaway (approved):

PR #864, opened 2022-06-06
Print total tree size, including inactive shards in `rekor-cli loginfo`

priyawadhwa:

#### Ticket Link

Fixes

#### Release Note

```release-note
BREAKING CHANGE: Rename `TreeSize` in `rekor-cli loginfo` output to `ActiveTreeSize`. This helps clarify that this is the length of the active tree. Also add in `TotalTreeSize`, which is the length of the entire log across all inactive and active shards.
```

### DIFF
--- a/cmd/rekor-cli/app/log_info.go
+++ b/cmd/rekor-cli/app/log_info.go
@@ -44,6 +44,7 @@ import (
 
 type logInfoCmdOutput struct {
 	TreeSize       int64
+	TotalTreeSize  int64
 	RootHash       string
 	TimestampNanos uint64
 	TreeID         string
@@ -54,11 +55,12 @@ func (l *logInfoCmdOutput) String() string {
 	ts := time.Unix(0, int64(l.TimestampNanos)).UTC().Format(time.RFC3339)
 
 	return fmt.Sprintf(`Verification Successful!
-Tree Size: %v
-Root Hash: %s
-Timestamp: %s
-TreeID:    %s
-`, l.TreeSize, l.RootHash, ts, l.TreeID)
+Tree Size:       %v
+Total Tree Size: %v
+Root Hash:       %s
+Timestamp:       %s
+TreeID:          %s
+`, l.TreeSize, l.TotalTreeSize, l.RootHash, ts, l.TreeID)
 }
 
 // logInfoCmd represents the current information about the transparency log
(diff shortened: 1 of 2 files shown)


cmd/rekor-cli/app/log_info.go
```diff
@@ -44,6 +44,7 @@ import (
 
 type logInfoCmdOutput struct {
 	TreeSize       int64
+	TotalTreeSize  int64
```
asraa, line 47: Could you add some documentation either here or in the output that `Tree Size` is the active shard size and `Total Tree Size` is Active and Inactive Tree?

Maybe Tree Size -> Active Tree Size

asraa (dismissed):
Otherwise LGTM!

cmd/rekor-cli/app/log_info.go
priyawadhwa: I think `ActiveTreeSize` would be a better name! I was worried about making a breaking change in case people are grabbing `TreeSize` via `jq` or something, but I think renaming it is definitely more clear.

asraa: Oh good point! Let's add that to the release notes.

dlorenc: +1 on doing the breaking change now before 1.0 and putting it in the release notes!

cmd/rekor-cli/app/log_info.go
```diff
@@ -54,11 +55,12 @@ func (l *logInfoCmdOutput) String() string {
 	ts := time.Unix(0, int64(l.TimestampNanos)).UTC().Format(time.RFC3339)
 
 	return fmt.Sprintf(`Verification Successful!
-Tree Size: %v
-Root Hash: %s
-Timestamp: %s
-TreeID:    %s
-`, l.TreeSize, l.RootHash, ts, l.TreeID)
+Tree Size:       %v
```
lkatalin: "Active" here too? Otherwise lgtm

cmd/rekor-cli/app/log_info.go
priyawadhwa: SGTM, added in the release note.

dlorenc (approved):

asraa (approved):

PR #731, opened 2022-03-13
WIP: Add COSE support to Rekor

dlorenc:
This commit adds COSE Sign1 support to rekor via a new data type.
COSE is defined in RFC8152, and provides a signing message envelope.

This is supported in rekor using the veraison/go-cose library. The new
API type requires the signed content, the signature envelope, and the public key.

The public key is in the standard rekor PKI format, at the moment only ECDSA P256
with SHA256 is supported. The signed message only supports in-line bodies (no URL fetching),
and there is no support for pre-hashed entries.

Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

PR #867, opened 2022-06-09
Add COSE support to Rekor

kommendorkapten:
#### Summary
Basic support for COSE envelopes.
Added is functionality to upload COSE records, optionally signed with AAD. The AAD is not stored in Rekor.
The COSE envelope is stored as an attestation if attestation storage is enabled, and can so be retrieved via Rekor. The returned envelope is returned in the `attestation` field, not within the `body` field. 

What's stored is the hash of the payload and the entire envelope. If the content type of the COSE envelope is an in-toto statement, the payload is parsed and the subject hashes are extracted and indexed for searching.

#### Ticket Link
This pull requests completes https://github.com/sigstore/rekor/pull/731.

#### Release Note

```release-note
Added new request/response type: COSE envelopes. When uploading a COSE envelope, only the hash of the payload and the hash of the envelope is stored, unless attestation storage is enabled, then the entire envelope is stored as an attestation.
A new command line parameter `--aad` is added. This parameter provides extra Additional Authenticated Data to be used during COSE signature verification (using AAD with COSE is optional).
```


cmd/rekor-cli/app/pflag_groups.go
```diff
@@ -87,6 +87,11 @@ func addArtifactPFlags(cmd *cobra.Command) error {
 			"path or URL to pre-formatted entry file",
 			false,
 		},
+		"aad": {
+			base64Flag,
+			"base64 encoded aad",
```
bobcallaway: you may want to spell out what `aad` means here, since this will be printed as part of command line flag help output.

pkg/types/cose/cose.go
```diff
@@ -0,0 +1,73 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
bobcallaway: can you switch to 2022 throughout all new files please?

dlorenc:
Looks like there's a merge conflict, and github isn't running for some reason. Maybe fixing the merge conflict will poke actions?

kommendorkapten:
Yes, will do. Some merge conflicts I have to resolve first 👍

pkg/types/cose/v0.0.1/entry.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	pk := v.keyObj.(*x509.PublicKey)
+	cryptoPub := pk.CryptoPubKey()
+
+	var alg gocose.Algorithm
+	switch t := cryptoPub.(type) {
+	case *rsa.PublicKey:
+		alg = gocose.AlgorithmPS256
+	case *ecdsa.PublicKey:
```
dlorenc: Do we need to be a bit more careful here to avoid curve mismatches? Is there a possibility someone actually used PS384 or PS512?

pkg/types/cose/v0.0.1/entry.go
```diff
@@ -0,0 +1,326 @@
...
+	if v.sign1Msg != nil {
+		result = append(result, formatKey(v.sign1Msg.Payload))
+	} else {
+		// If no payload exists (it's unpacked in validate() method)
+		// return now, as we will not be able to extract any headers
+		return result, nil
+	}
+
+	// If payload is an in-toto statement, let's grab the subjects.
+	if rawContentType, ok := v.sign1Msg.Headers.Protected[gocose.HeaderLabelContentType]; ok {
```
dlorenc, line 108: This part is interesting, I'm not sure I've seen in-toto statements embedded into COSE envelopes before. I think it makes sense overall though.

dlorenc (commented):
One main question on algorithm checking.

pkg/types/cose/v0.0.1/entry.go
kommendorkapten: Yeah, good catch. I can play around a bit more early next week to see if we can tighten up this.

dlorenc:
Overall LGTM. I'll wait for @bobcallaway to take a look before merging.

dlorenc:
Overall LGTM. I'll wait for @bobcallaway to take a look before merging.

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/mediocregopher/radix/v4 v4.1.0
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/mapstructure v1.5.0
+	github.com/pkg/errors v0.9.1
```
bobcallaway: there was a recent effort to stop using this and use `errors` instead.

pkg/types/cose/cose.go
```diff
@@ -0,0 +1,73 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cose
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
```
bobcallaway: can you swap this with `"errors"` instead?

pkg/types/cose/v0.0.1/entry.go
```diff
@@ -0,0 +1,350 @@
...
+	// This also gets called in the CLI, where we won't have this data
+	// or during record retrieval (message is the raw COSE object) which
+	// is only stored as an attestation.
+	if len(v.CoseObj.Message) == 0 {
+		return nil
+	}
+
+	alg, pk, err := getPublicKey(v.keyObj)
+	if err != nil {
+		return nil
```
bobcallaway: did you mean to return `err` here?

pkg/types/cose/v0.0.1/entry.go
```diff
@@ -0,0 +1,350 @@
...
+
+	bv, err := gocose.NewVerifier(alg, pk)
+	if err != nil {
+		return err
+	}
+	v.sign1Msg = gocose.NewSign1Message()
+	if err := v.sign1Msg.UnmarshalCBOR(v.CoseObj.Message); err != nil {
+		return err
+	}
+
```
bobcallaway: should we ensure unmarshaling is successful before assigning the output to `v.sign1Msg`?

pkg/types/entries.go
```diff
@@ -121,12 +121,13 @@ func CanonicalizeEntry(ctx context.Context, entry EntryImpl) ([]byte, error) {
...
 type ArtifactProperties struct {
-	ArtifactPath   *url.URL
-	ArtifactHash   string
-	ArtifactBytes  []byte
-	SignaturePath  *url.URL
-	SignatureBytes []byte
-	PublicKeyPath  *url.URL
-	PublicKeyBytes []byte
-	PKIFormat      string
+	AdditionalAuthenticatedData string
```
bobcallaway: should this get changed to `[]byte` and leave it up to the caller to decode the input correctly?

bobcallaway (commented):
looking good overall, just a few questions/nits

pkg/types/cose/cose.go
kommendorkapten: Of course, my bad!

pkg/types/cose/v0.0.1/entry.go
kommendorkapten: Of course, will go over some test coverage here to make sure we got it covered.

pkg/types/cose/v0.0.1/entry.go
kommendorkapten: You mean like this:
```go
        coseMsg := gocose.NewSign1Message()
	if err := coseMsg.UnmarshalCBOR(v.CoseObj.Message); err != nil {
		return err
	}
        v.sign1Msg = coseMsg
```

I think that makes sense, we can even wait until the signature is verified until we assign it to `v.sign1Msg`.

pkg/types/entries.go
kommendorkapten: Yes, that makes sense! So basically moving the base64 decoding into the cli part instead of `CreteFromArtifactProperties`. I'll give it a shot to see if I can make it happen.

cmd/rekor-cli/app/pflag_groups.go
```diff
@@ -157,7 +158,10 @@ func CreatePropsFromPflags() *types.ArtifactProperties {
 	}
 
 	props.PKIFormat = viper.GetString("pki-format")
-	props.AdditionalAuthenticatedData = viper.GetString("aad")
+	b64aad := viper.GetString("aad")
+	if b64aad != "" {
+		props.AdditionalAuthenticatedData, _ = base64.StdEncoding.DecodeString(b64aad)
```
bobcallaway, line 163: is there a reason you're squelching this error here?

cmd/rekor-cli/app/pflags_test.go
```diff
@@ -346,6 +347,23 @@ func TestArtifactPFlags(t *testing.T) {
 			expectParseSuccess:    true,
 			expectValidateSuccess: false,
 		},
+		{
```
bobcallaway, line 350: maybe one more test case for invalid base64 data passed on the command line?

bobcallaway (commented):
LGTM, would like one more test case to ensure we have coverage for non-base64 encoded input on the `aad` cmd line argument

pkg/types/cose/v0.0.1/entry.go
bobcallaway: yes, ideally I wouldn't want to partially change the pointer receiver unless all validation is successful.

bobcallaway (dismissed):

kommendorkapten:
The sharding e2e test failed (`curl: (22) The requested URL returned error: 422 Unprocessable Entity`). Is there any documentation on running this locally?

kommendorkapten:
Oh, found a script, will try that.

kommendorkapten:
I have run the sharding-e2e test now locally three times and am not able to reproduce, the execution succeeds all the time for me. Anyone having an idea?

kommendorkapten:
Hm, this seems to be the issue:
```
++ printf %x 7051007558037933689
++ Awk '{printf "%016s", $0}'
+ HEX_INITIAL_TREE_ID=61da360425720e79
++ printf %x 151938258151519870
++ awk '{printf "%016s", $0}'
+ HEX_INITIAL_SHARD_ID=' 21bcb0a70e9b67e'
+ ENTRY_ID_1=61da360425720e79d2f305428d7c222d7b77f56453dd4b6e6851752ecacc78e5992779c8f9b61dd9
+ ENTRY_ID_2=' 21bcb0a70e9b67e59a575f157274702c38de3ab1e1784226f391fb79500ebf9f02b4439fb77574c'
++ curl -f http://localhost:3000/api/v1/log/entries/retrieve -H 'Content-Type: application/json' -H 'Accept: application/json' -d '{ "entryUUIDs": ["61da360425720e79d2f305428d7c222d7b77f56453dd4b6e6851752ecacc78e5992779c8f9b61dd9", " 21bcb0a70e9b67e59a575f157274702c38de3ab1e1784226f391fb79500ebf9f02b4439fb77574c"]}'
++ jq '. | length'
```

The shard id gets a white space (` `) as the first character, which is then input to the UUIDs: `" 21bcb0a70e9b67e59a575f157274702c38de3ab1e1784226f391fb79500ebf9f02b4439fb77574c"`.

Running the printf/awk on my machine (macOS) I get a different output
```
 printf %x 151938258151519870 | awk '{printf "%016s", $0}'
021bcb0a70e9b67e
```
Running same command on Ubuntu:
```
# printf %x 151938258151519870 | awk '{printf "%016s", $0}'
 21bcb0a70e9b67e
```
 I'll prepare a patch for the awk script.

bobcallaway:
I'll push up a separate PR to fix that, and also add log collection to
sharding-e2e-test.sh since if there is a failure right now we don't capture
the docker-compose logs.

kommendorkapten:
From the GNU awk manual (https://www.gnu.org/software/gawk/manual/html_node/Format-Modifiers.html): (emphasis mine)
> A leading ‘0’ (zero) acts as a flag indicating that output should be padded with zeros instead of spaces. This applies only to the *numeric output formats*. This flag only has an effect when the field width is wider than the value to print.

kommendorkapten:
Shamelessly copied this from stackoverflow (some minor modification to work with `printf` instead of `echo`):
```
printf abc123 | awk '{ for(c = 0; c < 16 ; c++) s = s"0"; s = s$1; print substr(s, 1 + length(s) - 16);}'
0000000000abc123
```

Tested on both Ubuntu and macOS

bobcallaway:
> Shamelessly copied this from stackoverflow (some minor modification to work with `printf` instead of `echo`):
> 
> ```
> printf abc123 | awk '{ for(c = 0; c < 16 ; c++) s = s"0"; s = s$1; print substr(s, 1 + length(s) - 16);}'
> 0000000000abc123
> ```
> 
> Tested on both Ubuntu and macOS

@priyawadhwa @lkatalin is the intent for the value to be left-padded with zeros?

lkatalin:
> @priyawadhwa @lkatalin is the intent for the value to be left-padded with zeros?

@bobcallaway 
Still going through trying to understand this error, but the [padding](https://github.com/sigstore/rekor/blob/45da1afab938270c881d0ce2cee8b2169fb51289/pkg/sharding/sharding.go#L85) is something I had included so that we can get a [consistent length](https://github.com/sigstore/rekor/blob/45da1afab938270c881d0ce2cee8b2169fb51289/pkg/sharding/sharding.go#L32), probably so that we can easily parse the TreeID from the UUID.

dlorenc (approved):

Issue #854, opened 2022-06-01
IndexKeys is run on pre-canonicalized entries


asraa:
**Description**

**Version**

The flow of creating a log entry is curious to me:
https://github.com/sigstore/rekor/blob/4b934999dc28386d22dfa3004930767d203e7580/pkg/api/entries.go#L173

1. A new entry is created based on the parameter proposed entry.
2. A leaf node is created by canonicalizing the entry
3. The original entry calls index keys.

In many entry types, logic is performed in `Canonicalize` to set up the entry, and it's not clear that `IndexKeys`  does not have access to the canonicalized entry fields. This is also a problem if Canonicalize does some operation that may modify IndexKeys.

I realized this was the root cause to the bug that @priyawadhwa fixed from https://github.com/sigstore/rekor/pull/800 where assumed that intoto entries populated the Content Hash (but only canonicalized ones did).

Shouldn't `IndexKeys` operate on the canonicalized entry?

asraa:
From @haydentherapper :+1: 

"Maybe we need to figure out how to do canonicalization very early on in the lifecycle of a request

so nothing ever operates on a non-canonical entry"

dlorenc:
I'm not sure this is doable since we strip a lot of information during canonicalization, on purpose. It means the index entries aren't always verifiable or repairable, though.

PR #870, opened 2022-06-15
Fix indexing bug for intoto attestations

priyawadhwa:
If the `IntotoObj.Content` field isn't specified, we end up returning too soon instead of adding provenance subject digests as indexing keys. `cosign attest` never sets the Content field, so searching for attestations uploaded with cosign isn't working.

This should fix it and adds in a unit test.


dlorenc (approved):

bobcallaway (approved):

asraa:
Again, the root cause is because we index on pre-canonicalized entries.
https://github.com/sigstore/rekor/issues/854

PR #897, opened 2022-06-30
feat: add subject URIs to index for x509 certificates

asraa:

#### Summary

Adds URIs to x509 certificate "email addresses". This picks up on job workflow ref for Fulcio issued certificates.

Ideally I'd like to deprecate `EmailAddresses()` in favor of `Subjects()` -- should I add `Subjects()` and mark `EmailAddresses()` as deprecated? I'm not sure there's much external usage.

#### Ticket Link

Fixes

#### Release Note

```release-note
* pki: Deprecates `pki.PublicKey.EmailAddresses()`. Use `pki.PublicKey.Subjects()` instead to include subject URIs from public keys like x509 certificates.
```

### DIFF
--- a/pkg/pki/x509/x509.go
+++ b/pkg/pki/x509/x509.go
@@ -204,14 +204,12 @@ func (k PublicKey) Subjects() []string {
 	if cert != nil {
 		validate := validator.New()
 		for _, name := range cert.EmailAddresses {
-			errs := validate.Var(name, "required,email")
-			if errs == nil {
+			if errs := validate.Var(name, "required,email"); errs == nil {
 				names = append(names, strings.ToLower(name))
 			}
 		}
 		for _, name := range cert.URIs {
-			errs := validate.Var(name.String(), "required,uri")
-			if errs == nil {
+			if errs := validate.Var(name.String(), "required,uri"); errs == nil {
 				names = append(names, strings.ToLower(name.String()))
 			}
 		}
--- a/pkg/pki/x509/x509.go
+++ b/pkg/pki/x509/x509.go
@@ -188,6 +188,13 @@ func (k PublicKey) EmailAddresses() []string {
 				names = append(names, strings.ToLower(name))
 			}
 		}
+		for _, name := range cert.URIs {
+			validate := validator.New()
+			errs := validate.Var(name.String(), "required,uri")
+			if errs == nil {
+				names = append(names, strings.ToLower(name.String()))
+			}
+		}
 	}
 	return names
 }
--- a/pkg/pki/x509/x509.go
+++ b/pkg/pki/x509/x509.go
@@ -181,15 +181,35 @@ func (k PublicKey) EmailAddresses() []string {
 		cert = k.certs[0]
 	}
 	if cert != nil {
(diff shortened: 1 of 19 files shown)


pkg/pki/x509/x509.go
```diff
@@ -188,6 +188,13 @@ func (k PublicKey) EmailAddresses() []string {
 				names = append(names, strings.ToLower(name))
 			}
 		}
+		for _, name := range cert.URIs {
+			validate := validator.New()
+			errs := validate.Var(name.String(), "required,uri")
+			if errs == nil {
+				names = append(names, strings.ToLower(name.String()))
+			}
+		}
```
bobcallaway: nit:
```suggestion
			if errs := validate.Var(name.String(), "required,uri"); errs == nil {
				names = append(names, strings.ToLower(name.String()))
			}
		}
```

pkg/pki/x509/x509.go
```diff
@@ -188,6 +188,13 @@ func (k PublicKey) EmailAddresses() []string {
 				names = append(names, strings.ToLower(name))
 			}
 		}
+		for _, name := range cert.URIs {
+			validate := validator.New()
```
bobcallaway: also, should we create `validate` once rather than churn objects across multiple loops?

bobcallaway:
> Ideally I'd like to deprecate `EmailAddresses()` in favor of `Subjects()` -- should I add `Subjects()` and mark `EmailAddresses()` as deprecated? I'm not sure there's much external usage.

I'm fine with that.

pkg/pki/x509/x509.go
imjasonh: Are these validation errors surfaced anywhere before/after this? It seems a waste to drop perfectly good validation errors from end users (...says the end user most likely to send invalid values 🙃 )

imjasonh:
> Ideally I'd like to deprecate `EmailAddresses()` in favor of `Subjects()` -- should I add `Subjects()` and mark `EmailAddresses()` as deprecated? I'm not sure there's much external usage.

I looked, and [`sget` uses `EmailAddresses()`](https://github.com/sigstore/sget/blob/b46a2aa4debc2683d7206221007231726b72b396/main.go#L140), but I don't mind having to switch when I bump the dep.

pkg/pki/x509/x509.go
asraa: This ones a little tougher to surface -- I expect that this one is relatively difficult to hit, because Fulcio is the one populating certificates and it's not user-controlled. 

On that note maybe it's worth filing an issue that we surface what keys WERE successfully indexed when you upload something?

asraa:
Added `Subjects()` and a deprecation notice on `EmailAddresses()`

pkg/pki/x509/x509.go
Hayden-IO: FWIW, it's also worth noting that it might not be Fulcio's certs uploaded to the log, as users can upload certs from their own PKI.

imjasonh (approved):

dlorenc (approved):

Issue #902, opened 2022-07-04
createdb.sh contains wrong SQL directive


xens:
**Description**

I was trying to deploy a rekor server from scratch using a MariaDB backend with the `scripts/createdb.sh` script but the script contains a typo that prevents the provisioning of the user.

**Version**

I was using the script from the master branch.

bobcallaway:
Thanks for the fix!

PR #903, opened 2022-07-04
fix: sql syntax in dbcreate script

xens:
#### Summary

This commit fixes an issue in the mariadb createdb script that prevents it to work properly #902.


bobcallaway (approved):

lukehinds (approved):
well spotted, thanks

PR #904, opened 2022-07-05
Switch to go 1.18 and pin release-utils to v0.7.1

saschagrunert:
#### Summary
We should use the release versions rather than the commits on `main` to let dependabot update it automatically.

#### Release Note

NONE

#### Documentation

NONE


cpanato:
I think we need to have go1.18

cpanato (approved):
thank you!

lukehinds:
I know this is merged now, but we need to agree on a go versioning strategy. I am of the mind that we should pin go.mod to the last supported release and then test against newer releases. When a go version deprecates, we then move up one release within go.mod. Either way, even if that is not seen as the correct approach, we right now we have some projects running go.1.17 (sigstore/sigstore) and others running go.1.18

I will put it on the community agenda for next week.

saschagrunert:
As a side note, in Kubernetes we follow the principle of always using the latest available go version, because this allows us to update critical dependencies when they require a recent go version, too.

dlorenc:
The main driver for 1.18 has been the hashicorp dependencies that require it.

PR #878, opened 2022-06-18
ensure fallback logic executes if attestation key is empty when fetching attestation

bobcallaway:
Fixes: #872

Signed-off-by: Bob Callaway <bcallaway@google.com>

PR #909, opened 2022-07-06
Remove unnecessary lookup of non-existent attestations from storage layer

bobcallaway:
Currently only two Rekor pluggable types support the storage of attestations (`intoto`, `cose`); the previous code to fetch attestations was type-agnostic, but due to the fix #878, the server was doing unnecessary lookups for all types, regardless of whether they store attestation content or not.

This makes the attestation storage an explicit interface, which we can test casting for and avoid a round trip to the storage layer for types that don't support storing attestations.

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -106,31 +106,33 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 			return nil, err
 		}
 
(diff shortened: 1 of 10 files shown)


dlorenc:
Oooh nice!

dlorenc (approved):

priyawadhwa (approved):

PR #915, opened 2022-07-11
Return 404 if entry isn't found in log

priyawadhwa:

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -299,8 +299,11 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
 func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware.Responder {
 	logEntry, err := retrieveLogEntry(params.HTTPRequest.Context(), params.EntryUUID)
 	if err != nil {
+		if errors.Is(err, ErrNotFound) {
+			return handleRekorAPIError(params, http.StatusNotFound, err, "")
+		}
 		if _, ok := (err).(types.ValidationError); ok {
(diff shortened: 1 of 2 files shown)


pkg/api/entries.go
```diff
@@ -299,6 +299,9 @@ func getEntryURL(locationURL url.URL, uuid string) strfmt.URI {
 func GetLogEntryByUUIDHandler(params entries.GetLogEntryByUUIDParams) middleware.Responder {
 	logEntry, err := retrieveLogEntry(params.HTTPRequest.Context(), params.EntryUUID)
 	if err != nil {
+		if errors.Is(err, ErrNotFound) {
+			return handleRekorAPIError(params, http.StatusNotFound, err, "")
+		}
 		if _, ok := (err).(types.ValidationError); ok {
 			return handleRekorAPIError(params, http.StatusBadRequest, err, "incorrectly formatted uuid %s", params.EntryUUID)
```
imjasonh: I think this will also need a `fmt.Sprintf("incorrect formatted uuid %s", params.EntryUUID)`, otherwise the error message includes the `%s`.

~~(I couldn't suggest a change since this line didn't change, thanks GitHub! 🤬)~~

imjasonh:
```suggestion
			return handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf("incorrectly formatted uuid %s", params.EntryUUID), params.EntryUUID)
```

priyawadhwa: good catch! added :)

imjasonh (approved):

asraa (approved):
cc @loosebazooka hit this i think

PR #921, opened 2022-07-13
Add rekor test harness to presubmit tests

priyawadhwa:
This will test critical tests against the last three versions of rekor released.

I picked out a subset of the current e2e tests which I think are important to run against previous versions. 

addresses https://github.com/sigstore/public-good-instance/issues/417


tests/rekor-harness.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+#
+# Copyright 2021 The Sigstore Authors.
```
bobcallaway:
```suggestion
# Copyright 2022 The Sigstore Authors.
```

bobcallaway (commented):
IIUC, this uses the source at head to compile a new `rekor-cli` that is then used against prior versions of the server? I'm not sure how frequent of a situation that will be.

I was expecting to see something more similar to this:
- Define a queue of versions to test (similar to how you've picked the last 3 versions here). Call that sorted (ascending) queue `Q`
- Define a set of versions you have tested, let's call that `S`, initially empty
- Start by popping a version off of the queue `Q` and adding it to the set `S`
- let `v=max(S)`
- start the server with `docker-compose up`  `rekor-server` and dependencies (redis, mysql, trillian) for version `v`, 
- for all values in `S`, run `HARNESS_TESTS` using the `rekor-cli` version `v` (we should also capture entryIDs for all inserted entries, and ensure all are readable with `rekor-cli` versions in `S`)
- Pop the next version off of the queue `Q` and add to `S` 
- Repeat this process, assuming the database and trillian instances are persisted across `rekor-server` version upgrades (similar to what was done in sharding). This is critical, since it mirrors what will happen in prod.

priyawadhwa:
@bobcallaway so I set up something more like what you were describing:

- take the last three versions (example v1, v2, v3)
- start by setting up the server at v1, & test against all 3 versions of the CLI
- when upgrading the server for v2 and v3, restart just rekor-server and maintain the same mysql/trillian backend. test each against all 3 versions of the CLI as well.

does that seem reasonable as a skeleton?

priyawadhwa:
update: i have tests which should more closely mimic updates in prod. for every server version upgrade, they'll add new entries and also make sure they can still get all the old entries.

i might add that in a follow-up pr so this one doesn't get too big

tests/rekor-harness.sh
```diff
@@ -0,0 +1,103 @@
...
+    done
+    echo
+}
+
+function build_cli () {
+    echo "Building CLI at version $cli_version"
+    cli_version=$1
+    current_branch=$(git rev-parse --abbrev-ref HEAD)
+    git checkout $cli_version
+    go build -o rekor-cli ./cmd/rekor-cli
```
bobcallaway: Should we do `make rekor-cli` to reuse any other build flags, etc?

tests/rekor-harness.sh
```diff
@@ -0,0 +1,103 @@
...
+            echo "Failing due to panics detected in logs"
+            docker-compose logs --no-color > /tmp/docker-compose.log
+            exit 1
+        fi
+    done
+}
+
+# Get last 3 server versions
+git fetch origin
+VERSIONS=$(git tag --sort=-version:refname | head -n 3 | tac)
```
bobcallaway, line 82: what is `tac` here?

tests/rekor-harness.sh
```diff
@@ -0,0 +1,103 @@
...
+function run_tests () {
+    REKORTMPDIR="$(mktemp -d -t rekor_test.XXXXXX)"
+    touch $REKORTMPDIR.rekor.yaml
+    trap "rm -rf $REKORTMPDIR" EXIT
+
+
+    go clean -testcache
+    for test in $HARNESS_TESTS
+    do
+        if ! REKORTMPDIR=$REKORTMPDIR go test -run $test -v -tags=e2e ./tests/ > logs ; then
```
bobcallaway: you could write `logs` into `$REKORTMPDIR` and save the `.gitignore` change

bobcallaway (commented):
For inserts at version X and testing if we can run the exact same insert at version X+1, I think this does it. As you note this doesn't cover reading back previously inserted values inserted under code versions that were backlevel from the proposed change.

tests/rekor-harness.sh
priyawadhwa: sure, replaced

tests/rekor-harness.sh
priyawadhwa: thanks!

tests/rekor-harness.sh
priyawadhwa: it reverses the order of strings printed out by `git tag --sort=-version:refname | head -n 3` so that we get the versions in chronological order. that way, we start the server at v0.8.1, then update to v0.8.2 and onwards as we would in prod.

priyawadhwa:
Thanks @bobcallaway! Addressed comments, and I'll open a follow-up PR once this is in to check reading previously inserted values.

tests/rekor-harness.sh
dlorenc: heh - it's just the reverse of `cat`

dlorenc (approved):

tests/rekor-harness.sh
bobcallaway: my mind went to another definition of the term :) all good

bobcallaway (approved):

PR #933, opened 2022-07-21
Drop application/yaml content type

### LINKED ISSUE #593: Rekor API sometimes responds with YAML instead of JSON
### To reproduce

1. Load https://rekor.sigstore.dev/api/v1/log/ in a browser (I'm using `Chrome 96.0.4664.55 (Official Build) (x86_64)`)
2. Reload until eventually the page will be downloaded as `download` (or `download (1)`, `(2)`, etc., if you keep reloading)

Looking at the Network tab in Chrome Devtools, it looks like response headers randomly change from `content-type: application/json;q=1` (no download) to `content-type: application/yaml` (download).

I'm not sure if this is hitting different backends that set the response `Content-Type` differently, but it's weird in any case.

You can also observe this using `curl`, and for other API methods:

```
$ curl -v https://rekor.sigstore.dev/api/v1/log/entries\?logIndex\=1045462 2>&1 | grep yaml
$ curl -v https://rekor.sigstore.dev/api/v1/log/entries\?logIndex\=1045462 2>&1 | grep yaml
$ curl -v https://rekor.sigstore.dev/api/v1/log/entries\?logIndex\=1045462 2>&1 | grep yaml
< content-type: application/yaml
$ curl -v https://rekor.sigstore.dev/api/v1/log/entries\?logIndex\=1045462 2>&1 | grep yaml
```

Hayden-IO:
This is a breaking change for any clients that are currently requesting
application/yaml.

This is the simplest fix to #593, as openapi randomly switches between
json and yaml responses when no content type is provided.

Fixes #593

#### Release Note
Dropped support for YAML content type


Hayden-IO:
Looks like there is a difference between how the comments are generated. This should be ok.

danishhansari:
hey 👋 when i open a PR there was a dependabot that making unnecessary commit automatically i have mo idea  about that please help me i am stuck

bobcallaway:
> Looks like there is a difference between how the comments are generated. This should be ok.

I think there's one additional thing in the Makefile that needs addressed: `--default-consumes application/json\;q=1` should be removed

bobcallaway:
> hey wave when i open a PR there was a dependabot that making unnecessary commit automatically i have mo idea about that please help me i am stuck

@ansari-danish please open a separate issue instead of discussing on an unrelated PR. Thanks!

Hayden-IO:
@bobcallaway Updated the Makefile. Any advice on how to handle the client differences due to the comment? Can we just ignore the required check when merging?

Hayden-IO:
@bobcallaway Ran `make clean; make`, and got the same result.

Hayden-IO:
Figured it out, I was using go 1.19 and not .18.

bobcallaway (approved):

PR #945, opened 2022-07-28
Add rekor harness tests 

priyawadhwa:
This PR adds in tests specifically for the test harness. Instead of running tests we already have, the harness will run specific tests:

1. `TestHarnessAddEntry` -- adds in the standard HashedRekord type
2.  `TestHarnessAddIntoto` -- adds in an intoto attestation
3.  `TestHarnessGetAllEntriesLogIndex` -- makes sure the current combo of rekor-server/rekor-cli can get every entry in the log by log index. If the entry is an attestation, confirms that the "Attestation" field is not empty.

We still test against the 3 most recent versions of the server, against the 3 most recent versions of the CLI. We  start the server, run the tests, and then upgrade. At the end, we make sure there are 18 entries in the log as expected.

TODO: We want to make sure that getting entries by UUID/EntryID will also work. Will add that in an upcoming PR.

cc @bobcallaway

### DIFF
--- a/tests/harness_test.go
+++ b/tests/harness_test.go
@@ -1,0 +1,211 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/base64"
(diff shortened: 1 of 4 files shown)


cpanato (approved):

dlorenc (approved):

tests/harness_test.go
```diff
@@ -0,0 +1,211 @@
...
+		if !strings.Contains(out, "IntotoObj") {
+			continue
+		}
+		var intotoObj struct {
+			Attestation string
+		}
+		if err := json.Unmarshal([]byte(out), &intotoObj); err != nil {
+			t.Fatal(err)
+		}
+		if intotoObj.Attestation == "" {
```
bobcallaway, line 193: how difficult would it be to not just test if they're not empty, but ensure the same value was returned?

bobcallaway (commented):
lgtm, one question regarding the logic

tests/harness_test.go
priyawadhwa: shouldn't be too hard, i think we could try storing expected values in a file so they persist across test runs. cool if we merge this & i add it in a follow-up?

bobcallaway (approved):

tests/harness_test.go
bobcallaway: opened #951 to track

priyawadhwa:
thanks for the reviews! i don't think i have write access, could someone merge this for me? 😄

PR #956, opened 2022-08-05
api: fix inclusion proof verification flake

asraa:

Fixes:
* https://github.com/sigstore/cosign/issues/2123
* https://github.com/slsa-framework/slsa-github-generator/issues/658
* https://github.com/slsa-framework/slsa-github-generator/issues/601

#### Summary
This fixes a flakey problem with Rekor entry verification.

Every once in a while, we receive errors in validating the inclusion proof from Rekor:

it may either be an error verifying the inclusion proof:
```
 validating log entry 37: verifying inclusion proof: calculated root:
[73 180 215 55 7 58 247 212 29 155 130 228 142 72 218 20 131 130 3 179 61 150 19 180 243 114 50 157 177 182 130 69]
 does not match expected root:
[237 220 193 22 243 233 200 112 169 46 110 170 96 240 121 161 104 148 224 127 132 219 64 234 88 87 57 75 232 144 114 197]
```

or may be caught earlier than that with incorrect inclusion proof size:
```
2022/08/04 15:00:38 validating log entry 25: verifying inclusion proof: wrong proof size 3, want 4
```

This happens because we
1.  get the current tree size X
2.  ANOTHER client uploads an entry increasing the tree size to Y and then
3.  trillian returns an inclusion proof for the index for tree size X.

See code [here](https://github.com/sigstore/rekor/blob/45fd37dca40a910910c3292f1f3384b470c7c937/pkg/api/trillian_client.go#L213-L235)

The trillian response for the inclusion proof includes a `SignedLogRoot` for the tree size at size Y.
https://github.com/google/trillian/blob/44841b0bad99d6b7ed5ab20ff24cfa5ca6add9d3/trillian_log_api.proto#L231-L235

The Rekor server validates the inclusion proof at size X, returning the proof response successfully.
https://github.com/sigstore/rekor/blob/547eb3cef9690bc97e73e35eb3bf1f40b18a3504/pkg/api/trillian_client.go#L229

Rekor returns the proof response with the `SignedLogRoot` at tree size Y, client attempts to validate with this. This errors out.
https://github.com/sigstore/rekor/blob/45fd37dca40a910910c3292f1f3384b470c7c937/pkg/api/entries.go#L445

This PR returns the SignedLogRoot for the tree size in the requested proof.

On a side note, I dislike very much that Trillian's proof response does not contain the requested tree size X.

#### Release Note

* bug fix: fix inclusion proof response in Rekor server that causes flakey verification problems

### DIFF
--- a/pkg/api/trillian_client.go
+++ b/pkg/api/trillian_client.go
@@ -69,6 +69,14 @@ type Response struct {
 	getConsistencyProofResult *trillian.GetConsistencyProofResponse
 }
 
+func unmarshalLogRoot(logRoot []byte) (types.LogRootV1, error) {
+	var root types.LogRootV1
+	if err := root.UnmarshalBinary(logRoot); err != nil {
+		return types.LogRootV1{}, err
+	}
+	return root, nil
+}
+
 func (t *TrillianClient) root() (types.LogRootV1, error) {
 	rqst := &trillian.GetLatestSignedLogRootRequest{
 		LogId: t.logID,
@@ -77,11 +85,7 @@ func (t *TrillianClient) root() (types.LogRootV1, error) {
 	if err != nil {
 		return types.LogRootV1{}, err
 	}
-	var root types.LogRootV1
-	if err := root.UnmarshalBinary(resp.SignedLogRoot.LogRoot); err != nil {
-		return types.LogRootV1{}, err
-	}
-	return root, nil
+	return unmarshalLogRoot(resp.SignedLogRoot.LogRoot)
 }
 
 func (t *TrillianClient) addLeaf(byteValue []byte) *Response {
@@ -210,8 +214,15 @@ func (t *TrillianClient) getLeafAndProofByIndex(index int64) *Response {
 	ctx, cancel := context.WithTimeout(t.context, 20*time.Second)
 	defer cancel()
 
-	root, err := t.root()
-	if err != nil {
+	rootResp := t.getLatest(0)
+	if rootResp.err != nil {
+		return &Response{
+			status: status.Code(rootResp.err),
+			err:    rootResp.err,
+		}
+	}
+	var root types.LogRootV1
+	if err := root.UnmarshalBinary(rootResp.getLatestResult.SignedLogRoot.LogRoot); err != nil {
 		return &Response{
 			status: status.Code(err),
 			err:    err,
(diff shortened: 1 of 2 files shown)


dlorenc:
Amazing find!

Hayden-IO:
This is really fantastic, great work!

Can you add a comment briefly summarizing this design decision in the client so we have context for why it's designed this way?

pkg/api/trillian_client.go
```diff
@@ -210,8 +214,15 @@ func (t *TrillianClient) getLeafAndProofByIndex(index int64) *Response {
...
+		return &Response{
+			status: status.Code(rootResp.err),
+			err:    rootResp.err,
+		}
+	}
+	var root types.LogRootV1
+	if err := root.UnmarshalBinary(rootResp.getLatestResult.SignedLogRoot.LogRoot); err != nil {
 		return &Response{
 			status: status.Code(err),
 			err:    err,
```
bobcallaway:
```suggestion
	root, err := unmarshalLogRoot(rootResp.getLatestResult.SignedLogRoot.LogRoot)
	if err != nil {
		return &Response{
			status: status.Code(err),
			err:    err,
```

pkg/api/trillian_client.go
```diff
@@ -232,21 +243,36 @@ func (t *TrillianClient) getLeafAndProofByIndex(index int64) *Response {
...
-	if err != nil {
+	rootResp := t.getLatest(0)
+	if rootResp.err != nil {
+		return &Response{
+			status: status.Code(rootResp.err),
+			err:    rootResp.err,
+		}
+	}
+	var root types.LogRootV1
+	if err := root.UnmarshalBinary(rootResp.getLatestResult.SignedLogRoot.LogRoot); err != nil {
```
bobcallaway:
```suggestion
	root, err := unmarshalLogRoot(rootResp.getLatestResult.SignedLogRoot.LogRoot)
	if err != nil {
```

bobcallaway (commented):
great catch and RCA - just some recommendations to reuse your helper function here.

any thoughts on how we could add regression testing for this?

asraa:
> any thoughts on how we could add regression testing for this?

Adding a regression test that fires two concurrent goroutine uploading and verifying. I'm able to repro the failure at HEAD, and fix here. I had to use 50 uploads/verifies, although I suspect the number can be less.

dlorenc (approved):

bobcallaway (approved):
great work @asraa!

PR #957, opened 2022-08-05
Add harness test for getting all entries by UUID and EntryID

priyawadhwa:
This should be the last PR for harness testing.

Closes https://github.com/sigstore/public-good-instance/issues/417

### DIFF
--- a/tests/harness_test.go
+++ b/tests/harness_test.go
@@ -31,6 +31,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


dlorenc (dismissed):

tests/harness_test.go
```diff
@@ -197,11 +206,11 @@ func getAttestations(t *testing.T) (string, map[int]string) {
 	return file, attestations
 }
 
-func saveAttestation(t *testing.T, attestation string) {
-	file, attestations := getAttestations(t)
+func saveEntry(t *testing.T, entry StoredEntry) {
+	file, attestations := getEntries(t)
 	logIndex := activeTreeSize(t) - 1
```
bobcallaway: is this a safe assumption? we could update `getUUIDFromUploadOutput` to extract the actual index and set it in `StoredEntry` and not have to assume here.

bobcallaway (commented):
one question otherwise LGTM

tests/harness_test.go
priyawadhwa: sure, done!

dlorenc (approved):

PR #966, opened 2022-08-11
Add prometheus summary to track metric latency

priyawadhwa:
The summary will allow us to use more aggregations across data, including max().
This will be useful for debugging noisy latency alerts and getting more info about latency.

ref https://github.com/sigstore/public-good-instance/issues/533


bobcallaway (approved):
lgtm

PR #967, opened 2022-08-11
compute payload and envelope hashes upon validating intoto proposed entries

### LINKED ISSUE #854: IndexKeys is run on pre-canonicalized entries

bobcallaway:
This ensures that the server-side values used by `IndexKeys()` are computed during validation time rather than deferring it to when the indices are requested. This also ensures that client-provided values for these digests are not used when computing the list of indices. 

Fixes #854

### DIFF
--- a/pkg/types/intoto/v0.0.1/entry.go
+++ b/pkg/types/intoto/v0.0.1/entry.go
@@ -194,14 +194,12 @@ func (v *V001Entry) Canonicalize(ctx context.Context) ([]byte, error) {
 	}
 	pkb := strfmt.Base64(pk)
 
-	h := sha256.Sum256([]byte(v.IntotoObj.Content.Envelope))
-
 	canonicalEntry := models.IntotoV001Schema{
 		PublicKey: &pkb,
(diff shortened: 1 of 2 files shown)


dlorenc (approved):

asraa (approved):
Beautiful clean-up too! Thank you so, so much!

Issue #849, opened 2022-05-31
`rekor-cli`: intersection vs union search


di:
**Description**

Currently if I provide multiple search flags to `rekor-cli search`, I get the union of results with those individual terms (without duplicates removed):

```
$ rekor-cli search --sha `sha256sum hi.txt`
Found matching entries (listed by UUID):
5dbf6d1b87e46f0bdf218c42befffd9a77dc678fc6868ee63f97db6145db4625

$ rekor-cli search --email di@python.org
Found matching entries (listed by UUID):
5dbf6d1b87e46f0bdf218c42befffd9a77dc678fc6868ee63f97db6145db4625
00f1e767aea63eeec6f01005c7a9a6b0eb578a77ebc6a0dc3f4798a172d7562b
3c2ebb9e75bf89efeb52e2a465ad57c0f93f4d86c2aca63f42a1d2f34a835226
89473f4630b65edef4927dc5f02f8e810325455087297d8d1dec8998127f2420
19700e194c143e7aedb379004330f7636fa76afff4f59fcea2ed762412299ab7

$ rekor-cli search --sha `sha256sum hi.txt` --email di@python.org
Found matching entries (listed by UUID):
5dbf6d1b87e46f0bdf218c42befffd9a77dc678fc6868ee63f97db6145db4625
5dbf6d1b87e46f0bdf218c42befffd9a77dc678fc6868ee63f97db6145db4625
00f1e767aea63eeec6f01005c7a9a6b0eb578a77ebc6a0dc3f4798a172d7562b
3c2ebb9e75bf89efeb52e2a465ad57c0f93f4d86c2aca63f42a1d2f34a835226
89473f4630b65edef4927dc5f02f8e810325455087297d8d1dec8998127f2420
19700e194c143e7aedb379004330f7636fa76afff4f59fcea2ed762412299ab7
```

It'd be helpful to be able to provide a flag which switches this from the implicit `OR` to an explicit `AND` operator, something like:

```
$ rekor-cli search --operator AND --sha `sha256sum hi.txt` --email di@python.org
Found matching entries (listed by UUID):
5dbf6d1b87e46f0bdf218c42befffd9a77dc678fc6868ee63f97db6145db4625
```

As a workaround for now, folks can do something like:

```
$ comm -12 <(rekor-cli search --sha `sha256sum hi.txt` | sort) <(rekor-cli search --email di@python.org | sort)
Found matching entries (listed by UUID):
Found matching entries (listed by UUID):
[...]

PR #968, opened 2022-08-12
add support for `intersection` & `union` in search operations

dsa0x:
#### Summary

This PR addresses this enhancement issue https://github.com/sigstore/rekor/issues/849, which suggested adding support for intersection and union search. The PR adds support for `and` and `or` operators.
- `and` operation returns a union of values in each list (i.e no duplicates)
- `or` returns a list of all values, not minding duplicates.
#### Release Note

new:   support for intersection and union search through the use of `and` and `or` operators.


dlorenc:
Oh wow, this is really cool!

cmd/rekor-cli/app/search.go
```diff
@@ -58,6 +58,8 @@ func addSearchPFlags(cmd *cobra.Command) error {
 	cmd.Flags().Var(NewFlagValue(shaFlag, ""), "sha", "the SHA256 or SHA1 sum of the artifact")
 
 	cmd.Flags().Var(NewFlagValue(emailFlag, ""), "email", "email associated with the public key's subject")
+
+	cmd.Flags().Var(NewFlagValue(operatorFlag, ""), "operator", "operator to use for the search")
```
asraa: nit: add supported value are `and` and `or`

pkg/api/index.go
```diff
@@ -100,3 +107,32 @@ func addToIndex(ctx context.Context, key, value string) error {
...
 }
+
+type Container interface {
+	Add([]string)
+	Result() []string
+}
+
+type Uniq map[string]struct{}
+
+func (u Uniq) Add(elements []string) {
```
asraa: Nice implementation!

Question: If this is for `And` then won't we want an intersection? This seems to add any one of them, with a unique result.

For example, I expect the last query to be the intersection of the first two results, and only have 1 UUID
```
asraa@asraa1:~/git/rekor$ ./rekor search --public-key publickey.pem  --pki-format x509 --rekor_server http://localhost:3000
Found matching entries (listed by UUID):
64b1e670a1cf84b4d636d2d578fea277126279529759f537879796d8480f284f05708ba02b10ca38
64b1e670a1cf84b4af3fda3822b3db033f55be3af22324fa16381dc432d60d1ef1bd7ea7036538b1
asraa@asraa1:~/git/rekor$ ./rekor search --artifact COPYRIGHT.txt --rekor_server http://localhost:3000
Found matching entries (listed by UUID):
64b1e670a1cf84b47d1d6bedeb657dde5830a09aea72c528358786ee37183d7fb547f05692bb8af0
64b1e670a1cf84b4d636d2d578fea277126279529759f537879796d8480f284f05708ba02b10ca38
asraa@asraa1:~/git/rekor$ ./rekor search --public-key publickey.pem --operator and --artifact COPYRIGHT.txt  --pki-format x509 --rekor_server http://localhost:3000
Found matching entries (listed by UUID):
64b1e670a1cf84b47d1d6bedeb657dde5830a09aea72c528358786ee37183d7fb547f05692bb8af0
64b1e670a1cf84b4d636d2d578fea277126279529759f537879796d8480f284f05708ba02b10ca38
64b1e670a1cf84b4af3fda3822b3db033f55be3af22324fa16381dc432d60d1ef1bd7ea7036538b1
```

asraa (commented):
Thanks! Quick question on correctness/expectation here.

pkg/api/index.go
dsa0x: I think intersection/union are not the correct words to use here. It's not a set. Basically, what we are returning is either the union of both values (i.e no duplicates), or just return all values, not minding the duplicates.

I am not sure if there is usecase for intersection, or we could have an operator like `intersect` for intersection. what do you think?

asraa: What is the use-case of not minding duplicates? For anyone iterating through results, that would be inefficient.

Yes! I thought intersection is the most useful: e.g. "I want signatures on package X signed by user Y."

dsa0x: Yeah. That makes sense. I'll go about it this way. if there's no operator, it'll work like it does now, and return all values. Then we can use the operator for intersection or union.

cmd/rekor-cli/app/search.go
dsa0x: addressed in 12b38af93af0481aaf44015a3cd4462bfc48d486

asraa (approved):
This looks amazing now! Great work. Good job on the unit tests and testing in e2e. 
The implementation was very cool to read.

dlorenc (approved):

PR #969, opened 2022-08-12
update field documentation on publicKey for hashedrekord

### LINKED ISSUE #809: Improve clarity that the publicKey field accepts an X509 certificate
**Description**

Several folks have been caught up by the fact that the `publicKey` field accepts an X509 cert.

https://github.com/sigstore/sigstore-python/pull/66

https://github.com/wlynch/smimecosign/commit/501b64357f9bbc9c2fc8f578501867cff0d9f026

> Had the same confusion writing sget.

\- @imjasonh 

We should consider creating a new field or renaming it.

I think this is specific just to HashedRekord and derivatives?

bobcallaway:
Fixes: #809 

### DIFF
--- a/pkg/types/hashedrekord/v0.0.1/hashedrekord_v0_0_1_schema.json
+++ b/pkg/types/hashedrekord/v0.0.1/hashedrekord_v0_0_1_schema.json
@@ -15,11 +15,11 @@
                     "format": "byte"
                 },
                 "publicKey" : {
-                    "description": "The public key that can verify the signature",
+                    "description": "The public key that can verify the signature; this can also be an X509 code signing certificate that contains the raw public key information",
                     "type": "object",
                     "properties": {
                         "content": {
-                            "description": "Specifies the content of the public key inline within the document",
+                            "description": "Specifies the content of the public key or code signing certificate inline within the document",
                             "type": "string",
                             "format": "byte"
                         }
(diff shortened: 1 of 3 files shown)


pkg/types/hashedrekord/v0.0.1/hashedrekord_v0_0_1_schema.json
```diff
@@ -15,11 +15,11 @@
                     "format": "byte"
                 },
                 "publicKey" : {
-                    "description": "The public key that can verify the signature",
+                    "description": "The public key that can verify the signature; this can also be an X509 code signing certificate that contains the raw public key information",
```
asraa, line 18: nit: specify encoding?
base64 encoding of the PEM encoded public key or X.509 cert right?

asraa (approved):
otherwise lgtm, thanks!

pkg/types/hashedrekord/v0.0.1/hashedrekord_v0_0_1_schema.json
bobcallaway: the `format: byte` below says that they should be base64 encoded. I think PEM or DER will work for the raw key, and probably the same for the cert.

asraa: Ah you're right, DER works as well, LGTM then

PR #973, opened 2022-08-15
Intoto v0.0.2

### LINKED ISSUE #582: in-toto records don't contain signatures

pxp928:
#### Summary

The current implementation of the intoto type within Rekor does not persist the signatures from the wrapping DSSE envelope into the log entry stored by Trillian. This makes it impossible to independently verify the cryptographic validity of the entry without possession of the original DSSE envelope.

Based on the discussion and design doc, it was decided to create a v0.0.2 of the intoto Rekor type with changes from v0.0.1. In addition also persisting multiple public keys and signatures. 

Design Doc - https://docs.google.com/document/d/17gB598uEkoxx8j9sDhrvfhuNFHW5BSsWEiMP8POn8xo/edit?resourcekey=0-1H4eG4-4-UQYIEXZgj6AKQ#heading=h.6dq5va2kfzsw

Fixes https://github.com/sigstore/rekor/issues/582

#### Release Note

Adds new v0.0.2 for intoto type into rekor for support of DSSE envelope with multi signature and public key


pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,417 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
bobcallaway: s/2021/2022 for new files

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,417 @@
...
+	if err := json.Unmarshal(p, &predicate); err != nil {
+		return nil, err
+	}
+	return &predicate, nil
+}
+
+func (v *V002Entry) Unmarshal(pe models.ProposedEntry) error {
+	it, ok := pe.(*models.Intoto)
+	if !ok {
+		return errors.New("cannot unmarshal non Intoto v0.0.1 type")
```
bobcallaway:
```suggestion
		return errors.New("cannot unmarshal non Intoto v0.0.2 type")
```

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	// field validation
+	if err := v.IntotoObj.Validate(strfmt.Default); err != nil {
+		return err
+	}
+
+	if string(v.IntotoObj.Content.Envelope.Payload) == "" {
+		return nil
+	}
```
bobcallaway, line 174: why do you return nil here? If the payload were missing this feels like an error.

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,417 @@
...
+	returnVal.Spec = re.IntotoObj
+	returnVal.APIVersion = swag.String(re.APIVersion())
+
+	return &returnVal, nil
+}
+
+// verifyEnvelope takes in an array of possible key bytes and attempts to parse them as x509 public keys.
+// it then uses these to verify the envelope and makes sure that every signature on the envelope is verified.
+// it returns a map of verifiers indexed by the signature the verifier corresponds to.
+func verifyEnvelope(allPubKeyBytes [][]byte, env *dsse.Envelope) (map[string]*x509.PublicKey, error) {
```
bobcallaway, line 380: Is there a way we can use the DSSE signer/verifier implementation from https://pkg.go.dev/github.com/sigstore/sigstore@v1.4.0/pkg/signature/dsse instead of re-implementing it here?

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
```diff
@@ -0,0 +1,101 @@
...
+                        }
+                    },
+                    "required": [
+                        "algorithm",
+                        "value"
+                    ],
+                    "readOnly": true
+                },
+                "payloadHash": {
+                    "description": "Specifies the hash algorithm and value covering the envelope's payload after being PAE encoded",
```
bobcallaway: I'm curious as to why you decided to take the hash over the PAE encoded payload vs the decoded payload?

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	return nil
+}
+
+func (v *V002Entry) Canonicalize(ctx context.Context) ([]byte, error) {
+
+	canonicalEntry := models.IntotoV002Schema{
+		Content: &models.IntotoV002SchemaContent{
+			Envelope:    v.IntotoObj.Content.Envelope,
```
bobcallaway: This will persist the public key(s), signature(s), as well as the payload itself (which we don't do in v0.0.1). Is there a reason we are storing the payload here too?

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	v.env = *env
+
+	decodedPayload, err := base64.StdEncoding.DecodeString(string(v.IntotoObj.Content.Envelope.Payload))
+	if err != nil {
+		return fmt.Errorf("could not decode envelope payload: %w", err)
+	}
+
+	paeEncodedPayload := dsse.PAE(*v.IntotoObj.Content.Envelope.PayloadType, decodedPayload)
```
bobcallaway: the payloadHash computed here won't match the []byte returned on line 241 since it is not paeEncoded; these need to match.

pkg/types/entries.go
```diff
@@ -134,5 +134,7 @@ type ArtifactProperties struct {
 	SignatureBytes              []byte
 	PublicKeyPath               *url.URL
 	PublicKeyBytes              []byte
+	MultiPublicKeyBytes         [][]byte
```
bobcallaway: I think you'll need to find a way to pass multiple public keys in via command line args to fill this slice, which I don't see in this PR.

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,417 @@
...
+	for _, sig := range v.IntotoObj.Content.Envelope.Signatures {
+		keyHash := sha256.Sum256(sig.PublicKey)
+		result = append(result, "sha256:"+hex.EncodeToString(keyHash[:]))
+
+		keyObj, err := x509.NewPublicKey(bytes.NewReader(sig.PublicKey))
+		if err != nil {
+			return nil, err
+		}
+
+		result = append(result, keyObj.EmailAddresses()...)
```
bobcallaway:
```suggestion
		result = append(result, keyObj.Subjects()...)
```

pkg/types/intoto/v0.0.2/entry.go
pxp928: Not sure what you mean by this question. Can you elaborate? The DSSE envelope contains the payload. Is it also getting stored somewhere in rekor that I need to account for?

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
pxp928: Since DSSE uses PAE based on the protocol definition it makes sense to store the hash PAE format, which includes the payloadType.

pkg/types/intoto/v0.0.2/entry.go
pxp928: Fixed.

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
```diff
@@ -0,0 +1,101 @@
...
+                                "description": "a signature of the envelope's payload along with the public key for the signature",
+                                "type": "object",
+                                "properties": {
+                                    "keyid": {
+                                        "description": "optional id of the key used to create the signature",
+                                        "type": "string"
+                                    },
+                                    "sig": {
+                                        "description": "signature of the payload",
+                                        "type": "string"
```
bobcallaway: this should probably have `"format": "byte"` added since it will be base64 encoded.

pkg/types/intoto/v0.0.2/entry.go
bobcallaway: the JSON schema for `v.IntotoObj.Content.Envelope` shows that it contains the payload, payload type, and signatures (that include the signature values & public keys). By copying through the entire value here during canonicalization, the entire value will be persisted as part of the log entry - whereas in v0.0.1 we only persisted the public key, envelopeHash and payloadHash values. 

If rekor is configured to store attestations, the payload will be written to the storage layer (`pkg/storage`).

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
bobcallaway: having `payloadHash` match the digest of the content (agnostic of any encoding scheme) would help users know that the actual intoto attestation bytes match a local copy of the attestation.

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
pxp928: Changed.

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
pxp928: Makes sense. I can remove the PAE encoding.

pkg/types/intoto/v0.0.2/entry.go
pxp928: Ah yes. I see what you mean. I will remove the payload from being stored.

pkg/types/entries.go
pxp928: Before moving further down this route, wanted to ask if this method was proper. I can add in the command line args if there was no other better alternative to this.

pkg/types/intoto/v0.0.2/entry.go
pxp928: Let me look to use this instead.

pkg/types/intoto/v0.0.2/entry.go
pxp928: Changed to payload only without the PAE encoded.

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
pxp928: Actually this is already coming base64 encoded from the DSSE envelope so this will result in double encoding. Same for payload and signature
```
{
  "payload": "<Base64(SERIALIZED_BODY)>",
  "payloadType": "<PAYLOAD_TYPE>",
  "signatures": [{
    "keyid": "<KEYID>",
    "sig": "<Base64(SIGNATURE)>"
  }]
}
```

pkg/types/intoto/v0.0.2/entry.go
pxp928: Using the implementation provided by sigstore does not provide the functionality in mapping the keys to the valid signature. The sigstore implementation just validates all the signatures with the pub keys but we further do mapping of the keys to sigs to create the log entry. We could create a issue to track this and update sigstore to add this functionality there?

pkg/types/entries.go
pxp928: added the command line args and unit tests.

pxp928:
@bobcallaway Ready for another review.

cmd/rekor-server/app/serve.go
```diff
@@ -38,7 +38,7 @@ import (
 	"github.com/sigstore/rekor/pkg/types/helm"
 	helm_v001 "github.com/sigstore/rekor/pkg/types/helm/v0.0.1"
 	"github.com/sigstore/rekor/pkg/types/intoto"
-	intoto_v001 "github.com/sigstore/rekor/pkg/types/intoto/v0.0.1"
+	intoto_v002 "github.com/sigstore/rekor/pkg/types/intoto/v0.0.2"
```
bobcallaway, line 42: I think you'll want to leave v001 to ensure that entries that are fetched from the log could still be parsed

cmd/rekor-server/app/serve.go
```diff
@@ -88,7 +88,7 @@ var serveCmd = &cobra.Command{
 			rekord.KIND:       rekord_v001.APIVERSION,
 			rpm.KIND:          rpm_v001.APIVERSION,
 			jar.KIND:          jar_v001.APIVERSION,
-			intoto.KIND:       intoto_v001.APIVERSION,
+			intoto.KIND:       intoto_v002.APIVERSION,
```
bobcallaway: same here, probably want to leave both

pkg/types/intoto/intoto_schema.json
```diff
@@ -6,7 +6,7 @@
     "type": "object",
     "oneOf": [
         {
-            "$ref": "v0.0.1/intoto_v0_0_1_schema.json"
+            "$ref": "v0.0.2/intoto_v0_0_2_schema.json"
```
bobcallaway, line 12: I think this makes sense, we need to make sure entries that were made under v001 can still be read out successfully but all new entries should ideally be stored using v002. I think we should add an e2e test that guarantees that no additional v001 entries could be made (because of the lack of signatures), but existing v001 entries could still be read out of the log (see issue872 tests which might be helpful in this).

dlorenc:
Looks like something is up with codegen here....

bobcallaway:
looks like some of the path names in `Makefile.swagger` have an extra `/` in them for some reason.

pxp928:
Looks like the Makefile is doing unwanted substitution?
```
@echo "SWAGGER_GEN=`find pkg/generated/client pkg/generated/models/ pkg/generated/restapi/ -iname '*.go' | grep -v 'configure_rekor_server' | sort -d | tr '\n' ' ' | sed 's/ $$//'`" >> Makefile.swagger;
```

bobcallaway:
> Looks like the Makefile is doing unwanted substitution?
> 
> ```
> @echo "SWAGGER_GEN=`find pkg/generated/client pkg/generated/models/ pkg/generated/restapi/ -iname '*.go' | grep -v 'configure_rekor_server' | sort -d | tr '\n' ' ' | sed 's/ $$//'`" >> Makefile.swagger;
> ```

https://github.com/sigstore/rekor/pull/984 pushed to fix this

cmd/rekor-cli/app/pflag_groups.go
```diff
@@ -73,6 +73,11 @@ func addArtifactPFlags(cmd *cobra.Command) error {
 			"path or URL to public key file",
 			false,
 		},
+		"multi-public-key": {
```
bobcallaway: What are your thoughts on just changing `public-key` to take multiple values instead of creating an entirely new cmd line argument?

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
bobcallaway: I agree they'll already come base64 encoded, but the code should account for that. The schema should document the expectation that the values are base64 encoded.

cmd/rekor-cli/app/pflag_groups.go
pxp928: Yes, I agree with this path. I actually started to go down this route but wanted the maintainers feedback before I changed an existing flag. Changing `public-key` would be the better path, I did not want to break something unintentionally because of it.

dlorenc:
Tests seem stuck here.

cmd/rekor-server/app/serve.go
pxp928: This would result in changing the intoto.KIND to handle both versions. Thoughts on what the key should look like?

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
pxp928: Would this not result double encoding? As a result we have to decode twice to get for example the payload and signature back?
```
re.IntotoObj.Content.Envelope.Payload = strfmt.Base64(env.Payload)
```

dlorenc:
Github is back, you might need to force push/rebase to kick the tests though.

cmd/rekor-server/app/serve.go
pxp928: This would result in changing the `intoto.KIND` to handle both versions. Thoughts on what the key should look like?

cmd/rekor-cli/app/pflag_groups.go
pxp928: Changed.

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
pxp928: Done.

pkg/types/intoto/intoto_schema.json
pxp928: Need some help on writing up the test for this. I see the script `issue-872-e2e-test.sh` to test this but not sure what needs to be done to get a e2e test for it.

pxp928:
Updating the tests.

pxp928:
Any thoughts on why the e2e test fail on `[GET /api/v1/log/entries/{entryUUID}][404] getLogEntryByUuidNotFound`. The entry seems to get created successfully.

cmd/rekor-server/app/serve.go
priyawadhwa: Maybe we just have keys with the version included, something like

```
intoto.KINDV001
intoto.KINDV002
```

etc

pkg/types/intoto/intoto_schema.json
priyawadhwa: I think the existing harness tests might cover this? Older versions will insert intoto attestations at v001, and later when we test while running @ HEAD we can ensure that they can still be retrieved.

We would just need to update the harness tests to test against the 3 previous versions AND main (we don't test against main at the moment).

priyawadhwa:
> Any thoughts on why the e2e test fail on [GET /api/v1/log/entries/{entryUUID}][404] getLogEntryByUuidNotFound. The entry seems to get created successfully.

I ran it locally and got this error:

```
    e2e_test.go:527: [POST /api/v1/log/entries][400] createLogEntryBadRequest  &{Code:400 Message:Error processing entry: intoto implementation for version '0.0.1' not found: unable to locate entry for version 0.0.1}
```


probably related to this comment https://github.com/sigstore/rekor/pull/973/files#r948283591

cmd/rekor-server/app/serve.go
pxp928: @priyawadhwa How did we want to handle the changes to the keys downstream. To initialize both versions of in-toto
```
func init() {
	types.TypeMap.Store(KIND, New)
}

func New() types.TypeImpl {
	bit := BaseIntotoType{}
	bit.Kind = KIND
	bit.VersionMap = VersionMap
	return &bit
}
```

pxp928: Is there a good way to handle the old and new version of intoto entries?

bobcallaway: just make this a `pluggableTypeMap := map[string][]string` and iterate through all the values for each key rather than a 1:1 mapping

pxp928: I was thinking it would be more complicated then that lol.

pxp928:
> > Any thoughts on why the e2e test fail on [GET /api/v1/log/entries/{entryUUID}][404] getLogEntryByUuidNotFound. The entry seems to get created successfully.
> 
> I ran it locally and got this error:
> 
> ```
>     e2e_test.go:527: [POST /api/v1/log/entries][400] createLogEntryBadRequest  &{Code:400 Message:Error processing entry: intoto implementation for version '0.0.1' not found: unable to locate entry for version 0.0.1}
> ```
> 
> probably related to this comment https://github.com/sigstore/rekor/pull/973/files#r948283591

Still getting the same error on the e2e test: `[GET /api/v1/log/entries/{entryUUID}][404] getLogEntryByUuidNotFound`. Did some testing and it looks like the entry does get created. When I run it twice, it does say an entry is present already.

cmd/rekor-cli/app/search.go
```diff
@@ -164,15 +164,18 @@ var searchCmd = &cobra.Command{
...
-			publicKeyStr := viper.GetString("public-key")
-			if isURL(publicKeyStr) {
-				params.Query.PublicKey.URL = strfmt.URI(publicKeyStr)
-			} else {
-				keyBytes, err := ioutil.ReadFile(filepath.Clean(publicKeyStr))
-				if err != nil {
-					return nil, fmt.Errorf("error reading public key file: %w", err)
+
+			splitPubKeyString := strings.Split(publicKeyStr, ",")
+			if len(splitPubKeyString) == 1 {
```
asraa, line 169: should there be an else clause warning that only one public key is provided? or else if someone specifies multiple, they won't search for any at all currently.

pkg/types/alpine/v0.0.1/entry.go
```diff
@@ -322,14 +322,21 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
 
 	re.AlpineModel.PublicKey = &models.AlpineV001SchemaPublicKey{}
 	publicKeyBytes := props.PublicKeyBytes
-	if publicKeyBytes == nil {
-		publicKeyBytes, err = ioutil.ReadFile(filepath.Clean(props.PublicKeyPath.Path))
-		if err != nil {
-			return nil, fmt.Errorf("error reading public key file: %w", err)
+	if len(publicKeyBytes) == 0 {
+		if len(props.PublicKeyPath) > 1 {
```
asraa: nit: for simplicity just use `if len(props.PublicKeyPath) != 1`? then no need for the second if clause

pkg/types/cose/v0.0.1/entry.go
```diff
@@ -316,19 +316,25 @@ func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.A
...
-	if publicKeyBytes == nil {
-		if props.PublicKeyPath == nil {
+	if len(publicKeyBytes) == 0 {
+		if len(props.PublicKeyPath) == 0 {
 			return nil, errors.New("public key must be provided to verify signature")
 		}
-		publicKeyBytes, err = ioutil.ReadFile(filepath.Clean(props.PublicKeyPath.Path))
-		if err != nil {
-			return nil, fmt.Errorf("error reading public key file: %w", err)
+		if len(props.PublicKeyPath) > 1 {
```
asraa: again, could consolidate the `>1` and `==1` check

pkg/types/alpine/v0.0.1/entry.go
```diff
@@ -322,14 +322,21 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
...
+			if err != nil {
+				return nil, fmt.Errorf("error reading public key file: %w", err)
+			}
+			publicKeyBytes = append(publicKeyBytes, keyBytes)
+			re.AlpineModel.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes[0])
+		}
+
 	} else {
-		re.AlpineModel.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
+		re.AlpineModel.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes[0])
```
asraa: should there be a warning when more than one public key byte files are provided?

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,418 @@
...
+func (v V002Entry) IndexKeys() ([]string, error) {
+	var result []string
+
+	if v.IntotoObj.Content == nil || v.IntotoObj.Content.Envelope == nil {
+		log.Logger.Info("IntotoObj content or dsse envelope is nil")
+		return result, nil
+	}
+
+	for _, sig := range v.IntotoObj.Content.Envelope.Signatures {
+		keyHash := sha256.Sum256(sig.PublicKey)
```
asraa: maybe take the `keyObj` below's CanonicalValue for the index of the key?

pkg/types/intoto/v0.0.2/entry.go
```diff
@@ -0,0 +1,418 @@
...
+func (v *V002Entry) Canonicalize(ctx context.Context) ([]byte, error) {
+
+	canonicalEntry := models.IntotoV002Schema{
+		Content: &models.IntotoV002SchemaContent{
+			Envelope: &models.IntotoV002SchemaContentEnvelope{
+				PayloadType: v.IntotoObj.Content.Envelope.PayloadType,
+				Signatures:  v.IntotoObj.Content.Envelope.Signatures,
+			},
+			Hash:        v.IntotoObj.Content.Hash,
+			PayloadHash: v.IntotoObj.Content.PayloadHash,
```
asraa, line 220: why not include the payloadhash inside the envelope? So in the envelope scheme there is currently a payload field, and maybe it can be extended to have either a payload content or a payloadHash. That way users can support adding verifiable attestations with hidden/private payload content.

asraa (commented):
Thanks! This may be a very late drive-by on the schema: but I have a major question about the schema's payload and payloadHash

cmd/rekor-cli/app/search.go
pxp928: agreed will add.

pkg/types/alpine/v0.0.1/entry.go
pxp928: yup will add.

pkg/types/intoto/v0.0.2/entry.go
pxp928: Should this be a large discussion on if we want to handle this now or in the future? I understand the use case but has the community adopted this yet?

asraa: I think we haven't been able to adopt it yet because there wasn't support -- since it's just a minor change to schema, allowing users to upload and set the payload hash would be nice to do in this iteration.

But on the other hand that does change `verifyEnvelope` -- and the impl https://github.com/secure-systems-lab/go-securesystemslib/blob/main/dsse/verify.go here would need to change. 

Would movidying a `readOnly` attribute constitute a breaking API change? @bobcallaway 

But regardless, moving the payloadhash inside the payload field wouldn't change anything and I think it looks organizationally better to group the payload attributes

bobcallaway: why should users be able to upload the payload hash? I think we should compute it server-side only. If people only want to provide a hash, then they should use `hashedrekord` only.

asraa: >  If people only want to provide a hash, then they should use hashedrekord only.

This is fair, closing this. There is only one benefit: the multi-sig properties of intoto. But don't get any benefit from the type-specific information that intoto gives you over hashedrekord.

pkg/types/alpine/v0.0.1/entry.go
pxp928: done.

pkg/types/cose/v0.0.1/entry.go
pxp928: done

pkg/types/intoto/v0.0.2/entry.go
pxp928: done.

pxp928:
> Still getting the same error on the e2e test: [GET /api/v1/log/entries/{entryUUID}][404] getLogEntryByUuidNotFound. Did some testing and it looks like the entry does get created. When I run it twice, it does say an entry is present already.

Any thoughts on why this error is occurring? I double checked everything multiple times and nothing apparent.

bobcallaway:
> > Still getting the same error on the e2e test: [GET /api/v1/log/entries/{entryUUID}][404] getLogEntryByUuidNotFound. Did some testing and it looks like the entry does get created. When I run it twice, it does say an entry is present already.
> 
> Any thoughts on why this error is occurring? I double checked everything multiple times and nothing apparent.

Off of a quick debug, I believe it is because you have `content.Envelope.Payload` as required. It will be present upon upload but since it is not persisted into the log you have errors reading it back out. I think you have to set it as optional and deal with the conditional in `Validate()` (see v0.0.1 for the corollary)

priyawadhwa:
Edit : just saw Bob's comment, I think that makes sense!

I was able to repro this locally and keep getting this error when doing `rekor-cli get` with either logIndex or uuid:

```
=== RUN   TestIntoto
    e2e_test.go:527: [../rekor-cli upload --artifact /var/folders/fz/xhw4wqps4d165vcmnt_dhy040000gn/T/TestIntoto32180235/001/attestation.json --type intoto --public-key /var/folders/fz/xhw4wqps4d165vcmnt_dhy040000gn/T/TestIntoto32180235/001/pub.pem --rekor_server=http://localhost:3000]
        Created entry at index 0, available at: http://localhost:3000/api/v1/log/entries/66e8fc5ce39f63dbc3a7dbc9fe0188d7c050c89216235416360a834a6d5442e4
        
    e2e_test.go:531: Got uuid:  66e8fc5ce39f63dbc3a7dbc9fe0188d7c050c89216235416360a834a6d5442e4
    e2e_test.go:533: [../rekor-cli get --uuid 66e8fc5ce39f63dbc3a7dbc9fe0188d7c050c89216235416360a834a6d5442e4 --format=json --rekor_server=http://localhost:3000]
        validation failure list:
        validation failure list:
        validation failure list:
        content.content.envelope.content.envelope.payload in body is required
        
    e2e_test.go:533: exit status 1
--- FAIL: TestIntoto (0.31s)
```

Looks like `rekor-cli upload` is successful, but when `rekor-cli get` is run it fails on verification. Seems like some fields are expected but not present? The error is coming from the call to [IntotoObj.Validate](https://github.com/sigstore/rekor/pull/973/files#diff-5baa5429646882e5392613a47233048ad95214ed1c23e4caee9cd9d692f4e254R168). Maybe the attestation that is being uploaded needs to be formatted differently (in which case I'd expect `upload` to fail, not `get`), or it's a bug in validation. 

That's all I've got now, but I'll try to keep digging.

pxp928:
Ah so I tested with the `payload` not being required but because I thought the payload needed to be required based on the comment https://github.com/sigstore/rekor/pull/973#discussion_r945898495 I never got the e2e to work. Thanks for the help @priyawadhwa and @bobcallaway! I tested it locally now and the test run well now.

pkg/types/alpine/v0.0.1/entry.go
```diff
@@ -322,16 +322,21 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
 
 	re.AlpineModel.PublicKey = &models.AlpineV001SchemaPublicKey{}
 	publicKeyBytes := props.PublicKeyBytes
-	if publicKeyBytes == nil {
-		publicKeyBytes, err = ioutil.ReadFile(filepath.Clean(props.PublicKeyPath.Path))
+	if len(publicKeyBytes) == 0 {
+		if len(props.PublicKeyPath) != 1 {
```
bobcallaway: nit: if you end up pushing up another patch, could we rename `PublicKeyPath` to `PublicKeyPaths`?

pxp928: done.

pkg/types/intoto/intoto_schema.json
bobcallaway: i tested locally and you can still upload 0.0.1 intoto entries in to the log. Once you rebase ontop of #1003 and #1004, I think we should add the following code to `pkg/types/intoto/intoto.go`:
```
// SupportedVersions returns the supported versions for this type;
// it deliberately omits 0.0.1 from the list of supported versions as that
// version did not persist signatures inside the log entry
func (it BaseIntotoType) SupportedVersions() []string {
	return []string{"0.0.2"}
}
```
We may have to figure out how to adapt the unit tests accordingly as I believe the `TestIntoto` is used by the harness and issue872 tests, but will probably fail once you do the rebase.

bobcallaway: also I think we should have both of versions listed in the JSON schema (otherwise the codegen becomes disconnected from the source, even though you're technically not deleting it in the PR)

pxp928: Rebased to include the new PRs and added `SupportedVersions`. The harness tests do fail when set to `intotoModel := &models.IntotoV001Schema{}`. Currently setting them to `intotoModel := &models.IntotoV002Schema{}`.

dlorenc:
Everything is passing! Is this on track to merge this week @bobcallaway?

bobcallaway:
> Everything is passing! Is this on track to merge this week @bobcallaway?

@pxp928 and I are debugging hopefully the final issue before merge now

tests/e2e_test.go
```diff
@@ -559,6 +575,234 @@ func TestIntoto(t *testing.T) {
...
+	collectedKeys := []*url.URL{{Path: pubKeyPath}}
+	props.PublicKeyPaths = collectedKeys
+
+	entry, err = types.NewProposedEntry(context.Background(), "intoto", "0.0.1", *props)
+	if err != nil {
+		t.Fatal(err)
+	}
+	params.SetProposedEntry(entry)
+
+	_, err = rekorClient.Entries.CreateLogEntry(params)
```
bobcallaway, line 800: if `err` (incorrectly) comes back as `nil`, you'll panic here; perhaps just do a check `if err == nil { t.Fatal("insertion of v0.0.1 entry should fail") }`

pkg/types/alpine/v0.0.1/entry.go
```diff
@@ -322,16 +322,21 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
...
+		keyBytes, err := ioutil.ReadFile(filepath.Clean(props.PublicKeyPaths[0].Path))
 		if err != nil {
 			return nil, fmt.Errorf("error reading public key file: %w", err)
 		}
-		re.AlpineModel.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
-	} else {
-		re.AlpineModel.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
+		publicKeyBytes = append(publicKeyBytes, keyBytes)
+	} else if len(publicKeyBytes) != 1 {
+		return nil, errors.New("only one public key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one public key must be provided")
```

pkg/types/cose/v0.0.1/entry.go
```diff
@@ -316,19 +316,20 @@ func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.A
...
+			return nil, errors.New("only one public key must be provided to verify signature")
 		}
-		publicKeyBytes, err = ioutil.ReadFile(filepath.Clean(props.PublicKeyPath.Path))
+		keyBytes, err := ioutil.ReadFile(filepath.Clean(props.PublicKeyPaths[0].Path))
 		if err != nil {
 			return nil, fmt.Errorf("error reading public key file: %w", err)
 		}
+		publicKeyBytes = append(publicKeyBytes, keyBytes)
+	} else if len(publicKeyBytes) != 1 {
+		return nil, errors.New("only one public key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one public key must be provided")
```

pkg/types/hashedrekord/v0.0.1/entry.go
```diff
@@ -217,17 +217,20 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
...
+			return nil, errors.New("only one public key must be provided to verify detached signature")
 		}
-		publicKeyBytes, err = ioutil.ReadFile(filepath.Clean(props.PublicKeyPath.Path))
+		keyBytes, err := ioutil.ReadFile(filepath.Clean(props.PublicKeyPaths[0].Path))
 		if err != nil {
 			return nil, fmt.Errorf("error reading public key file: %w", err)
 		}
+		publicKeyBytes = append(publicKeyBytes, keyBytes)
+	} else if len(publicKeyBytes) != 1 {
+		return nil, errors.New("only one public key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one public key must be provided")
```

pkg/types/helm/v0.0.1/entry.go
```diff
@@ -317,16 +317,20 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
...
+		keyBytes, err := ioutil.ReadFile(filepath.Clean(props.PublicKeyPaths[0].Path))
 		if err != nil {
 			return nil, fmt.Errorf("error reading public key file: %w", err)
 		}
-		re.HelmObj.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
-	} else {
-		re.HelmObj.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
+		publicKeyBytes = append(publicKeyBytes, keyBytes)
+	} else if len(publicKeyBytes) != 1 {
+		return nil, errors.New("only one public key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one public key must be provided")
```

pkg/types/intoto/v0.0.1/entry.go
```diff
@@ -292,16 +292,20 @@ func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.A
...
+			return nil, errors.New("only one public key must be provided to verify signature")
 		}
-		publicKeyBytes, err = ioutil.ReadFile(filepath.Clean(props.PublicKeyPath.Path))
+		keyBytes, err := ioutil.ReadFile(filepath.Clean(props.PublicKeyPaths[0].Path))
 		if err != nil {
 			return nil, fmt.Errorf("error reading public key file: %w", err)
 		}
+		publicKeyBytes = append(publicKeyBytes, keyBytes)
+	} else if len(publicKeyBytes) != 1 {
+		return nil, errors.New("only one public key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one public key must be provided")
```

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
```diff
@@ -0,0 +1,103 @@
...
+                                        "type": "string",
+                                        "format": "byte",
+                                        "readOnly": true
+                                    }
+                                }
+                            }
+                        }
+                    },
+                    "required": ["payloadType", "signatures"],
+                    "writeOnly": true
```
bobcallaway: this shouldn't affect the code generation, but I would remove this since it's incorrect
```suggestion
```

pkg/types/rekord/v0.0.1/entry.go
```diff
@@ -391,19 +391,21 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
...
+		keyBytes, err := ioutil.ReadFile(filepath.Clean(props.PublicKeyPaths[0].Path))
 		if err != nil {
 			return nil, fmt.Errorf("error reading public key file: %w", err)
 		}
-		re.RekordObj.Signature.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
-	} else {
-		re.RekordObj.Signature.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
+		publicKeyBytes = append(publicKeyBytes, keyBytes)
+	} else if len(publicKeyBytes) != 1 {
+		return nil, errors.New("only one public key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one public key must be provided")
```

pkg/types/rpm/v0.0.1/entry.go
```diff
@@ -342,19 +342,21 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
...
+		keyBytes, err := ioutil.ReadFile(filepath.Clean(props.PublicKeyPaths[0].Path))
 		if err != nil {
 			return nil, fmt.Errorf("error reading public key file: %w", err)
 		}
-		re.RPMModel.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
-	} else {
-		re.RPMModel.PublicKey.Content = (*strfmt.Base64)(&publicKeyBytes)
+		publicKeyBytes = append(publicKeyBytes, keyBytes)
+	} else if len(publicKeyBytes) != 1 {
+		return nil, errors.New("only one public key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one public key must be provided")
```

pkg/types/tuf/v0.0.1/entry.go
```diff
@@ -333,26 +333,25 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
...
-	} else {
-		s := &data.Signed{}
-		if err := json.Unmarshal(rootBytes, s); err != nil {
-			return nil, err
-		}
-		re.TufObj.Root.Content = s
+		rootBytes = append(rootBytes, keyBytes)
+
+	} else if len(rootBytes) != 1 {
+		return nil, errors.New("only one root key byte must be provided")
```
bobcallaway:
```suggestion
		return nil, errors.New("only one root key must be provided")
```

tests/e2e_test.go
pxp928: done.

pkg/types/alpine/v0.0.1/entry.go
pxp928: changed.

pkg/types/cose/v0.0.1/entry.go
pxp928: changed.

pkg/types/hashedrekord/v0.0.1/entry.go
pxp928: changed.

pkg/types/helm/v0.0.1/entry.go
pxp928: changed.

pkg/types/intoto/v0.0.1/entry.go
pxp928: changed.

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
pxp928: changed.

pkg/types/rekord/v0.0.1/entry.go
pxp928: changed.

pkg/types/rpm/v0.0.1/entry.go
pxp928: changed.

pkg/types/tuf/v0.0.1/entry.go
pxp928: changed.

bobcallaway (approved):
LGTM, thanks for the hard work!

priyawadhwa (approved):
nice!!

dlorenc:
YESSS!

PR #974, opened 2022-08-15
Allow sharding config to be written in yaml or json

priyawadhwa:
Right now we only support yaml, which is getting annoying with Helm (Helm has a bunch of issues trying to parse it).

### DIFF
--- a/pkg/sharding/ranges.go
+++ b/pkg/sharding/ranges.go
@@ -18,6 +18,7 @@ package sharding
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


dlorenc (approved):

pkg/sharding/ranges.go
```diff
@@ -82,6 +83,10 @@ func logRangesFromPath(path string) (Ranges, error) {
 		return Ranges{}, nil
 	}
 	if err := yaml.Unmarshal(contents, &ranges); err != nil {
+		// Try to use JSON
+		if jerr := json.Unmarshal(contents, &ranges); jerr == nil {
+			return ranges, nil
```
dsa0x, line 88: If we don't assign this jerr to err, there can be a situation where we have a nil error, but a zero-value range.

PR #986, opened 2022-08-18
feat: add verification functions

### LINKED ISSUE #789: Expose Go function for validating log entry
Currently the logic for validating the signature of a log entry needs to be reconstructed at least in part by other projects that wish to rely on it, e.g.:

- https://github.com/google/ent/blob/b0804661406ee428d2127d7f2d653d6c1b45f766/cmd/ent/cmd/status.go#L110-L207
- https://github.com/slsa-framework/slsa-verifier/blob/aee753f58fa9a4de2be6290e3eefae2d6e998a9d/pkg/provenance.go#L164-L285

This logic is mostly already available in rekor, but not particularly well packaged or exposed yet.

cc @asraa @laurentsimon @eddiezane @dlorenc

### LINKED ISSUE #891: Improvements to rekor-cli verify command
**Description**

* verify should verify a signed tree head (STH) signature
* verify should verify the STH signature uses the Rekor public key from TUF, not from the public key from the instance
* verify needs to be shard aware

asraa:

#### Summary

* Adds verification functions for inclusion proof, root hash, consistency, and full entry

#### Documentation

Fixes https://github.com/sigstore/rekor/issues/891
Fixes https://github.com/sigstore/rekor/issues/789

Creating follow-up issue for the TUF usage. Contingent on https://github.com/sigstore/sigstore/pull/500/ which I'll go back to today.

### DIFF
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -77,7 +77,7 @@ func ProveConsistency(ctx context.Context, rClient *client.Rekor,
 // against a newly fetched Checkpoint.
 //nolint
 func VerifyCurrentCheckpoint(ctx context.Context, rClient *client.Rekor, verifier signature.Verifier,
-	oldSTH *util.SignedCheckpoint, treeID string) error {
+	oldSTH *util.SignedCheckpoint) error {
 	// The oldSTH should already be verified, but check for robustness.
 	if !oldSTH.Verify(verifier) {
 		return errors.New("signature on old tree head did not verify")
@@ -102,7 +102,7 @@ func VerifyCurrentCheckpoint(ctx context.Context, rClient *client.Rekor, verifie
 	}
 
 	// Now verify consistency up to the STH.
-	return ProveConsistency(ctx, rClient, oldSTH, &sth, treeID)
+	return ProveConsistency(ctx, rClient, oldSTH, &sth, *logInfo.TreeID)
 }
 
 // VerifyInclusion verifies an entry's inclusion proof. Clients MUST either verify
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -36,7 +36,7 @@ import (
 )
 
 // ProveConsistency verifies consistency between an initial, trusted STH
-// and a second new STH. Assumes that the signature on the STHs' are verified.
+// and a second new STH. Callers MUST verify signature on the STHs'.
 func ProveConsistency(ctx context.Context, rClient *client.Rekor,
 	oldSTH *util.SignedCheckpoint, newSTH *util.SignedCheckpoint, treeID string) error {
 	oldTreeSize := int64(oldSTH.Size)
@@ -78,6 +78,11 @@ func ProveConsistency(ctx context.Context, rClient *client.Rekor,
 //nolint
 func VerifyCurrentCheckpoint(ctx context.Context, rClient *client.Rekor, verifier signature.Verifier,
 	oldSTH *util.SignedCheckpoint, treeID string) error {
+	// The oldSTH should already be verified, but check for robustness.
+	if !oldSTH.Verify(verifier) {
+		return errors.New("signature on old tree head did not verify")
+	}
+
 	// Get and verify against the current STH.
 	infoParams := tlog.NewGetLogInfoParamsWithContext(ctx)
 	result, err := rClient.Tlog.GetLogInfo(infoParams)
@@ -100,8 +105,9 @@ func VerifyCurrentCheckpoint(ctx context.Context, rClient *client.Rekor, verifie
 	return ProveConsistency(ctx, rClient, oldSTH, &sth, treeID)
 }
 
-// VerifyInclusion verifies an entry's inclusion proof. Clients MUST also
-// verify the root hash via VerifyRootHash.
+// VerifyInclusion verifies an entry's inclusion proof. Clients MUST either verify
+// the root hash against a new STH (via VerifyCurrentCheckpoint) or against a
+// trusted, existing STH (via ProveConsistency).
 //nolint
 func VerifyInclusion(ctx context.Context, e *models.LogEntryAnon) error {
 	if e.Verification == nil || e.Verification.InclusionProof == nil {
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -185,7 +185,7 @@ func VerifySignedEntryTimestamp(ctx context.Context, e *models.LogEntryAnon, ver
 // Performs inclusion proof verification up to a verified root hash and
 // SignedEntryTimestamp verification.
 //nolint
-func VerifyLogEntry(ctx context.Context, rClient *client.Rekor, e *models.LogEntryAnon, verifier signature.Verifier) error {
+func VerifyLogEntry(ctx context.Context, e *models.LogEntryAnon, verifier signature.Verifier) error {
 	// Verify the inclusion proof using the body's leaf hash.
 	if err := VerifyInclusion(ctx, e); err != nil {
 		return err
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -77,32 +77,35 @@ func ProveConsistency(ctx context.Context, rClient *client.Rekor,
 // against a newly fetched Checkpoint.
 //nolint
 func VerifyCurrentCheckpoint(ctx context.Context, rClient *client.Rekor, verifier signature.Verifier,
-	oldSTH *util.SignedCheckpoint) error {
+	oldSTH *util.SignedCheckpoint) (*util.SignedCheckpoint, error) {
 	// The oldSTH should already be verified, but check for robustness.
 	if !oldSTH.Verify(verifier) {
-		return errors.New("signature on old tree head did not verify")
+		return nil, errors.New("signature on old tree head did not verify")
 	}
 
 	// Get and verify against the current STH.
 	infoParams := tlog.NewGetLogInfoParamsWithContext(ctx)
 	result, err := rClient.Tlog.GetLogInfo(infoParams)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	logInfo := result.GetPayload()
 	sth := util.SignedCheckpoint{}
 	if err := sth.UnmarshalText([]byte(*logInfo.SignedTreeHead)); err != nil {
-		return err
+		return nil, err
 	}
 
 	// Verify the signature on the SignedCheckpoint.
 	if !sth.Verify(verifier) {
-		return errors.New("signature on tree head did not verify")
+		return nil, errors.New("signature on tree head did not verify")
 	}
 
 	// Now verify consistency up to the STH.
-	return ProveConsistency(ctx, rClient, oldSTH, &sth, *logInfo.TreeID)
+	if err := ProveConsistency(ctx, rClient, oldSTH, &sth, *logInfo.TreeID); err != nil {
+		return nil, err
+	}
+	return &sth, nil
 }
 
 // VerifyInclusion verifies an entry's inclusion proof. Clients MUST either verify
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -1,0 +1,198 @@
+//
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package verify
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
+	"github.com/sigstore/rekor/pkg/generated/client"
+	"github.com/sigstore/rekor/pkg/generated/client/tlog"
+	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/sigstore/rekor/pkg/util"
+	"github.com/sigstore/sigstore/pkg/signature"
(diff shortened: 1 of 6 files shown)


pkg/verify/verify.go
```diff
@@ -0,0 +1,203 @@
...
+		return fmt.Errorf("missing verification")
+	}
+	if e.Verification.SignedEntryTimestamp == nil {
+		return fmt.Errorf("signature missing")
+	}
+
+	type bundle struct {
+		Body           interface{} `json:"body"`
+		IntegratedTime int64       `json:"integratedTime"`
+		// Note that this is the virtual index.
```
priyawadhwa, line 160: I thought we were supposed to verify against the shard log index, not the virtual log index?

asraa: Nope. Inclusion proof requires verification against the internal log index. But we sign the bundle over the virtual index:
https://github.com/sigstore/rekor/blob/6093a3391c2399d384b50bb9b6f48e4011111729/pkg/api/entries.go#L214-L258

priyawadhwa: gotcha, thanks!

asraa: i don't think it's a security issue either way though -- since the signed payload also contains the log ID, so it disambiguates between shards anyway

pkg/verify/verify.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/sigstore/rekor/pkg/util"
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"github.com/transparency-dev/merkle/proof"
+	"github.com/transparency-dev/merkle/rfc6962"
+)
+
+func ProveConsistency(ctx context.Context, rClient *client.Rekor,
+	oldStateHash []byte, oldTreeSize int64, sth util.SignedCheckpoint, treeID string) error {
```
Hayden-IO: nit: Should the function take a signed checkpoint, or just take old hash/size and current hash/size? I think it might be a cleaner abstraction that way.

pkg/verify/verify.go
```diff
@@ -0,0 +1,203 @@
...
+	switch {
+	case oldTreeSize == int64(sth.Size):
+		if !bytes.Equal(oldStateHash, sth.Hash) {
+			return errors.New("old root hash does not match STH hash")
+		}
+	case oldTreeSize < int64(sth.Size):
+		consistencyParams := tlog.NewGetLogProofParamsWithContext(ctx)
+		consistencyParams.FirstSize = &oldTreeSize   // Root hash at the time the proof was returned
```
Hayden-IO: nit: Comments might not be accurate depending on the use. For example, given an inclusion proof, you may already have a trusted state, and want to verify consistency between that state and the proof.

pkg/verify/verify.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/sigstore/rekor/pkg/generated/client/tlog"
+	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/sigstore/rekor/pkg/util"
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"github.com/transparency-dev/merkle/proof"
+	"github.com/transparency-dev/merkle/rfc6962"
+)
+
+func ProveConsistency(ctx context.Context, rClient *client.Rekor,
```
Hayden-IO, line 40: nit: Function comment

pkg/verify/verify.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	// Verify the root hash.
+	rootHash, err := hex.DecodeString(*e.Verification.InclusionProof.RootHash)
+	if err != nil {
+		return err
+	}
+	tid := fmt.Sprintf("%x", *e.LogID)
+
+	if err := VerifyRootHash(ctx, rClient, verifier,
```
Hayden-IO: Messaged offline about this, but I don't think this is necessary. Consistency should be verified from a known trusted state I believe. Otherwise, you're effectively asking the log for both ends of the consistency proof, so it could forge it.

I think this function should optionally take in a signed tree head to verify consistency from it to the root hash returned by the inclusion proof.

pkg/verify/verify.go
```diff
@@ -0,0 +1,203 @@
...
+		return errors.New("inclusion proof returned a tree size larger than the verified tree size")
+	}
+	return nil
+
+}
+
+// VerifyRootHash verifies the provided root hash at the treeSize by verifying consistenct
+// against an STH.
+//nolint
+func VerifyRootHash(ctx context.Context, rClient *client.Rekor, verifier signature.Verifier,
```
Hayden-IO: See comment below, but I think this function still has value, just used differently. I could see this used by witnesses to calculate consistency from a trusted state (`rootHash`) to current. a nit, maybe a different function name, like VerifyCurrentCheckpoint?

pkg/verify/verify.go
dlorenc: I think this is why we never added it in cosign. Adding this check is essentially only to prevent split tree attacks, and mitigation for that requires storing client state in between invocations.

I'm +1 on doing this (and adding the library function is fine), but the change in cosign will be a bit more invasive if we want to make it useful.

Hayden-IO: Yea, consistency checks are for enforcing the append-only property of a log, and would require client state. I think the function should support such a verification, and it could be added to cosign as an optional online lookup.

Split view attacks would require a gossip protocol which is out of scope for this.

asraa: Yeah, I made all the helper methods here public so people could pick and choose.

However, note there's not point in verifying an inclusion proof without some STH verification. So TBH, cosign should only verify the SET

dlorenc: I think that's correct - we could leave verification of the inclusion proof to the signer.

dlorenc: > Split view attacks would require a gossip protocol which is out of scope for this.

I think if clients store state you still get protection, it's just that the gossip protocol is manual and via Twitter or a Github issue.

priyawadhwa:
Hey @asraa is this ready for review?

asraa:
Reviewing now! sorry about that, was out for a few

pkg/verify/verify.go
asraa: Yeah, why not -- let's do STHs. It would require some refactoring from consumers who use it, but it's way cleaner!

asraa: OK. Done, but note that when proving consistency with the root hash given in the inclusion proof with a signed tree head, we have to "internally construct" an old STH. When https://github.com/sigstore/rekor/issues/988 is resolved we can fix. I added a comment TODO in there.

pkg/verify/verify.go
asraa: Ah, copy pasted. updated. those comments don't even make sense.

pkg/verify/verify.go
asraa: OK to resolve this: I'll keep `VerifyInclusion` and `VerifySignedEntryTimestamp` in `VerifyLogEntry`. 
Once https://github.com/sigstore/rekor/issues/988 is resolved, then we can add verification of the STH given in the return response against an optional old state in there too. TODO added.

I'll re-name `VerifyRootHash` to `VerifyCurrentCheckpoint`: that verifies a given checkpoint against a newly fetched checkpoint (by a consistency proof)

pkg/verify/verify.go
asraa: I think done as above.

pkg/verify/verify.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/sigstore/rekor/pkg/util"
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"github.com/transparency-dev/merkle/proof"
+	"github.com/transparency-dev/merkle/rfc6962"
+)
+
+// ProveConsistency verifies consistency between an initial, trusted STH
+// and a second new STH. Assumes that the signature on the STHs' are verified.
```
Hayden-IO: I wonder if it'd be better to check the signatures on the STHs, this seems like something that'd be easy to skip as a client. Same with `VerifyCurrentCheckpoint` - It verifies the newest checkpoint, but not the previous.

pkg/verify/verify.go
```diff
@@ -0,0 +1,198 @@
...
+	if !sth.Verify(verifier) {
+		return errors.New("signature on tree head did not verify")
+	}
+
+	// Now verify consistency up to the STH.
+	return ProveConsistency(ctx, rClient, oldSTH, &sth, treeID)
+}
+
+// VerifyInclusion verifies an entry's inclusion proof. Clients MUST also
+// verify the root hash via VerifyRootHash.
```
Hayden-IO: nit: VerifyRootHash is the old function name

Nonetheless, a consistency proof is effectively an inclusion proof for each checkpoint, correct? So just checking consistency would be sufficient if you have a previous checkpoint

Hayden-IO (commented):
Will you be adding tests?

asraa:
> Will you be adding tests?

I do have some local tests that I haven't committed yet: the trillian/merkle libs don't expose an easy way to generate proofs, so I'd have to hard-code some examples.

pkg/verify/verify.go
asraa: I'm confused about the statement "consistency proof is effectively an inclusion proof for each checkpoint".

I'll update the comment 
```
 Clients MUST either verify
// the root hash against a new STH (via VerifyCurrentCheckpoint) or against a
// trusted, existing STH (via ProveConsistency).
```

pkg/verify/verify.go
asraa: The previous/old one should already be trusted in `VerifyCurrentCheckpoint`, but I can add for rigor

asraa: Note that for that, we would need to pass in the verifier. I think including this in the doc string is fine? ProveConsistency would mean prove consistency, not verify tree head? wdyt?

pkg/verify/verify.go
Hayden-IO: Chatted offline, I was incorrect cause I was missing the context on a comment on a doc. 

The comment SG. Should we note that if you verify against the current checkpoint, the inclusion proof should have been one that's been persisted and is trusted?

asraa:
@haydentherapper added tests, didn't test things that were trivial and already tested in other APIs. feel free to add more

pkg/verify/verify.go
asraa: against the current checkpoint? then wouldn't you persist the new one? 

i think that's usage / client storage is out of scope here

pkg/verify/verify.go
```diff
@@ -0,0 +1,204 @@
...
+		return errors.New("inclusion proof returned a tree size larger than the verified tree size")
+	}
+	return nil
+
+}
+
+// VerifyCurrentCheckpoint verifies the provided checkpoint by verifying consistency
+// against a newly fetched Checkpoint.
+//nolint
+func VerifyCurrentCheckpoint(ctx context.Context, rClient *client.Rekor, verifier signature.Verifier,
```
priyawadhwa, line 79: It seems like this func assumes the we are verifying against the active shard, so can we omit the `treeID` argument? I think we can grab it from `logInfo := result.GetPayload()` later in the function

asraa: Nice catch!

pkg/verify/verify.go
Hayden-IO: Yea, I'm ok with that. An example somewhere of a few different ways of composing these functions might be useful

pkg/verify/verify.go
Hayden-IO: I was asking when you'd use VerifyInclusion then VerifyCurrentCheckpoint. What is the use case? You're right that you'd just persist the new one, so I guess my point is that you'd probably only call ProveConsistency after VerifyInclusion

priyawadhwa (dismissed):
lgtm!

pkg/verify/verify.go
Hayden-IO: Like the above comment, I think we should document the examples for various use cases to make it clear when these are to be used.

* A witness verifying from a previous checkpoint - VerifyCurrentCheckpoint
* Verifying an SET - VerifyLogEntry
* Verifying an inclusion proof - (I think) VerifyInclusion and ProveConsistency
* Verifying an inclusion proof offline - Just VerifyInclusion (and verifying STH signature, which isn't a function currently)

asraa: I think this should be in a documentation file. Documentation in code would only be understood anyway from code examples.

pkg/verify/verify.go
```diff
@@ -0,0 +1,204 @@
...
+		return fmt.Errorf("unable to verify bundle: %w", err)
+	}
+	return nil
+}
+
+// VerifyLogEntry performs verification of a LogEntry given a Rekor verifier.
+// Performs inclusion proof verification up to a verified root hash and
+// SignedEntryTimestamp verification.
+//nolint
+func VerifyLogEntry(ctx context.Context, rClient *client.Rekor, e *models.LogEntryAnon, verifier signature.Verifier) error {
```
Hayden-IO: Should this take a `rClient`? I think we'll need a function that is solely for offline verification (Inclusion, STH signature, SET), and one for online (Inclusion, signature, consistency, maybe SET too?)

asraa: I can remove now that we omit STH verification.

asraa: The one that also verifies consistency can be added as a separate function, agreed, like `VerifyLogEntryWithConsistency`

pkg/verify/verify.go
Hayden-IO: Discussed, examples to be added elsewhere.

Hayden-IO (dismissed):

pkg/verify/verify.go
Hayden-IO, line 79: Sorry, one last comment - Should this return the newly verified checkpoint?

asraa: You can modify it if need be.

asraa:
@priyawadhwa @bobcallaway  sorry, could I get one more look on this?

pkg/verify/verify.go
asraa: easy enough, done

priyawadhwa (approved):
nice!

Hayden-IO (approved):

PR #1011, opened 2022-08-29
Add bounds on number of elements in api/v1/log/entries/retrieve

### LINKED ISSUE #901: Add bounds on number of elements in api/v1/log/entries/retrieve.
The API definition for `retrieve` does not specify an upper bound for number of entries in the request. It is trivial to construct a request that takes O(many seconds) to return with a response size O(MB). Adding upper bounds on number of entries per request will let us offer an achievable SLO for this endpoint.

priyawadhwa:
Currently set the limit to 10 entries, we can increase it if users request it.

It's difficult to know how much this will affect users. Based on parsing Rekor logs in the last 1 week, there are 290 requests to this endpoint that took >1s, which I'm assuming was requesting more than 10 entries (but could also just be related to something else). A lot of these logs are manual testing @var-sdk and I were doing, and there aren't any logs on the weekend. If anyone has suggestions on how to roll this out I'm happy to make changes. 

closes https://github.com/sigstore/rekor/issues/901

[logs query](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22project-rekor%22%0Aresource.labels.location%3D%22us-central1%22%0Aresource.labels.cluster_name%3D%22sigstore-prod%22%0Aresource.labels.namespace_name%3D%22rekor-system%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fcomponent%3D%22server%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Finstance%3D%22sigstore-scaffold%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fname%3D%22rekor%22%20severity%3E%3DDEFAULT%0AjsonPayload.message%3D~%22.*%5C%22.*api%2Fv1%2Flog%2Fentries.*%22%0A-jsonPayload.message%3D~%22.*10%5C.1%5C.2%5C.16%5C.*%22%0A-jsonPayload.message%3D~%22.*GET*%22%0A-jsonPayload.message%3D~%22.*ms%22%0A-jsonPayload.message%3D~%22.*%C2%B5s%22;timeRange=2022-08-22T14:00:00.000Z%2F2022-08-29T03:35:00.000Z;cursorTimestamp=2022-08-26T20:04:59.302595357Z?project=project-rekor)

cc @var-sdk


bobcallaway:
can you also add the limit to the various fields in `openapi.yaml`?

priyawadhwa:
> can you also add the limit to the various fields in openapi.yaml?

yup, added!

openapi.yaml
```diff
@@ -522,10 +522,12 @@ definitions:
         items:
           type: string
           minItems: 1
+          maxItems: 10
```
bobcallaway: these should all be at the same level as `type:array`

bobcallaway (dismissed):
lgtm

var-sdk:
For rollout there are options:
1. Roll it out as dark-launch with no enforcement and add a metric to count occurrences where a caller would be hit by the limit. Turn on enforcement after gathering data for some bake period that demonstrates no "real" usage that would be broken.
2. Roll it out under a flag so the limit can be adjusted or killswitched if this causes problems for users.  
3. Roll it out as is and potentially break customers. We're pre-GA so no guarantees on API stability yet.

Post GA we will need to be following some combination of 1&2 for any breaking or risky change.

priyawadhwa:
I'm open to 2 or 3 for this change. @bobcallaway or @var-sdk, any preferences?

dlorenc:
I think a notification period (email and slack), wait a few days, then roll it out with a potential breakage would be fine.

priyawadhwa:
SGTM. Had to fix a merge conflict, if someone could LGTM this :)

bobcallaway (approved):

PR #1012, opened 2022-08-29
fix: use entry uuid uniformly in return responses

### LINKED ISSUE #980: API: Always return EntryUUID (treeID+leafHash) in all responses
**Description**

The server should always return the full entry UUID. Currently, the `models.LogEntry` uses keys that are only the raw UUID. See:
* https://cs.github.com/sigstore/rekor/blob/9099d400c515cbe496aff29941880b4132555d36/pkg/api/entries.go?q=repo%3Asigstore%2Frekor+models.LogEntry#L146
* https://cs.github.com/sigstore/rekor/blob/9099d400c515cbe496aff29941880b4132555d36/pkg/api/entries.go?q=repo%3Asigstore%2Frekor+models.LogEntry#L265

Return the Entry UUID uniformly. Make sure that existing clients won't break!

@bobcallaway 

**Version**

asraa:

#### Summary

Fixes https://github.com/sigstore/rekor/issues/980

#### Documentation

@priyawadhwa could you please take a look at sharding tests to make sure i didn't omit something purposely tested?

### DIFF
--- a/cmd/rekor-cli/app/verify.go
+++ b/cmd/rekor-cli/app/verify.go
@@ -31,7 +31,6 @@ import (
 	"github.com/sigstore/rekor/pkg/generated/client/entries"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/log"
-	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/types"
 	"github.com/sigstore/rekor/pkg/verify"
 )
@@ -151,12 +150,8 @@ var verifyCmd = &cobra.Command{
 		}
 
 		if viper.IsSet("uuid") {
-			uuid, err := sharding.GetUUIDFromIDString(viper.GetString("uuid"))
-			if err != nil {
+			if err := compareEntryUUIDs(viper.GetString("uuid"), o.EntryUUID); err != nil {
 				return nil, err
-			}
-			if uuid != o.EntryUUID {
-				return nil, fmt.Errorf("unexpected entry returned from rekor server")
 			}
 		}
 
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -100,6 +100,13 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 	}
 
 	uuid := hex.EncodeToString(leaf.MerkleLeafHash)
+	treeID := fmt.Sprintf("%x", tid)
+	entryIDstruct, err := sharding.CreateEntryIDFromParts(treeID, uuid)
+	if err != nil {
+		return nil, fmt.Errorf("error creating EntryID from active treeID %v and uuid %v: %w", treeID, uuid, err)
+	}
+	entryID := entryIDstruct.ReturnEntryIDString()
+
 	if viper.GetBool("enable_attestation_storage") {
 		pe, err := models.UnmarshalProposedEntry(bytes.NewReader(leaf.LeafValue), runtime.JSONConsumer())
 		if err != nil {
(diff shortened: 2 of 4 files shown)


tests/sharding-e2e-test.sh
```diff
@@ -243,20 +243,17 @@ fi
...
+ENTRY_ID_1=$($REKOR_CLI get --log-index 1 --rekor_server http://localhost:3000 --format json | jq -r .UUID)
+ENTRY_ID_2=$($REKOR_CLI get --log-index 3 --rekor_server http://localhost:3000 --format json | jq -r .UUID)
 
 
 # Make sure retrieve by UUID in the inactive shard works
-NUM_ELEMENTS=$(curl -f http://localhost:3000/api/v1/log/entries/retrieve -H "Content-Type: application/json" -H "Accept: application/json" -d "{ \"entryUUIDs\": [\"$UUID1\"]}" | jq '. | length')
+NUM_ELEMENTS=$(curl -f http://localhost:3000/api/v1/log/entries/retrieve -H "Content-Type: application/json" -H "Accept: application/json" -d "{ \"entryUUIDs\": [\"$ENTRY_ID_1\"]}" | jq '. | length')
 stringsMatch $NUM_ELEMENTS "1"
 
 HEX_INITIAL_TREE_ID=$(printf "%x" $INITIAL_TREE_ID | awk '{ for(c = 0; c < 16 ; c++) s = s"0"; s = s$1; print substr(s, 1 + length(s) - 16);}')
```
priyawadhwa: I think you can remove `HEX_INITIAL_TREE_ID` and `HEX_INITIAL_SHARD_ID` now that we are no longer calculating the EntryID ourselves!

pkg/api/entries.go
```diff
@@ -96,6 +96,13 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 	}
 
 	uuid := hex.EncodeToString(leaf.MerkleLeafHash)
+	activeTree := fmt.Sprintf("%x", tc.logID)
```
bobcallaway: are we guaranteed to only be pulling from the activeTree?

asraa: Hmmm I assumed the passed `tc` was the one who's log ID would be matching, but I also see a `tid` parameter in this function.

Regardless, if the comment was just on naming, yes, i don't think this is always the active tree. will update

asraa:
Trying to figure out the json end of line issue with attestations...

asraa:
Fixed, it was an error handling mistake! I thought `!errors.Is` would only work on non-nil errors, which is stupid, because it's negated.

asraa:
@lkatalin could i get your LGTM on this? I'm not sure if I'm breaking this comment. I'm allowing long EntryUUIDs to be passed in
```
// NOTE: This undoes the change that let people pass in longer UUIDs without
// trouble even if their client is old, a.k.a. it will be able to use the TreeID
// (if present) for routing in the GetLogEntryByUUIDHandler
```

priyawadhwa:
> I'm not sure if I'm breaking this comment. I'm allowing long EntryUUIDs to be passed in

I think that comment addresses an earlier change in which EntryIDs would be shortened to UUIDs since they were unsupported. That should have been fixed in #671, so I think this change is safe.

This LGTM, think it just needs a rebase.

cmd/rekor-cli/app/verify.go
```diff
@@ -164,8 +164,11 @@ var verifyCmd = &cobra.Command{
 			}
 		}
 
-		// Note: the returned entry UUID is the UUID (not include the Tree ID)
-		leafHash, _ := hex.DecodeString(o.EntryUUID)
+		outputUUID, err := sharding.GetUUIDFromIDString(o.EntryUUID)
```
priyawadhwa: Would verification in cosign also need to be updated to take into account that it might get an EntryID instead of a UUID?

asraa: Nope, I already did that! https://github.com/sigstore/cosign/blob/af575bb5a260d9f28bf5a9814c16a7c08923221e/pkg/cosign/tlog.go#L288

there are tests for those too

priyawadhwa: awesome 🎉

asraa:
all set! (or I can ninja remove that comment since it seems to no longer apply?)

priyawadhwa:
> or I can ninja remove that comment since it seems to no longer apply?

up to you! i'll lgtm for now & i can lgtm again if you want to remove it

priyawadhwa (dismissed):

priyawadhwa (approved):

bobcallaway:
this change lgtm, I wonder when/if it would be good to update openapi.yaml to only accept the full entryID (80 characters) instead of the 64 character leaf hash we started with.

asraa:
> this change lgtm, I wonder when/if it would be good to update openapi.yaml to only accept the full entryID (80 characters) instead of the 64 character leaf hash we started with.

I'll file a tracking issue for this for some discussion. i THINK people will either have the full entry UUID, OR they will be querying by proposed entry or log index, so i think it will be safe to start restricting but not sure.

Issue #988, opened 2022-08-18
Verification: Return an STH in a Verification response


asraa:
**Description**

I propose adding an STH in a rekor entry `Verification`. 

Right now, a full verification on an entry would do the following:
1. Verify the inclusion proof using the entry Body as the leaf hash
2. Verify the root hash against an STH
3. Verify the SET

See impl here: https://github.com/sigstore/rekor/blob/3ee08104c8a72c1adbb914adb8178041981d7c99/pkg/verify/verify.go#L192

Note that most of this can be done with the returned info in an entry LogEntry EXCEPT 2. That requires a Rekor client to fetch `LogInfo` to retrieved an STH and then also a consistency proof from the root hash of the inclusion proof to the STH. Instead, we *should* return the STH for the root hash of the inclusion proof. Overall, the struct would look something like
```
Entry
  Verification:
    - InclusionProof
    - SignedTreeHead (correspnding to root hash of inclusion proof)
    - SignedEntryTimestamp
```
alternatively we can also place the SignedTreeHead checkpoint inside the inclusion proof, and remove the roothash/treesize fields and consume them from the sth.

Related https://github.com/sigstore/rekor/pull/986

cc @haydentherapper @bobcallaway 

**Version**

Hayden-IO:
Wrote up a doc with more details on offline and online verification - https://docs.google.com/document/d/1_Ykbb1n_rXQjZIairCBKj6ZTpvw5s19fjpwGTksGCbI/edit# (sigstore-dev has access)

Hayden-IO:
Task: Update verification functions once the STH is returned

Hayden-IO:
The plan for changes:
* Sign a tree head on successful inclusion
* Return the tree head in the response - I'd propose that a `checkpoint` field be added to `InclusionProof` instead of alongside `InclusionProof`. Then verification is either an SET or inclusion proof (A checkpoint doesn't need to be included with an SET)
* Persist the checkpoint in a database (including the current tree, size, etc)
* Change GetLogInfo to fetch the latest checkpoint from the db

The last step is necessary for gossiping. We can't sign a new tree head - Otherwise the tree head signatures would differ between an inclusion proof's checkpoint and GetLogInfo's checkpoint.

@asraa @bobcallaway @priyawadhwa Please review

asraa:
> Then verification is either an SET or inclusion proof (A checkpoint doesn't need to be included with an SET)

[...]

PR #1015, opened 2022-08-30
Include checkpoint (STH) in entry upload and retrieve responses

Hayden-IO:
This associates a root hash in an inclusion proof with a signed
commitment from the log. Previously, without this included, there was no
connection between an inclusion proof and the log. An inclusion proof
and checkpoint can be an alternative proof of inclusion instead of a
SET.

Ref #988

#### Release Note

Added a signed checkpoint when requesting an inclusion proof

### DIFF
--- a/cmd/rekor-cli/app/upload.go
+++ b/cmd/rekor-cli/app/upload.go
@@ -141,6 +141,16 @@ var uploadCmd = &cobra.Command{
 		if err := verify.VerifySignedEntryTimestamp(ctx, &logEntry, verifier); err != nil {
 			return nil, fmt.Errorf("unable to verify entry was added to log: %w", err)
 		}
+		if logEntry.Verification.InclusionProof != nil {
+			// verify inclusion proof
+			if err := verify.VerifyInclusion(ctx, &logEntry); err != nil {
+				return nil, fmt.Errorf("error verifying inclusion proof: %w", err)
+			}
+			// verify checkpoint
+			if err := verify.VerifyCheckpointSignature(&logEntry, verifier); err != nil {
+				return nil, err
+			}
+		}
 
 		return &uploadCmdOutput{
 			Location: string(resp.Location),
--- a/cmd/rekor-cli/app/upload.go
+++ b/cmd/rekor-cli/app/upload.go
@@ -141,6 +141,7 @@ var uploadCmd = &cobra.Command{
 		if err := verify.VerifySignedEntryTimestamp(ctx, &logEntry, verifier); err != nil {
 			return nil, fmt.Errorf("unable to verify entry was added to log: %w", err)
 		}
+		// TODO: Remove conditional once inclusion proof/checkpoint is always returned by server.
 		if logEntry.Verification.InclusionProof != nil {
 			// verify inclusion proof
 			if err := verify.VerifyInclusion(ctx, &logEntry); err != nil {
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -42,6 +42,7 @@ import (
 	"github.com/sigstore/rekor/pkg/log"
 	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/types"
+	"github.com/sigstore/rekor/pkg/util"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
 )
@@ -92,11 +93,32 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 		return nil, fmt.Errorf("signing entry error: %w", err)
 	}
 
+	// sign a checkpoint as a commitment to the current root hash
+	sth, err := util.CreateSignedCheckpoint(util.Checkpoint{
+		Origin: fmt.Sprintf("%s - %d", viper.GetString("rekor_server.hostname"), tc.logID),
+		Size:   root.TreeSize,
+		Hash:   root.RootHash,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling checkpoint: %w", err)
+	}
+	sth.SetTimestamp(uint64(*logEntryAnon.IntegratedTime))
+	_, err = sth.Sign(viper.GetString("rekor_server.hostname"), api.signer, options.WithContext(ctx))
+	if err != nil {
+		return nil, fmt.Errorf("error signing checkpoint: %w", err)
+	}
+	scBytes, err := sth.SignedNote.MarshalText()
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling checkpoint: %w", err)
+	}
+	scString := string(scBytes)
+
 	inclusionProof := models.InclusionProof{
-		TreeSize: swag.Int64(int64(root.TreeSize)),
-		RootHash: swag.String(hex.EncodeToString(root.RootHash)),
-		LogIndex: swag.Int64(proof.GetLeafIndex()),
-		Hashes:   hashes,
+		TreeSize:   swag.Int64(int64(root.TreeSize)),
+		RootHash:   swag.String(hex.EncodeToString(root.RootHash)),
+		LogIndex:   swag.Int64(proof.GetLeafIndex()),
+		Hashes:     hashes,
+		Checkpoint: &scString,
 	}
 
 	uuid := hex.EncodeToString(leaf.MerkleLeafHash)
@@ -261,7 +283,45 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 		return nil, handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("signing entry error: %v", err), signingError)
 	}
 
+	root := &ttypes.LogRootV1{}
+	if err := root.UnmarshalBinary(resp.getLeafAndProofResult.SignedLogRoot.LogRoot); err != nil {
+		return nil, handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("error unmarshalling log root: %v", err), sthGenerateError)
+	}
+	hashes := []string{}
+	for _, hash := range resp.getLeafAndProofResult.Proof.Hashes {
+		hashes = append(hashes, hex.EncodeToString(hash))
+	}
+
+	// sign a checkpoint as a commitment to the current root hash
+	sth, err := util.CreateSignedCheckpoint(util.Checkpoint{
+		Origin: fmt.Sprintf("%s - %d", viper.GetString("rekor_server.hostname"), tc.logID),
+		Size:   root.TreeSize,
+		Hash:   root.RootHash,
+	})
+	if err != nil {
+		return nil, handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("error creating checkpoint: %v", err), sthGenerateError)
+	}
+	sth.SetTimestamp(uint64(*logEntryAnon.IntegratedTime))
+	_, err = sth.Sign(viper.GetString("rekor_server.hostname"), api.signer, options.WithContext(ctx))
+	if err != nil {
+		return nil, handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("error signing checkpoint: %v", err), sthGenerateError)
+	}
+	scBytes, err := sth.SignedNote.MarshalText()
+	if err != nil {
+		return nil, handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("error marshalling checkpoint: %v", err), sthGenerateError)
+	}
+	scString := string(scBytes)
+
+	inclusionProof := models.InclusionProof{
+		TreeSize:   swag.Int64(int64(root.TreeSize)),
+		RootHash:   swag.String(hex.EncodeToString(root.RootHash)),
+		LogIndex:   swag.Int64(queuedLeaf.LeafIndex),
+		Hashes:     hashes,
+		Checkpoint: &scString,
+	}
+
 	logEntryAnon.Verification = &models.LogEntryAnonVerification{
+		InclusionProof:       &inclusionProof,
 		SignedEntryTimestamp: strfmt.Base64(signature),
 	}
 
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -93,7 +93,7 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 		return nil, fmt.Errorf("signing entry error: %w", err)
 	}
 
-	scBytes, err := util.CreateAndSignCheckpoint(viper.GetString("rekor_server.hostname"), tc.logID, root, api.signer, ctx)
+	scBytes, err := util.CreateAndSignCheckpoint(ctx, viper.GetString("rekor_server.hostname"), tc.logID, root, api.signer)
 	if err != nil {
 		return nil, err
 	}
@@ -277,7 +277,7 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 		hashes = append(hashes, hex.EncodeToString(hash))
 	}
 
-	scBytes, err := util.CreateAndSignCheckpoint(viper.GetString("rekor_server.hostname"), tc.logID, root, api.signer, ctx)
+	scBytes, err := util.CreateAndSignCheckpoint(ctx, viper.GetString("rekor_server.hostname"), tc.logID, root, api.signer)
 	if err != nil {
 		return nil, handleRekorAPIError(params, http.StatusInternalServerError, err, sthGenerateError)
 	}
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -93,32 +93,17 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 		return nil, fmt.Errorf("signing entry error: %w", err)
 	}
 
-	// sign a checkpoint as a commitment to the current root hash
-	sth, err := util.CreateSignedCheckpoint(util.Checkpoint{
-		Origin: fmt.Sprintf("%s - %d", viper.GetString("rekor_server.hostname"), tc.logID),
-		Size:   root.TreeSize,
-		Hash:   root.RootHash,
-	})
+	scBytes, err := util.CreateAndSignCheckpoint(viper.GetString("rekor_server.hostname"), tc.logID, root, api.signer, ctx)
 	if err != nil {
-		return nil, fmt.Errorf("error marshalling checkpoint: %w", err)
+		return nil, err
 	}
-	sth.SetTimestamp(uint64(*logEntryAnon.IntegratedTime))
-	_, err = sth.Sign(viper.GetString("rekor_server.hostname"), api.signer, options.WithContext(ctx))
-	if err != nil {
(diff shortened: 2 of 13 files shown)


Hayden-IO:
Tests are lacking - Not sure what's the best way to test this without significant refactoring. I added a check that a checkpoint is present on `rekor-cli verify`, and tested locally.

Hayden-IO:
@asraa, I can wait until #986 is in to avoid conflicts on the CLI.

Hayden-IO:
Sample response from an upload:

```
{
  "663755961b7b8f25a2134247f5ec4592fc1758c42e7ad158dd0a0cc0d2c1a4c3": {
    "body": {
      "apiVersion": "0.0.1",
      "kind": "rekord",
      "spec": {
        "data": {
          "hash": {
            "algorithm": "sha256",
            "value": "55148d64b214f0b75069feb4429a621fa225b8aabd0a4558bee6769c21d459d8"
          }
        },
        "signature": {
          "content": "MGUCMFHT3GKCpBzp1g+MUY44b4iTUIZlqHx4AY9jCyXtY8FVl3CxBAhQprOsjT4vyD4tUgIxAOBnm689mrROuWYqoU/Qb+hUHtvTvz2dqftKicadY3FVCNyHs/dLe+lNtNa9Lx3VBQ==",
          "format": "x509",
          "publicKey": {
            "content": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUM0RENDQW1XZ0F3SUJBZ0lVUUQ5a25mblJINFFoMGxFa2lqdE9XTndJSTNRd0NnWUlLb1pJemowRUF3TXcKTnpFVk1CTUdBMVVFQ2hNTWMybG5jM1J2Y21VdVpHVjJNUjR3SEFZRFZRUURFeFZ6YVdkemRHOXlaUzFwYm5SbApjbTFsWkdsaGRHVXdIaGNOTWpJd09ESTVNak13TlRVeFdoY05Nakl3T0RJNU1qTXhOVFV4V2pBQU1IWXdFQVlICktvWkl6ajBDQVFZRks0RUVBQ0lEWWdBRTRjMXZNMWQvV3R6ckhaQThYNVJMNFFZTU15Znk0SVlkWnFGZVF0UzEKQW9LdldHc2dQalhxSDIzTVhDREpOZit6YWNTQUMrZWF6RmpsQmpxaXFuVHJSeHNhNjRod0xocU5McTg4QVZ2bwozTlJyaXI1emc4N0JGTzBVSTlHUGZnMnNvNElCWnpDQ0FXTXdEZ1lEVlIwUEFRSC9CQVFEQWdlQU1CTUdBMVVkCkpRUU1NQW9HQ0NzR0FRVUZCd01ETUIwR0ExVWREZ1FXQkJUbVFMeE40dDAvcXhjTCtsZkVCbm9zb0RhT1N6QWYKQmdOVkhTTUVHREFXZ0JUZjArblBWaVFSbHZtbzJPa29WYUxHTGhoa1B6QkVCZ05WSFJFQkFmOEVPakE0Z1RaMApaWE4wTFdGalkyOTFiblJBYUdKc1lYVjZkbVZ5YmkxMFpYTjBMVEV1YVdGdExtZHpaWEoyYVdObFlXTmpiM1Z1CmRDNWpiMjB3S1FZS0t3WUJCQUdEdnpBQkFRUWJhSFIwY0hNNkx5OWhZMk52ZFc1MGN5NW5iMjluYkdVdVkyOXQKTUlHS0Jnb3JCZ0VFQWRaNUFnUUNCSHdFZWdCNEFIWUFDR0NTOENoUy8yaEYwZEZySjRTY1JXY1lyQlk5d3pqUwpiZWE4SWdZMmIzSUFBQUdDNjlxbElnQUFCQU1BUnpCRkFpRUE3cmhpYlJ1bys2OFRGL1JsRDZVZU9YeURGS3dECkZHanp6UzNsSDFaWnU5VUNJRTJVTzJodnBKWHFuTG00NnhCYllLemQyemdjVy9HcDVFUmszZ1o0MHRlSE1Bb0cKQ0NxR1NNNDlCQU1EQTJrQU1HWUNNUUNaOHRLYVg4V3dPeUw1eEJQVWg2cXVWV3J5di9DSkFjenhkRmlCSG1SKwpZZmUvSytaQU01TnVqam93T1YwaTFkc0NNUUR6N1A4T3hjUUNKUm1MZHNTSjVsVnhqcW9UY1R2c1llakx1VDF4CnlwKzNEd2ttRXp3UGJpNVp5czJyVzFWVGtDUT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQotLS0tLUJFR0lOIENFUlRJRklDQVRFLS0tLS0KTUlJQ0dqQ0NBYUdnQXdJQkFnSVVBTG5WaVZmblUwYnJKYXNtUmtIcm4vVW5mYVF3Q2dZSUtvWkl6ajBFQXdNdwpLakVWTUJNR0ExVUVDaE1NYzJsbmMzUnZjbVV1WkdWMk1SRXdEd1lEVlFRREV3aHphV2R6ZEc5eVpUQWVGdzB5Ck1qQTBNVE15TURBMk1UVmFGdzB6TVRFd01EVXhNelUyTlRoYU1EY3hGVEFUQmdOVkJBb1RESE5wWjNOMGIzSmwKTG1SbGRqRWVNQndHQTFVRUF4TVZjMmxuYzNSdmNtVXRhVzUwWlhKdFpXUnBZWFJsTUhZd0VBWUhLb1pJemowQwpBUVlGSzRFRUFDSURZZ0FFOFJWUy95c0grTk92dURaeVBJWnRpbGdVRjlObGFyWXBBZDlIUDF2QkJIMVU1Q1Y3CjdMU1M3czBaaUg0bkU3SHY3cHRTNkx2dlIvU1RrNzk4TFZnTXpMbEo0SGVJZkYzdEhTYWV4TGNZcFNBU3Ixa1MKME4vUmdCSnovOWpXQ2lYbm8zc3dlVEFPQmdOVkhROEJBZjhFQkFNQ0FRWXdFd1lEVlIwbEJBd3dDZ1lJS3dZQgpCUVVIQXdNd0VnWURWUjBUQVFIL0JBZ3dCZ0VCL3dJQkFEQWRCZ05WSFE0RUZnUVUzOVBwejFZa0VaYjVxTmpwCktGV2l4aTRZWkQ4d0h3WURWUjBqQkJnd0ZvQVVXTUFlWDVGRnBXYXBlc3lRb1pNaTBDckZ4Zm93Q2dZSUtvWkkKemowRUF3TURad0F3WkFJd1BDc1FLNERZaVpZRFBJYURpNUhGS25meFh4NkFTU1ZtRVJmc3luWUJpWDJYNlNKUgpuWlU4NC85RFpkbkZ2dnhtQWpCT3Q2UXBCbGM0Si8wRHh2a1RDcXBjbHZ6aUw2QkNDUG5qZGxJQjNQdTNCeHNQCm15Z1VZN0lpMnpiZENkbGlpb3c9Ci0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0KLS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUI5ekNDQVh5Z0F3SUJBZ0lVQUxaTkFQRmR4SFB3amVEbG9Ed3lZQ2hBTy80d0NnWUlLb1pJemowRUF3TXcKS2pFVk1CTUdBMVVFQ2hNTWMybG5jM1J2Y21VdVpHVjJNUkV3RHdZRFZRUURFd2h6YVdkemRHOXlaVEFlRncweQpNVEV3TURjeE16VTJOVGxhRncwek1URXdNRFV4TXpVMk5UaGFNQ294RlRBVEJnTlZCQW9UREhOcFozTjBiM0psCkxtUmxkakVSTUE4R0ExVUVBeE1JYzJsbmMzUnZjbVV3ZGpBUUJnY3Foa2pPUFFJQkJnVXJnUVFBSWdOaUFBVDcKWGVGVDRyYjNQUUd3UzRJYWp0TGszL09sbnBnYW5nYUJjbFlwc1lCcjVpKzR5bkIwN2NlYjNMUDBPSU9aZHhleApYNjljNWlWdXlKUlErSHowNXlpK1VGM3VCV0FsSHBpUzVzaDArSDJHSEU3U1hyazFFQzVtMVRyMTlMOWdnOTJqCll6QmhNQTRHQTFVZER3RUIvd1FFQXdJQkJqQVBCZ05WSFJNQkFmOEVCVEFEQVFIL01CMEdBMVVkRGdRV0JCUlkKd0I1ZmtVV2xacWw2ekpDaGt5TFFLc1hGK2pBZkJnTlZIU01FR0RBV2dCUll3QjVma1VXbFpxbDZ6SkNoa3lMUQpLc1hGK2pBS0JnZ3Foa2pPUFFRREF3TnBBREJtQWpFQWoxbkhlWFpwKzEzTldCTmErRURzRFA4RzFXV2cxdENNCldQL1dIUHFwYVZvMGpoc3dlTkZaZ1NzMGVFN3dZSTRxQWpFQTJXQjlvdDk4c0lrb0YzdlpZZGQzL1Z0V0I1YjkKVE5NZWE3SXgvc3RKNVRmY0xMZUFCTEU0Qk5KT3NRNHZuQkhKCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K"
          }
        }
      }
    },
    "integratedTime": 1661822960,
    "logID": "bb74d288f20e47568e97216e4e5cbed892cc8a4e634b1a7d9c4fcfc90c00c0ac",
    "logIndex": 2,
    "verification": {
      "inclusionProof": {
        "checkpoint": "71ea7e61d1c2 - 2187580196092226528\n3\nPTV9PPDu0TbAAMA+wIQBMztMTsaArMvsfParYzgldi8=\nTimestamp: 1661822960\n\n— 71ea7e61d1c2 u3TSiDBEAiB0cnATML2WLPlX8D2woVhUF6tRVzPkM0o/iC17jSvjeAIgQt6nkFDegIRqmtMaCYnbqFK66Iw1GvJbx4OfyCenmas=\n",
        "hashes": [
          "219bb939b401f04ab9eafe81e99b2e58e9da2ef0f0e4496f2defdd32fb119c95"
        ],
        "logIndex": 2,
        "rootHash": "3d357d3cf0eed136c000c03ec08401333b4c4ec680accbec7cf6ab633825762f",
        "treeSize": 3
      },
      "signedEntryTimestamp": "MEUCIBuwYaQ51NUOczXf/1hEMTUgErapNnlIgJ36T0pZjv06AiEAwA/JXxMWKi6ZQcGIhfWWYHK4/LkSlPJn/yBfJ4kbDZ0="
    }
  }
}```

File: cmd/rekor-cli/app/upload.go [hunk 1]
```diff
@@ -185,6 +190,43 @@ func verifyLogEntry(ctx context.Context, rekorClient *genclient.Rekor, logEntry
 	if !ecdsa.VerifyASN1(rekorPubKey, hash[:], []byte(logEntry.Verification.SignedEntryTimestamp)) {
 		return false, fmt.Errorf("unable to verify")
 	}
+
+	// verify inclusion proof if present
+	if logEntry.Verification.InclusionProof != nil {
```
asraa: shouldn't the inclusion proof always be present? or is there a case (maybe first upload) that it is not returned?

otherwise just gate checkpoint check for backwards compat

pkg/api/entries.go
```diff
@@ -92,11 +93,32 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 		return nil, fmt.Errorf("signing entry error: %w", err)
 	}
 
+	// sign a checkpoint as a commitment to the current root hash
+	sth, err := util.CreateSignedCheckpoint(util.Checkpoint{
+		Origin: fmt.Sprintf("%s - %d", viper.GetString("rekor_server.hostname"), tc.logID),
```
asraa: can we abstract this repeated Origin creation?

pkg/api/entries.go
```diff
@@ -92,11 +93,32 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
...
+		Size:   root.TreeSize,
+		Hash:   root.RootHash,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling checkpoint: %w", err)
+	}
+	sth.SetTimestamp(uint64(*logEntryAnon.IntegratedTime))
```
asraa: this seems odd? why is the STH's timestamp on the log entry's integrated time? it SHOULD be the time when the STH was generated.

pkg/api/entries.go
```diff
@@ -92,11 +93,32 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
...
+	_, err = sth.Sign(viper.GetString("rekor_server.hostname"), api.signer, options.WithContext(ctx))
```
asraa: you can inline this 
```
if _, err := ... ; err != nil {}
```

pkg/api/entries.go
```diff
@@ -92,11 +93,32 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
...
+	if err != nil {
+		return nil, fmt.Errorf("error signing checkpoint: %w", err)
+	}
+	scBytes, err := sth.SignedNote.MarshalText()
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling checkpoint: %w", err)
+	}
+	scString := string(scBytes)
```
asraa: you can use `stringPointer(string(scBytes))` below

pkg/api/entries.go
```diff
@@ -92,11 +93,32 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, tc TrillianC
 		return nil, fmt.Errorf("signing entry error: %w", err)
 	}
 
+	// sign a checkpoint as a commitment to the current root hash
+	sth, err := util.CreateSignedCheckpoint(util.Checkpoint{
```
asraa: overall is there any way to abstract this to a func? i think it's called about 4 separate times in this code, all the same

cmd/rekor-cli/app/upload.go
Hayden-IO: Upload does not currently include an inclusion proof. Code: https://github.com/sigstore/rekor/blob/main/pkg/api/entries.go#L264-L266. Tested with `curl -s https://rekor.sigstore.dev/api/v1/log/entries -H "Accept: application/json" -H "Content-Type: application/json" -o rekor_output --data-binary " { \"apiVersion\": \"0.0.1\", \"kind\": \"rekord\", \"spec\": { \"signature\": { \"format\": \"x509\", \"content\": \"$(base64 -w0 artifact.sig)\", \"publicKey\": { \"content\": \"$(base64 -w0 signingCertChain.pem)\" } }, \"data\": { \"content\": \"$(base64 -w0 artifact)\" } } }"` and verified it does not include an inclusion proof.

I added this for backwards compatibility. For upload, the inclusion proof and checkpoint are missing, so both are gated under this conditional. Verify includes the inclusion proof, but not checkpoint, so I only needed to gate the checkpoint check.

pkg/api/entries.go
Hayden-IO: Yep, agreed, I don't recall why I used this time. Changing to `sth.SetTimestamp(uint64(time.Now().UnixNano()))` (and will refactor to have only one function)

cmd/rekor-cli/app/upload.go
asraa: Gotcha, thanks! Confusing behavior.

pkg/api/entries.go
Hayden-IO: Done

pkg/api/entries.go
Hayden-IO: Done, moved to a new function (with tests!)

Hayden-IO:
Addressed feedback, there's a lot less code duplication now.

cmd/rekor-cli/app/upload.go
```diff
@@ -141,6 +141,16 @@ var uploadCmd = &cobra.Command{
 		if err := verify.VerifySignedEntryTimestamp(ctx, &logEntry, verifier); err != nil {
 			return nil, fmt.Errorf("unable to verify entry was added to log: %w", err)
 		}
+		if logEntry.Verification.InclusionProof != nil {
+			// verify inclusion proof
+			if err := verify.VerifyInclusion(ctx, &logEntry); err != nil {
+				return nil, fmt.Errorf("error verifying inclusion proof: %w", err)
+			}
+			// verify checkpoint
+			if err := verify.VerifyCheckpointSignature(&logEntry, verifier); err != nil {
```
asraa, line 151: what if checkpoint is nil? from old versions.

pkg/util/checkpoint.go
```diff
@@ -160,3 +166,24 @@ func (r *SignedCheckpoint) GetTimestamp() uint64 {
...
+
+// CreateAndSignCheckpoint creates a signed checkpoint as a commitment to the current root hash
+func CreateAndSignCheckpoint(ctx context.Context, hostname string, treeID int64, root *types.LogRootV1, signer signature.Signer) ([]byte, error) {
+	sth, err := CreateSignedCheckpoint(Checkpoint{
+		Origin: fmt.Sprintf("%s - %d", hostname, treeID),
+		Size:   root.TreeSize,
+		Hash:   root.RootHash,
+	})
+	if err != nil {
+		return []byte{}, fmt.Errorf("error creating checkpoint: %v", err)
```
asraa: you can return `nil` for an empty byte slice

pkg/verify/verify.go
```diff
@@ -187,7 +201,7 @@ func VerifySignedEntryTimestamp(ctx context.Context, e *models.LogEntryAnon, ver
 // VerifyLogEntry performs verification of a LogEntry given a Rekor verifier.
 // Performs inclusion proof verification up to a verified root hash and
 // SignedEntryTimestamp verification.
-//nolint
+// nolint
```
asraa, line 204: worth adding checkpoint verification here, see the TODO in there?

asraa (changes requested):
thanks! looking great

cmd/rekor-cli/app/upload.go
Hayden-IO: In this upload case, we weren’t returning an inclusion proof previously. So this case handles both, since the checkpoint is a part of the inclusion proof now.

asraa: Argh! I see, all of it's added into `createLogEntry` and we already return if we have a pre-existing entry here. gotcha, thx

Hayden-IO: Yea, exactly. For the other cases, we returned an inclusion, but just not the checkpoint.

pkg/util/checkpoint.go
Hayden-IO: Thanks, done!

pkg/verify/verify.go
Hayden-IO: Agh, one issue is that when we cut a release of Rekor's CLI, the server won't have rolled out yet. So I'll add this in with a guard around the check, and leave a TODO to a) remove the guard, b) consistency check against an optionally provided checkpoint

bobcallaway (approved):
lgtm

asraa (approved):
Yay!

Hayden-IO:
Thanks y'all, it's ready to merge!

priyawadhwa (approved):
nice!!

priyawadhwa (approved):
nice!!

PR #1017, opened 2022-08-30
Validate tree ID on calls to /api/v1/log/entries/retrieve

### LINKED ISSUE #1014: Rekor does not validate Tree ID in requests for `entries/retrieve`
**Description**

The Tree ID is ignored in the `retrieve` endpoint. 

```
$ curl -f https://rekor.sigstore.dev/api/v1/log/entries/retrieve -H "Content-Type: application/json" -H "Accept: application/json" -d "{ \"entryUUIDs\": [\"362f8ecba72f4325673d587e82b8819efe9261b541d742f35312bbb54cc9d8eecd7bef2a1afb9045\"]}"
```

and 
```
$ curl -f https://rekor.sigstore.dev/api/v1/log/entries/retrieve -H "Content-Type: application/json" -H "Accept: application/json" -d "{ \"entryUUIDs\": [\"362f8ecba72f4326673d587e82b8819efe9261b541d742f35312bbb54cc9d8eecd7bef2a1afb9045\"]}"  
[{"673d587e82b8819efe9261b541d742f35312bbb54cc9d8eecd7bef2a1afb9045"
```

Note the tree ID difference `362f8ecba72f4326` and `362f8ecba72f4325`. We retrieve the same UUID, but requested different Tree IDs. Note one does not exist.

priyawadhwa:
fixes https://github.com/sigstore/rekor/issues/1014, cc @asraa 

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -338,11 +338,11 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 				resultPayload = append(resultPayload, logEntry)
 				continue
 			}
-			uuid, err := sharding.GetUUIDFromIDString(entryID)
-			if err != nil {
-				return handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf("could not get UUID from ID string %v", entryID))
+			// At this point, check if we got a uuid instead of an EntryID, so search for the hash later
+			uuid := entryID
+			if err := sharding.ValidateUUID(uuid); err != nil {
+				return handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf("validating uuid %s", uuid))
 			}
-			// If we couldn't get the entry, search for the hash later
 			hash, err := hex.DecodeString(uuid)
 			if err != nil {
 				return handleRekorAPIError(params, http.StatusBadRequest, err, malformedUUID)
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -330,13 +330,17 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 
 		var searchHashes [][]byte
 		for _, entryID := range params.Entry.EntryUUIDs {
+			if sharding.ValidateEntryID(entryID) == nil {
+				logEntry, err := retrieveLogEntry(httpReqCtx, entryID)
+				if err != nil {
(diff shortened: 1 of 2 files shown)


asraa (changes requested):
Interesting! Thanks for the fix and debugging. So what was happening before was that we were actually retrieving the result via `searchHashes` on the raw `uuid` it seems.

So even now, if you send an something where `sharding.ValidateEntryID(entryID)` is an error, you will continue to do that behavior and return something from a different tree.

asraa:
I'm a little brain-fried right now, but I think the fix is going to need to propogate the entire EntryUUID when you go through this case:
https://github.com/sigstore/rekor/blob/d0e2a95f465c2d959e062651d8c246a966431195/pkg/api/entries.go#L346-L350

Then in here: extract the uuid and validate the returned tree ID when one was in the EntryUUID request:
https://github.com/sigstore/rekor/blob/d0e2a95f465c2d959e062651d8c246a966431195/pkg/api/entries.go#L396

priyawadhwa:
> So even now, if you send an something where sharding.ValidateEntryID(entryID) is an error, you will continue to do that behavior and return something from a different tree.

Yep, but I think `sharding.ValidateEntryID(entryID)` should only be an error if a UUID was originally requested instead of an EntryID, in which case expected behavior (at the moment) is we would search all shards and return all matching entries since there is no Tree ID.

asraa:
> Yep, but I think sharding.ValidateEntryID(entryID) should only be an error if a UUID was originally requested instead of an EntryID,

I think it'll error out if the tree ID doesn't parse as an int64 or if it's 0 
https://github.com/sigstore/rekor/blob/5159d01f9881088937e4553c80c2674cdb0872ae/pkg/sharding/sharding.go#L153-L162

priyawadhwa:
I think it's fine if the tree is 0, since that maps to having a UUID and putting a bunch of 0s in front of it to form an EntryID.

To address the failure on parsing int64 I added in a validation for UUID so it should do something like this:

1. Check if EntryID is valid, if yes then try to get it
2. If no, assume we have a UUID and validate it. Fail if we don't have a UUID, otherwise add it to search hashes.

asraa:
> To address the failure on parsing int64 I added in a validation for UUID so it should do something like this:

That's perfect! Thank you!! That made the fix so clean

asraa (approved):

dlorenc (approved):

PR #1018, opened 2022-08-31
upgrade to go1.19

cpanato:
#### Summary
- upgrade to go1.19

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
 
       - uses: imjasonh/setup-ko@78eea08f10db87a7a23a666a4a6fe2734f2eeb8d # v0.4
         with:
-          version: v0.11.2
+          version: v0.12.0
 
       - name: Set up Cloud SDK
         uses: google-github-actions/auth@dac4e13deb3640f22e3ffe758fd3f95e6e89f712 # v0.8.1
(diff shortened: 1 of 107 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor
 
-go 1.18
+go 1.19
```
dlorenc: Does this bump need to happen? IIUC it means everyone that imports us must also use go 1.19.

cpanato: yes, that is true, i will drop that

cpanato: should we add a job that tests using go1.18?

dlorenc: Honestly I don't think we need parallel testing here, it's somewhat overkill and just magnifies flakes.

dlorenc (dismissed):

.github/workflows/build.yml
```diff
@@ -43,13 +43,14 @@ jobs:
       - uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f # v3.1.0
         with:
           go-version: ${{ env.GOVERSION }}
+          check-latest: true
```
bobcallaway: why are we adding this here? The point of calculating `env.GOVERSION` is to ensure that the dockerfile golang version is consistent with the toolchain we build with locally. This might cause them to get out of sync.

cpanato: i did an update to use the go mod as base input and reverted, but forgot to remove this, will cleanup this, thanks for the review and catch

cpanato: done

bobcallaway (approved):

PR #1022, opened 2022-09-01
remove /api/v1/version endpoint

bobcallaway:
This patch removes the /api/v1/version endpoint in an effort to not disclose the server's operating version to malicious users. This is consistent with Fulcio where we also removed this endpoint.

Note there are some [uses](https://cs.github.com/?scopeName=All+repos&scope=&q=%2Fapi%2Fv1%2Fversion+rekor) in [sigstore/scaffolding](https://github.com/sigstore/scaffolding/pull/333), [sigstore/sigstore-rs](https://github.com/sigstore/sigstore-rs/issues/120) which we should get rid of as well.


Hayden-IO (commented):
LGTM once its usage is cleaned up elsewhere. We should also notify folks on Slack asap.

priyawadhwa (approved):

PR #1030, opened 2022-09-06
Fix rekor-cli backwards incompatibility & run harness tests against HEAD 

priyawadhwa:
also updates rekor-cli to iterate through all supported versions

### DIFF
--- a/pkg/types/intoto/intoto.go
+++ b/pkg/types/intoto/intoto.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/sigstore/rekor/pkg/generated/models"
+	"github.com/sigstore/rekor/pkg/log"
 	"github.com/sigstore/rekor/pkg/types"
 	"golang.org/x/exp/slices"
 )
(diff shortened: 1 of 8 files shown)


priyawadhwa:
Couple updates:
1. Harness tests are currently failing because server @ HEAD returns an EntryID but `rekor-cli get` v0.10.0/v0.11.0 return a UUID. With #623 already merged it's a breaking change and users will need to update to rekor-cli v0.12.0 for `rekor-cli get` to work.
2. I couldn't get the iterator to work but I can give it another go tomorrow

pkg/types/intoto/intoto.go
```diff
@@ -60,9 +61,41 @@ func (it BaseIntotoType) UnmarshalEntry(pe models.ProposedEntry) (types.EntryImp
...
+		}
+		head.ProposedEntry = pe
+		next = &head
+		for _, v := range it.SupportedVersions() {
+			if v == it.DefaultVersion() {
+				continue
+			}
+			ei, err := it.VersionedUnmarshal(nil, v)
+			if err != nil {
+				log.Logger.Errorf("fetching Intoto version (%v) implementation: %w", v, err)
```
bobcallaway: can we swap `log.Logger` for `log.ContextLogger(ctx)` here and a couple lines below?

priyawadhwa: yup, done!

bobcallaway (approved):

PR #1049, opened 2022-09-13
feat: add file based signer and password

### LINKED ISSUE #1026: Add --file-ca (or something flag) for taking in secret from files like Fulcio does.
**Description**

Currently Rekor has an ephemeral signer mode here:
https://github.com/sigstore/rekor/blob/main/cmd/rekor-server/app/root.go#L76

This works mostly fine, but with us moving to TUF everywhere (and getting rid of various env flags for specifying public keys) if Rekor ever restarts during the testing this will create a new in-mem public key and hence the TUF root doesn't work. Fulcio has a way to specify a file_ca:
https://github.com/sigstore/fulcio/blob/main/cmd/app/serve.go#L69

We talked about this in the Slack and agreed this would be useful to do, we just need to mark the flag clearly that it's used for testing.

https://sigstore.slack.com/archives/C01CX4E2K70/p1659553415708999

@asraa

asraa:

#### Summary

* feat: adds flag for specifying a file-based signer with a password, useful for testing

Fixes https://github.com/sigstore/rekor/issues/1026

#### Documentation

I did add unit testing for this: as far as an integration test, the best I did was spin it up locally with a cosign generated encrypted key pair and verify that I could get the public key and log info.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -74,7 +74,9 @@ func init() {
 	rootCmd.PersistentFlags().String("rekor_server.hostname", hostname, "public hostname of instance")
 	rootCmd.PersistentFlags().String("rekor_server.address", "127.0.0.1", "Address to bind to")
 
-	rootCmd.PersistentFlags().String("rekor_server.signer", "memory", "Rekor signer to use. Current valid options include: [gcpkms, file, memory]")
+	rootCmd.PersistentFlags().String("rekor_server.signer", "memory",
+		`Rekor signer to use. Valid options are: [gcpkms, memory, filename containing PEM encoded private key].
+		Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("rekor_server.signer-passwd", "", "Password to decrypt signer private key")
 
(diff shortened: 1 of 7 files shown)


asraa:
cc @vaikas

cmd/rekor-server/app/root.go
```diff
@@ -73,7 +73,9 @@ func init() {
 	}
 	rootCmd.PersistentFlags().String("rekor_server.hostname", hostname, "public hostname of instance")
 	rootCmd.PersistentFlags().String("rekor_server.address", "127.0.0.1", "Address to bind to")
-	rootCmd.PersistentFlags().String("rekor_server.signer", "memory", "Rekor signer to use. Current valid options include: [gcpkms, memory]")
+
+	rootCmd.PersistentFlags().String("rekor_server.signer", "memory", "Rekor signer to use. Current valid options include: [gcpkms, file, memory]")
```
Hayden-IO: Can you mention file and memory should be used for testing only?

Hayden-IO (dismissed):

vaikas (approved):
Thank you so much for doing this!!!

Hayden-IO (approved):

dlorenc (approved):

PR #1059, opened 2022-09-20
Adds new rekor metrics for latency and QPS.

var-sdk:
The current server metrics have two problems:
1. They don't capture method so we can't differentiate POST vs GET/write vs read.
2. They use the path as passed by the caller for the path field. This prevents us from aggregating correctly ("api/v1/log" is a different time series than "api/v1/log/") and can be a scale risk as it lets end users create new time series arbitrarily.

The two new metrics track latency by endpoint/method,  and qps by endpoint/method/status code for error budget. The old metrics will be removed once monitoring has been updated to use the new metrics.

### DIFF
--- a/pkg/api/metrics.go
+++ b/pkg/api/metrics.go
@@ -16,6 +16,8 @@
 package api
 
 import (
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


bobcallaway (approved):
lgtm, thanks for doing this!

dlorenc (approved):

PR #1065, opened 2022-09-21
Update checkpoint origin to match previous origin, make configurable

Hayden-IO:
Changing the origin breaks witnessing, so I've reverted this change, but made it configurable so we can change the value in staging.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -78,6 +78,7 @@ func init() {
 		`Rekor signer to use. Valid options are: [gcpkms, memory, filename containing PEM encoded private key].
 		Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("rekor_server.signer-passwd", "", "Password to decrypt signer private key")
+	rootCmd.PersistentFlags().String("rekor_server.checkpoint_origin", "Rekor", "Checkpoint (STH) Origin")
 
 	rootCmd.PersistentFlags().Uint16("port", 3000, "Port to bind to")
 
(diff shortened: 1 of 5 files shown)


Hayden-IO:
/HOLD

I am chatting with the Trillian team now to find out if we need to differentiate between shards.

Hayden-IO:
unrelated e2e test failure

dlorenc:
Is there any context on the breakage here? What broke without this change?

Hayden-IO:
Sorry for lack of context. The omniwitness monitor that's run by Trust Fabric is configurable with the origin of the checkpoint - https://github.com/google/trillian-examples/blob/master/witness/golang/omniwitness/feeder_configs/rekor.yaml#L2

Verification requires a known origin value. Once we roll out 0.12 to prod, the witness will break.

dlorenc:
It seems like that config should be updated, the correct name should probably be something closer to what v0.12.0 does right? This isn't meant to be a canonical "rekor" instance, we should at least prefix it with something like rekor.sigstore.dev.

Hayden-IO:
I was planning to have different names for staging, like "Rekor Staging". Note that the hostname is already included in the checkpoint note alongside the public key too.

I have asked Trillian how we should handle sharding - should each shard be considered a unique origin? That would mean we'd need a new witness for each shard.

Hayden-IO:
I see your point about not having this be the canonical rekor instance. I'll update this tomorrow after getting clarity on if the shard should be included.

Hayden-IO:
The current checkpoint origin is ideal, and I'll update the configuration.

PR #1080, opened 2022-09-27
add ability to enable/disable specific rekor API endpoints

bobcallaway:

### DIFF
--- a/pkg/generated/restapi/configure_rekor_server.go
+++ b/pkg/generated/restapi/configure_rekor_server.go
@@ -43,7 +43,7 @@ import (
 	"github.com/sigstore/rekor/pkg/log"
 	"github.com/sigstore/rekor/pkg/util"
 
-	"github.com/urfave/negroni"
+	"golang.org/x/exp/slices"
 )
 
 //go:generate swagger generate server --target ../../generated --name RekorServer --spec ../../../openapi.yaml --principal interface{} --exclude-main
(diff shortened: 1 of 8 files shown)


dlorenc:
This seems fine, just curious what the context is.

bobcallaway:
> This seems fine, just curious what the context is.

If we wanted isolation of specific read-only VS write-only paths for scaling, it might be useful to be able to narrow the API surface down. 

This also removes a couple incorrect middleware handlers and caching rules that no longer apply.

lukehinds:
code looks good, don't forget to update https://github.com/sigstore/sigstore-website/tree/docs/content/en/rekor too.

dlorenc:
> If we wanted isolation of specific read-only VS write-only paths for scaling, it might be useful to be able to narrow the API surface down.

Not sure I follow here... Rekor itself is stateless, if we had a bunch only doing reads I'm not sure how that would help with the database scaling. Would we point them at read only replicas of mysql/trillian or something?

var-sdk:
> > If we wanted isolation of specific read-only VS write-only paths for scaling, it might be useful to be able to narrow the API surface down.
> 
> Not sure I follow here... Rekor itself is stateless, if we had a bunch only doing reads I'm not sure how that would help with the database scaling. Would we point them at read only replicas of mysql/trillian or something?

That's one possibility. We could set up read replicas in other regions so the read path continues to operate even if our write instance goes down. The other goodness here might be splitting read & write paths within a cluster, so a query of death bug in write wouldn't completely take down read.

pkg/generated/restapi/configure_rekor_server.go
```diff
@@ -84,49 +84,75 @@ func configureAPI(api *operations.RekorServerAPI) http.Handler {
...
+		case "createLogEntry":
+			recordMetricsForAPI(api, "POST", "/api/v1/log/entries") // add metrics
+		case "getLogEntryByIndex":
+			api.AddMiddlewareFor("GET", "/api/v1/log/entries", middleware.NoCache) // not cacheable
+			recordMetricsForAPI(api, "GET", "/api/v1/log/entries")                 // add metrics
+		case "getLogEntryByUUID":
+			api.AddMiddlewareFor("GET", "/api/v1/log/entries/{entryUUID}", middleware.NoCache) // not cacheable
+			recordMetricsForAPI(api, "GET", "/api/v1/log/entries/{entryUUID}")                 // add metrics
+		case "searchLogQuery":
+			recordMetricsForAPI(api, "POST", "/api/v1/log/entries/retrieve") // add metrics
```
var-sdk, line 148: Thanks for correcting this from GET to POST, we're missing metrics because of this.

var-sdk (approved):

asraa (approved):
Thanks, look good! Thanks for adding in the validation/panic in case someone mis-types the endpoint.

PR #1082, opened 2022-09-28
Search through all shards when searching by hash

priyawadhwa:
This adds in a test with `cosign sign-blob` and `cosign verify-blob` to make sure it is working as expected

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -415,23 +415,27 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
 			searchHashes = append(searchHashes, hash)
 		}
 
-		searchByHashResults := make([]*trillian.GetEntryAndProofResponse, len(searchHashes))
+		searchByHashResults := make([][]*trillian.GetEntryAndProofResponse, len(searchHashes))
 		g, _ = errgroup.WithContext(httpReqCtx)
 		for i, hash := range searchHashes {
 			i, hash := i, hash // https://golang.org/doc/faq#closures_and_goroutines
 			g.Go(func() error {
-				resp := tc.getLeafAndProofByHash(hash)
-				switch resp.status {
-				case codes.OK:
-				case codes.NotFound:
-					code = http.StatusNotFound
-					return resp.err
-				default:
+				var results []*trillian.GetEntryAndProofResponse
+				for _, shard := range api.logRanges.AllShards() {
+					tcs := NewTrillianClientFromTreeID(httpReqCtx, shard)
+					resp := tcs.getLeafAndProofByHash(hash)
+					if resp.status != codes.OK {
+						continue
+					}
+					leafResult := resp.getLeafAndProofResult
+					if leafResult != nil && leafResult.Leaf != nil {
+						results = append(results, resp.getLeafAndProofResult)
+					}
 				}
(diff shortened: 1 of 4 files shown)


Hayden-IO:
Do we need to block on the Cosign fix? It's fairly unlikely that someone has uploaded the same entry cross-shard - It likely only affects the individual who reported the issue, and we can find a workaround (re-sign and reupload).

pkg/api/entries.go
```diff
@@ -415,23 +415,27 @@ func SearchLogQueryHandler(params entries.SearchLogQueryParams) middleware.Respo
...
+					leafResult := resp.getLeafAndProofResult
+					if leafResult != nil && leafResult.Leaf != nil {
+						results = append(results, resp.getLeafAndProofResult)
+					}
 				}
-				leafResult := resp.getLeafAndProofResult
-				if leafResult != nil && leafResult.Leaf != nil {
-					searchByHashResults[i] = leafResult
+				if results == nil {
+					return fmt.Errorf("no responses found")
```
Hayden-IO, line 442: do we need to set response code to `http.StatusNotFound`?

priyawadhwa:
@haydentherapper yep! removed that comment, this should be good to merge once it's green

dlorenc (dismissed):

tests/sharding-e2e-test.sh
```diff
@@ -24,6 +24,8 @@ set -ex
 echo "Installing createtree..."
 go install github.com/google/trillian/cmd/createtree@latest
 
+echo "Installing cosign..."
+go install github.com/sigstore/cosign/cmd/cosign@latest
```
bobcallaway: this logic may evolve as time goes on - are we better actually reimplementing the logic here by directly calling the rekor APIs?

priyawadhwa: i think we probably want to test with cosign somewhere, whether it's here or in a prober. not super opinionated on this one.

asraa (commented):
server changes look fine to me, but CLI logic is incorrect now, for e.g.
https://github.com/sigstore/rekor/blob/204d9afd1ba326284f9c4a91d92358ade513120e/cmd/rekor-cli/app/verify.go#L140

just create a follow-up issue, because this is imp

tests/sharding-e2e-test.sh
asraa: +1 let's remove cosign and use rekor: the problem was reproducible with a SearchLogQuery

tests/sharding-e2e-test.sh
```diff
@@ -251,6 +254,10 @@ ENTRY_ID_2=$($REKOR_CLI get --log-index 3 --rekor_server http://localhost:3000 -
 NUM_ELEMENTS=$(curl -f http://localhost:3000/api/v1/log/entries/retrieve -H "Content-Type: application/json" -H "Accept: application/json" -d "{ \"entryUUIDs\": [\"$ENTRY_ID_1\"]}" | jq '. | length')
 stringsMatch $NUM_ELEMENTS "1"
 
+# Make sure we can verify the blob we entered into the now-inactive shard
```
asraa: you can replace with a call to `./rekor verify --public-key blah --artifact blah --signature blah` with the artifact properties

asraa (approved):

Issue #1075, opened 2022-09-23
Revamp rekor e2e test coverage


bobcallaway:
**Description**

Right now, it can be quite difficult to map the e2e test cases we have along with the code base, given that they're all in a small number of fairly large golang files. I think we could break these apart, help to make them more idomatic and consistent, and help for the development of future types and new updates to the Rekor API going forward.

naveensrinivasan:
I would like to take this. @bobcallaway, Can you please assign it to me?

PR #1071, opened 2022-09-22
Adding e2e test coverage

cdris:
#### Summary

Instruments e2e tests for code coverage. Uploads e2e coverage report to codecov as part of github action.

#### Release Note

NONE

PR #1089, opened 2022-10-03
Revamping rekor e2e - part 1 of N

naveensrinivasan:
#### Summary

Refactoring e2e tests.
There will be multiple PRs to move the e2e to specific folders. Some code is duplicated up until all the e2e code is refactored. 
https://github.com/sigstore/rekor/issues/1075

#### Release Note
NONE

#### Documentation
NONE


bobcallaway:
Does this still generate appropriate code coverage after https://github.com/sigstore/rekor/pull/1071 @cdris ?

pkg/e2e-test.sh
```diff
@@ -0,0 +1,58 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+docker kill $(docker ps -q) || true
+echo "starting services"
+docker-compose up -d
```
cdris: See the changes to the original e2e-test.sh [here](https://github.com/sigstore/rekor/pull/1071/files#diff-80e32b72954ddce7412766a47030a4fec2af262e2bbf718c17ef6dad7c9ef84f).

the docker compose command needs to reference docker-compose.test.yml to run the server with the proper code coverage args

pkg/e2e-test.sh
```diff
@@ -0,0 +1,58 @@
...
+
+echo "building CLI and server"
+# set the path to the root of the repo
+dir=$(git rev-parse --show-toplevel)
+go build -o rekor-cli ./cmd/rekor-cli
+go build -o rekor-server ./cmd/rekor-server
```
cdris: Both build commands should be updated to build with code coverage:

```
go test -c ./cmd/rekor-cli -o rekor-cli -cover -covermode=count -coverpkg=./...
go test -c ./cmd/rekor-server -o rekor-server -covermode=count -coverpkg=./...
```

pkg/e2e-test.sh
```diff
@@ -0,0 +1,58 @@
...
+if ! REKORTMPDIR=$REKORTMPDIR go test -tags=e2e ./pkg/...; then
+   docker-compose logs --no-color > /tmp/docker-compose.log
+   exit 1
+fi
+if docker-compose logs --no-color | grep -q "panic: runtime error:" ; then
+   # if we're here, we found a panic
+   echo "Failing due to panics detected in logs"
+   docker-compose logs --no-color > /tmp/docker-compose.log
+   exit 1
+fi
```
cdris, line 62: See the changes at the end of the original e2e-test.sh [here](https://github.com/sigstore/rekor/pull/1071/files#diff-80e32b72954ddce7412766a47030a4fec2af262e2bbf718c17ef6dad7c9ef84f).

You'll need to add the additional steps to generate the server code coverage report by calling the /kill endpoint, and then you'll need to copy down the server coverage report. Right now the original script is copying it to /tmp/rekor-server.cov. You'll need to name this new coverage file differently. Maybe yours is /tmp/pkg-rekor-server.cov for ease of pattern matching.

Then when you copy the gocovmerge steps, just update the pattern match to /tmp/pkg-rekor-*.cov and the output to /tmp/pkg-rekor-merged.cov

.github/workflows/main.yml
```diff
@@ -84,6 +84,8 @@ jobs:
 
       - name: CLI
         run: ./tests/e2e-test.sh
+      - name: PKG-CLI # this will a WIP to move all the CLI tests to the pkg repo
+        run: ./pkg/e2e-test.sh
```
cdris, line 88: This actually refers to line 98 below, but github won't let me comment on it. Since you'll be generating a different code coverage file, you'll want to add it. If you follow my naming suggestion, that line will become:

`files: /tmp/rekor-merged.cov,/tmp/pkg-rekor-merged.cov`

pkg/util/util.go
```diff
@@ -0,0 +1,187 @@
+//
```
cdris, line 1: A number of [changes](https://github.com/sigstore/rekor/pull/1071/files#diff-05d01822b637dcfaa33f0867a1a327fe69bd9e8038c54b8e0ced6711a121b595) were made to tests/util.go to support code coverage that you'll want to add here.

Mainly you need to update all of the `run` functions to prepend the coverage flag and then strip the coverage info off of the output before returning it. You'll also need to update the coverage flag to use /tmp/pkg-rekor-cli... so that the correct files are merged.

pkg/e2e-test.sh
cdris: You'll probably also want to add the `rm -f /tmp/pkg-rekor-*.cov` line above. Doesn't impact when github runs the tests but makes getting accurate test coverage locally easier.

pkg/e2e-test.sh
```diff
@@ -0,0 +1,58 @@
...
+       echo -n "."
+       sleep 10
+       let 'count+=1'
+    fi
+done
+
+echo
+echo "running tests"
+REKORTMPDIR="$(mktemp -d -t rekor_test.XXXXXX)"
+cp $dir/rekor-cli $REKORTMPDIR/rekor-cli
```
cdris, line 50: Is there a reason you're copying the cli instead of calling it with ../ in util.go?

cdris (commented):
Hello! As Bob mentioned, I recently updated the e2e tests to create code coverage reports in #1071. I've commented what needs to change below to support code coverage. Once util.go and e2e-test.sh are updated, all of the moving of tests going forward should be fine as long as they're using all the same util run functions.

pkg/e2e-test.sh
naveensrinivasan: Go it, Thanks

pkg/e2e-test.sh
naveensrinivasan: Go it, Thanks

pkg/e2e-test.sh
naveensrinivasan: OK, Thanks

.github/workflows/main.yml
naveensrinivasan: Thanks

pkg/util/util.go
naveensrinivasan: I have updated it. Can you please check it and let me know? Thanks

cdris: You copied over the coverageFlag func but aren't actually calling it anywhere. You just need to make sure it's added as the first arg before exec is called.

pkg/util/util.go
```diff
@@ -0,0 +1,204 @@
...
+func rekorServer() string {
+	if s := os.Getenv("REKOR_SERVER"); s != "" {
+		return s
+	}
+	return "http://localhost:3000"
+}
+
+func coverageFlag() string {
+	return "-test.coverprofile=/tmp/rekor-cli." + randomSuffix(8) + ".cov"
+}
```
cdris, line 134: replace this with /tmp/pkg-rekor-cli

pkg/e2e-test.sh
```diff
@@ -0,0 +1,77 @@
...
+
+if ! docker cp $(docker ps -aqf "name=rekor_rekor-server"):go/rekor-server.cov /tmp/rekor-server.cov ; then
+   # failed to copy code coverage report from server
+   echo "Failed to retrieve server code coverage report"
+   docker-compose logs --no-color > /tmp/docker-compose.log
+   exit 1
+fi
+
+# merging coverage reports and filtering out /pkg/generated from final report
+hack/tools/bin/gocovmerge /tmp/rekor-*.cov | grep -v "/pkg/generated/" > /tmp/rekor-merged.cov
```
cdris: should be /tmp/pkg-rekor... same with rekor-merged files below. The naming all has to change for the pkg tests so there isn't overlap with the existing e2e tests coverage files.

pkg/e2e-test.sh
```diff
@@ -0,0 +1,77 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+testdir=$(dirname "$0")
+
+rm -f /tmp/rekor-*.cov
```
cdris: /tmp/pkg-rekor-*.cov

cdris (changes requested):
Just need to update all of the /tmp/rekor references to /tmp/pkg-rekor so the code coverage doesn't overlap with the existing e2e tests. Also need to actually add the coverage flag to args before calling the cli in tests.

pkg/util/util.go
naveensrinivasan: Fixed it.

pkg/e2e-test.sh
naveensrinivasan: Got it. Thanks

pkg/e2e-test.sh
naveensrinivasan: Got it , thanks

pkg/util/util.go
naveensrinivasan: Fixed it. Thanks

pkg/util/util.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		t.Log(string(b))
+		t.Fatal(err)
+	}
+	return stripCoverageOutput(string(b))
+}
+
+func RunCli(t *testing.T, arg ...string) string {
+	t.Helper()
+	arg = append([]string{coverageFlag()}, arg...)
```
cdris: you don't need to prepend coverageFlag here since this calls Run which prepends it

pkg/util/util.go
```diff
@@ -0,0 +1,207 @@
...
+	if os.Getenv("REKORTMPDIR") != "" {
+		arg = append(arg, "--config="+os.Getenv("REKORTMPDIR")+".rekor.yaml")
+	}
+	cmd := exec.Command(cli, arg...)
+	b, err := cmd.CombinedOutput()
+	if err == nil {
+		t.Log(string(b))
+		t.Fatalf("expected error, got %s", string(b))
+	}
+	return string(b)
```
cdris: return stripCoverageOutput(string(b))

cdris (changes requested):

pkg/util/util.go
naveensrinivasan: 👍

pkg/util/util.go
naveensrinivasan: 👍

cdris (dismissed):
lgtm

naveensrinivasan:
> lgtm

Thanks

naveensrinivasan:
@sigstore/rekor-codeowners A friendly reminder!

pkg/e2e-test.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+#
+# Copyright 2021 The Sigstore Authors.
```
bobcallaway:
```suggestion
# Copyright 2022 The Sigstore Authors.
```

pkg/util/util.go
```diff
@@ -0,0 +1,206 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
bobcallaway:
```suggestion
// Copyright 2022 The Sigstore Authors.
```

pkg/e2e-test.sh
```diff
@@ -0,0 +1,77 @@
...
+
+set -e
+testdir=$(dirname "$0")
+
+rm -f /tmp/pkg-rekor-*.cov
+echo "installing gocovmerge"
+make gocovmerge
+docker kill $(docker ps -q) || true
+echo "starting services"
+docker-compose -f docker-compose.yml -f docker-compose.test.yml up -d --force-recreate --build
```
bobcallaway: do we need `--force-recreate` here?

bobcallaway (commented):
one question and two nits

pkg/e2e-test.sh
naveensrinivasan: I reused most of them from here https://github.com/sigstore/rekor/blob/7807d1e5a3fbc9b8fdd17053da8163463e87a5ae/tests/e2e-test.sh#L26

which had the reference. I would suggest in my next PR, I can check if it is required and remove it. Would that work?

bobcallaway: I would vote fix pkg/e2e-test.sh here since you're adding a new file. we can remove from test/e2e-test.sh in another PR.

bobcallaway: thanks, otherwise we rebuild the same container (which can take a while in CI) for no reason.

naveensrinivasan: Fixed it. Thanks

pkg/e2e-test.sh
```diff
@@ -0,0 +1,77 @@
...
+if docker-compose logs --no-color | grep -q "panic: runtime error:" ; then
+   # if we're here, we found a panic
+   echo "Failing due to panics detected in logs"
+   docker-compose logs --no-color > /tmp/docker-compose.log
+   exit 1
+fi
+
+echo "generating code coverage"
+curl -X GET 0.0.0.0:2345/kill
+sleep 5
```
bobcallaway, line 66: Something we need to consider, is that part of the e2e tests have dependencies on specific entries; the Rekor instance is launched without specifying a particular TreeID (for these tests), and therefore as you break this apart you will need to think through which entries a test needs in order to be successful.

bobcallaway: Since we kill the instance to gather code coverage here, the entries are lost unless in the future we change these tests to start Rekor targeted at a specific treeID.

naveensrinivasan: So should we populate the rekor with specific entries before the tests? How is it working for the existing e2e? Is this a missing feature at present?

bobcallaway: At the moment, this is more of an FYI rather than something that's directly actionable. Just something for you to be aware of as you continue moving things over, is that unless we change the test harness to be more shard aware, you may run into issues around tests that are dependent on one another. We should probably make those explicit (with wrapping and use of `t.Run`) whenever we see them.

naveensrinivasan: OK, Thanks. What else is required to get this merged in?

bobcallaway: tests need to pass (I just re-triggered them).

bobcallaway (approved):

PR #1096, opened 2022-10-05
enable configurable client retries with backoff in RekorClient

bobcallaway:
This adds a configurable amount of retries for users of RekorClient, as well as the ability to customize the logger. The underlying library (github.com/hashicorp/go-retryablehttp) implements exponential backoff and respects `Retry-After` response headers.

I'll push up another PR that changes `rekor-cli` to take advantage of these changes.

### DIFF
--- a/pkg/client/rekor_client.go
+++ b/pkg/client/rekor_client.go
@@ -35,9 +35,7 @@ func GetRekorClient(rekorServerURL string, opts ...Option) (*client.Rekor, error
 
 	retryableClient := retryablehttp.NewClient()
 	retryableClient.RetryMax = int(o.RetryCount)
-	if o.Logger != nil {
-		retryableClient.Logger = o.Logger
-	}
+	retryableClient.Logger = o.Logger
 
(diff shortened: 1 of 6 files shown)


asraa (dismissed):
LOVE IT!

priyawadhwa (dismissed):
nice!

priyawadhwa (approved):

pkg/client/rekor_client.go
```diff
@@ -32,7 +33,14 @@ func GetRekorClient(rekorServerURL string, opts ...Option) (*client.Rekor, error
 	}
 	o := makeOptions(opts...)
 
-	rt := httptransport.New(url.Host, client.DefaultBasePath, []string{url.Scheme})
+	retryableClient := retryablehttp.NewClient()
```
asraa, line 36: was user-agent every set from the options?

bobcallaway: It’s set on line 41 based on what the caller passes in.

asraa: OHH thank you! i was double-checking this for a fulcio retry client. thank you!

PR #1097, opened 2022-10-05
add retry command line flag on rekor-cli

bobcallaway:
Depends-on: #1096 

#### Release Note

adds `--retry` option to `rekor-cli` commands to automatically retry calls to a rekor server based on transient errors; default value is to retry 3 times


cpanato (dismissed):
coool

priyawadhwa (approved):

PR #1102, opened 2022-10-06
update swagger API version to 1.0.0

bobcallaway:
@priyawadhwa this should ideally get included before we cut 1.0; it will just make sure the doc reflects the correct major version.


cpanato (approved):
Thos is something we need to remember to bump when we cut a relase

priyawadhwa (approved):

PR #1103, opened 2022-10-06
install gocovmerge in hack/tools

bobcallaway:
Adding this as a managed "tool" in hack/tools so we can ensure it is on any system where we run e2e tests.


cpanato (approved):
Nice. Thanks for that

asraa (approved):

priyawadhwa (approved):

cdris (approved):

PR #1106, opened 2022-10-06
Add some info and debug logging to commonly used funcs

priyawadhwa:
ref https://github.com/sigstore/public-good-instance/issues/387


cpanato (approved):
thanks

PR #1152, opened 2022-10-25
Refactor e2e tests - x509 apk

naveensrinivasan:
https://github.com/sigstore/rekor/issues/1075
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

#### Summary
- Refactored e2e test x509 into the respective folder.
#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,12 @@ ko:
 deploy:
 	LDFLAGS="$(SERVER_LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) ko apply -f config/
 
+e2e:
+	go test -c -tags=e2e ./tests
+	go test -c -tags=e2e ./pkg/pki/x509
+	go test -c -tags=e2e ./pkg/pki/tuf
+	go test -c -tags=e2e ./pkg/types/rekord
+
 sign-container: ko
 	cosign sign --key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) $(KO_DOCKER_REPO)/rekor-server:$(GIT_HASH)
 	cosign sign --key .github/workflows/cosign.key -a GIT_HASH=$(GIT_HASH) $(KO_DOCKER_REPO)/rekor-cli:$(GIT_HASH)
--- a/pkg/pki/x509/e2e.go
+++ b/pkg/pki/x509/e2e.go
@@ -16,7 +16,7 @@
 //go:build e2e
 // +build e2e
 
-package e2e
+package x509
 
 import (
 	"bytes"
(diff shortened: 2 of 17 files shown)


Hayden-IO:
cc @asraa

pkg/pki/x509/e2e.go
```diff
@@ -16,7 +16,7 @@
 //go:build e2e
```
Hayden-IO, line 16: I think these test keys and functions need to live in a separate package (`test`) or not be exported. I'd worry about these getting used otherwise.

Makefile
```diff
@@ -125,6 +125,12 @@ ko:
 deploy:
 	LDFLAGS="$(SERVER_LDFLAGS)" GIT_HASH=$(GIT_HASH) GIT_VERSION=$(GIT_VERSION) ko apply -f config/
 
+e2e:
+	go test -c -tags=e2e ./tests
+	go test -c -tags=e2e ./pkg/pki/x509
+	go test -c -tags=e2e ./pkg/pki/tuf
+	go test -c -tags=e2e ./pkg/types/rekord
```
bobcallaway, line 132: can this be `./pkg/...`?

naveensrinivasan: That was the first thing I tried, and it failed with this.
```
go test -c -tags=e2e ./pkg/...
cannot use -c flag with multiple packages
```

pkg/pki/x509/e2e.go
naveensrinivasan: I am on the fence about this. These are available only when the `build tag e2e` is used to compile. So someone must compile with `e2e` to get these additional exports.

Hayden-IO: Do they need to be tagged? They were previously because it relied on an instance of Rekor running, but now, since these are being split up, I think they can always be run.

Makefile
cdris: Do these need to be compiled into test binaries? Where is this make command actually being used?

naveensrinivasan: For sanity check. It is hard to ensure that the code is compiling without these. Anyone who would have to run e2e to ensure it is compiling.  I am using this manually to check if it is compiling. 

I don't want to add another check to CI as it will slow down things. This is to help anyone contributing to making it easier to test things. That was my intention.

Do you have other thoughts?

pkg/pki/x509/e2e.go
naveensrinivasan: They still need a `rekor` server running.

Hayden-IO: Oh yep, sorry, I see that now.

Still, I don't love exporting static key values. A better fix may be to generate these in memory, but this should be a follow up.

naveensrinivasan: >Still, I don't love exporting static key values. A better fix may be to generate these in memory, but this should be a follow up.

I agree. It can be a follow-up PR. Thanks

naveensrinivasan:
@bobcallaway A friendly ping.

bobcallaway (approved):

PR #1163, opened 2022-11-02
Add script to backfill Redis from Rekor

Hayden-IO:

### DIFF
--- a/cmd/backfill-redis/main.go
+++ b/cmd/backfill-redis/main.go
@@ -12,6 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/*
+	backfill-redis is a script to populate the Redis index with entries
+	from Rekor. This is sometimes necessary because Redis caching is best
+	effort. If Redis returns an error, Rekor will not, and so sometimes
+	we need to backfill missing entries into Redis for the search API.
+
+	To run:
+	go run cmd/backfill-redis/main.go --rekor-address <address> \
+	    --hostname <redis-hostname> --port <redis-port>
+		--start <first index to backfill> --end <last index to backfill>
+*/
+
 package main
 
 import (
@@ -44,17 +56,17 @@ import (
 )
 
 var (
-	redisAddress = flag.String("address", "", "Address for Redis application")
-	redisPort    = flag.String("port", "", "Port to Redis application")
-	startIndex   = flag.Int("start", -1, "First index to backfill")
-	endIndex     = flag.Int("end", -1, "Last index to backfill")
-	rekorAddress = flag.String("rekor-address", "", "Address for Rekor, e.g. https://rekor.sigstore.dev")
+	redisHostname = flag.String("hostname", "", "Hostname for Redis application")
+	redisPort     = flag.String("port", "", "Port to Redis application")
+	startIndex    = flag.Int("start", -1, "First index to backfill")
+	endIndex      = flag.Int("end", -1, "Last index to backfill")
+	rekorAddress  = flag.String("rekor-address", "", "Address for Rekor, e.g. https://rekor.sigstore.dev")
 )
 
 func main() {
 	flag.Parse()
 
-	if *redisAddress == "" {
+	if *redisHostname == "" {
 		log.Fatal("address must be set")
 	}
 	if *redisPort == "" {
@@ -71,7 +83,7 @@ func main() {
 	}
 
 	cfg := radix.PoolConfig{}
-	redisClient, err := cfg.New(context.Background(), "tcp", fmt.Sprintf("%s:%s", *redisAddress, *redisPort))
+	redisClient, err := cfg.New(context.Background(), "tcp", fmt.Sprintf("%s:%s", *redisHostname, *redisPort))
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -120,7 +132,7 @@ func main() {
 }
 
 // unmarshalEntryImpl decodes the base64-encoded entry to a specific entry type (types.EntryImpl).
-// from cosign
+// Taken from Cosign
 func unmarshalEntryImpl(e string) (types.EntryImpl, string, string, error) {
 	b, err := base64.StdEncoding.DecodeString(e)
 	if err != nil {
--- a/cmd/backfill-redis/main.go
+++ b/cmd/backfill-redis/main.go
@@ -1,0 +1,120 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


Hayden-IO:
I haven't tested this yet! Just wanted to get this out there for review

cmd/backfill-redis/main.go
```diff
@@ -0,0 +1,120 @@
...
+		log.Fatal("end must be set to >=0")
+	}
+
+	cfg := radix.PoolConfig{}
+	redisClient, err := cfg.New(context.Background(), "tcp", fmt.Sprintf("%s:%s", *redisAddress, *redisPort))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	rekorClient, err := client.GetRekorClient("https://rekor.sigstore.dev")
```
vaikas: since redisAddress is configurable, maybe this should be too?

cmd/backfill-redis/main.go
```diff
@@ -0,0 +1,120 @@
...
+		params.SetLogIndex(int64(*startIndex))
+		resp, err := rekorClient.Entries.GetLogEntryByIndex(params)
+		if err != nil {
+			log.Fatalf("retrieving log uuid by index: %v", err)
+		}
+		for uuid, entry := range resp.Payload {
+			// uuid is the global UUID - tree ID and entry UUID
+			e, _, _, err := unmarshalEntryImpl(entry.Body.(string))
+			if err != nil {
+				fmt.Printf("error unmarshalling entry for %s\n", uuid)
```
vaikas: maybe print the error here, L84 and L89

vaikas (commented):
Couple of small nits.

Hayden-IO:
To test:

```
docker-compose build; docker-compose up

# Insert some entries

docker run -it --network host --rm redis redis-cli
> KEYS *
# See key values
> lrange "<key>" 0 -1
FLUSHDB # Delete all keys
```

Example output, after inserting two entries:

```
$ go run cmd/backfill-redis/main.go --rekor-address http://localhost:3000 --address localhost --port 6379 --start 0 --end 1
Uploaded Redis entry 62e8aaf898e2779de217cfcb1152ee243a1428f38fb4866b77f36b781a2793332d5c4ecd9b3a9856, index 0, key 013dc006537df9c906667dbc1cb17601e6dc2720e5fd3f4782570db299957783
Uploaded Redis entry 62e8aaf898e2779de217cfcb1152ee243a1428f38fb4866b77f36b781a2793332d5c4ecd9b3a9856, index 0, key <service-account>
Uploaded Redis entry 62e8aaf898e2779de217cfcb1152ee243a1428f38fb4866b77f36b781a2793332d5c4ecd9b3a9856, index 0, key sha256:b678e836ee3ab5633a44361be6ead2fe856d3f7846915bb06c917f49fc82dbde
Completed log index 0
Uploaded Redis entry 62e8aaf898e2779dfd8c21db39ab530120100e05810cc4656d1964a9e95f768e193d1e31b624b6e0, index 1, key ea8e97a7a1f9236f2b4568d9d4e03e7c8e415e9f76db33b52fb8d07ca5458808
Uploaded Redis entry 62e8aaf898e2779dfd8c21db39ab530120100e05810cc4656d1964a9e95f768e193d1e31b624b6e0, index 1, key <service-account>
Uploaded Redis entry 62e8aaf898e2779dfd8c21db39ab530120100e05810cc4656d1964a9e95f768e193d1e31b624b6e0, index 1, key sha256:2460fddc38e527491fed7a9a127d9f4d75172a957ff1071668920f7ebb167629
Completed log index 1
```

cmd/backfill-redis/main.go
```diff
@@ -0,0 +1,144 @@
...
+
+	entry, err := types.UnmarshalEntry(pe)
+	if err != nil {
+		return nil, "", "", err
+	}
+	return entry, pe.Kind(), entry.APIVersion(), nil
+}
+
+func addToIndex(ctx context.Context, redisClient radix.Client, key, value string) error {
+	return redisClient.Do(ctx, radix.Cmd(nil, "LPUSH", key, value))
```
bobcallaway, line 168: a potential optimization:  `LPUSH key a b c` is equivalent to `LPUSH key c`, `LPUSH key b`, and `LPUSH key a` per https://redis.io/commands/lpush/

cmd/backfill-redis/main.go
```diff
@@ -0,0 +1,144 @@
...
+					fmt.Printf("error inserting UUID %s with key %s: %v\n", uuid, key, err)
+				}
+				fmt.Printf("Uploaded Redis entry %s, index %d, key %s\n", uuid, i, key)
+			}
+		}
+		if success {
+			fmt.Printf("Completed log index %d\n", i)
+		} else {
+			fmt.Printf("Errors with log index %d\n", i)
+		}
```
bobcallaway: nit: just log the index along with the error above instead of two separate lines?

bobcallaway (dismissed):

cmd/backfill-redis/main.go
Hayden-IO: I think we'd need the inverse, because for one entry, we have many keys mapping to one value, rather than one key mapping to many values.

cmd/backfill-redis/main.go
Hayden-IO: Cleaned up the error printing, now prints all errors per index

cmd/backfill-redis/main.go
Hayden-IO: Actually, just realized, if we have the wrong range and we accidentally backfill an existing index, then it'll push a second duplicate UUID into the list. Not sure if this causes any issues.

To be safe, I added a remove before the push.

Hayden-IO: Do you know why we are using lists and not sets?

bobcallaway (approved):

cmd/backfill-redis/main.go
vaikas: I was wondering about batch mode as well, but wasn't sure how the errors are reported / handled for if things fail.

Hayden-IO: I couldn't find any ways to batch this using the available commands. Performance shouldn't be an issue either for a few thousands entries.

vaikas: So for our test cases in staging then, I reckon we'll do something like this?

Say we have existing indices, [0,12] (these are obvs not right, but just for illustrative purposes)

Delete 8-10

Run the script from 5-12

And make sure things work still?

Hayden-IO: I'd narrow down the window of extra indices. Let's say we delete 8-10, and run it on 7-11. Before, we should check all key-value pairs for 7 and 11, and make sure that they remain the same after running

PR #1175, opened 2022-11-04
trim whitespace around public keys before parsing

### LINKED ISSUE #1137: Trailing whitespace in pub key results in HTTP 400
**Description**

When configuring my secrets I accidentially added an additional new line in my cosign public key and leads to errors when interacting with Rekor.

This seems to be server side. I tried to debug the client, but in case of providing a URL (which suffers from same error) the URL seems to be passed on to Rekor and key is fetched by server? Is that correct?

Would it make sense to have the implementation strip whitespace as necessary?

Here is a minimal working (breaking ;) ) example:

```
$ COSIGN_PASSWORD="" cosign generate-key-pair
Private key written to cosign.key
Public key written to cosign.pub
$ rekor-cli search --public-key cosign.pub --pki-format x509
no matching entries found
$ echo "" >> cosign.pub
$ rekor-cli search --public-key cosign.pub --pki-format x509
[POST /api/v1/index/retrieve][400] searchIndexBadRequest  &{Code:400 Message:Public key provided could not be parsed}
```

**Version**

```
rekor-cli version
  ____    _____   _  __   ___    ____             ____   _       ___
 |  _ \  | ____| | |/ /  / _ \  |  _ \           / ___| | |     |_ _|
 | |_) | |  _|   | ' /  | | | | | |_) |  _____  | |     | |      | |
 |  _ <  | |___  | . \  | |_| | |  _ <  |_____| | |___  | |___   | |
 |_| \_\ |_____| |_|\_\  \___/  |_| \_\          \____| |_____| |___|
rekor-cli: Rekor CLI

GitVersion:    v1.0.0
GitCommit:     7215f5c4782deef0b9c249d39ab6b9bc70d58a94
GitTreeState:  clean
BuildDate:     2022-10-17T17:35:23Z
GoVersion:     go1.19.2
Compiler:      gc
Platform:      linux/amd64
```

bobcallaway:
If there is trailing whitespace within a public key file, Rekor's parser was "tricked" into thinking that there was a series of concatenated certificates instead of a single key or certificate. 

This trims the leading and trailing whitespace for X509 public key/cert objects before parsing to ensure the logic holds correctly.

Fixes: #1137 

### DIFF
--- a/pkg/pki/x509/x509_test.go
+++ b/pkg/pki/x509/x509_test.go
@@ -79,6 +79,13 @@ const ed25519Pub = `-----BEGIN PUBLIC KEY-----
 MCowBQYDK2VwAyEAizWek2gKgMM+bad4rVJ5nc9NsbNOba0A0BNfzOgklRs=
 -----END PUBLIC KEY-----`
 
+const pubWithTrailingNewLine = `-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEx+ikqUxXurlxZltajRBV2ju31j32
+baT2ax2dXBcpInWaFESqGF35KISflP1EmMvEnfG+AzHecQ0WQp5QzNId+w==
+-----END PUBLIC KEY-----
+
+`
+
 func signData(t *testing.T, b []byte, pkey string) []byte {
 
 	priv, err := cryptoutils.UnmarshalPEMToPrivateKey([]byte(pkey), cryptoutils.SkipPassword)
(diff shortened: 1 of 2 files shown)


pkg/pki/x509/x509_test.go
```diff
@@ -303,4 +310,9 @@ func TestPublicKeyWithCertChain(t *testing.T) {
 	if err == nil || !strings.Contains(err.Error(), "too many certificates specified in PEM block") {
 		t.Fatalf("expected error with long certificate chain, got %v", err)
 	}
+
+	// Verify public key with trailing newline is parsed OK
+	if _, err = NewPublicKey(strings.NewReader(pubWithTrailingNewLine)); err != nil {
```
asraa: sorry for the extra ask for the test: but could you test this plus the same pub without the trailing new line both return the same canonicalvalue?

asraa (commented):
otherwise lgtm

asraa (approved):

Hayden-IO (approved):

PR #1200, opened 2022-11-10
types/alpine: add fuzzer

AdamKorcz:

#### Summary

Adds a fuzz test for the unmarshalling of the Apk type.

### DIFF
--- a/pkg/types/alpine/fuzz_test.go
+++ b/pkg/types/alpine/fuzz_test.go
@@ -1,0 +1,35 @@
+//
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


pkg/types/alpine/fuzz_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+func FuzzPackageUnmarshal(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte) {
+		ff := fuzz.NewConsumer(data)
+		tarBytes, err := ff.TarBytes()
+		if err != nil {
+			return
+		}
+		p := &Package{}
+		p.Unmarshal(bytes.NewReader(tarBytes))
```
bobcallaway, line 33: Unmarshal assumes a concatenation of 3 tar.gz files with some headers removed... so is just `TarBytes()` the correct input to be passing?

https://github.com/sigstore/rekor/blob/21e5b63e7d73cdfa3e9e6103e66173437240240b/pkg/types/alpine/apk.go#L107

AdamKorcz: Good point. I would say it is good enough to get started since the fuzzer does not do any assertions to assess whether the function behaves correctly. I tested this locally, and it does get into the file readings of the tar archives which is great.

Wrt to leaving out the footers, I will add that as an option in `go-fuzz-headers`. 

I plan to iterate over the fuzzer based on the feedback from OSS-Fuzz. If we get a lot of false crashes, then we can modify.

bobcallaway (approved):

PR #1210, opened 2022-11-18
Add Verifier to get public key/cert and identities for entry type

### LINKED ISSUE #1173: Add public key method to EntryImpl
**Description**

This would be useful for getting an entry's public key or certificate without having to first cast the entry body to a specific type.

Hayden-IO:
Looking for feedback before adding tests

Fixes https://github.com/sigstore/rekor/issues/1173


Hayden-IO:
cc @asraa @priyawadhwa - Looking for some feedback before adding more tests! The motivation for this work is to make it easy for users to access the public key or certificate of the entry type - Particularly useful for monitoring.

The only thing I don't love is exposing pgp in the function, not sure a way around that though.

For RFC3161, we can't implement support for public keys. For TUF, I chose to not implement support for now at least, because it's not clear how best to return a set of public keys - Just from the root? From targets too? For minisign, I added a TODO.

Hayden-IO:
One thing I’ll change is setting the key in unmarshal. I’ll just read the key field in in Verifier. More usable that way

pkg/types/cose/cose_test.go
```diff
@@ -17,10 +17,13 @@ package cose
 
 import (
 	"context"
+	"crypto"
+	"crypto/x509"
 	"errors"
 	"testing"
 
 	"github.com/go-openapi/swag"
+	"golang.org/x/crypto/openpgp"
```
lukehinds: might need to use the protonmail dep instead.

pkg/types/entries.go
```diff
@@ -36,6 +39,7 @@ type EntryImpl interface {
 	Canonicalize(ctx context.Context) ([]byte, error) // marshal the canonical entry to be put into the tlog
 	Unmarshal(e models.ProposedEntry) error           // unmarshal the abstract entry into the specific struct for this versioned type
 	CreateFromArtifactProperties(context.Context, ArtifactProperties) (models.ProposedEntry, error)
+	Verifier() (*x509.Certificate, crypto.PublicKey, openpgp.EntityList, error)
```
asraa: I would recommend sending back a single type for this, and it would simplify what the interface for `PublicKey` would be. Could you return canonicalized PEM bytes for each of these types? 

Then verifiers/monitors could grab those bytes and unmarshal it themselves, which is a downside, but it makes this interface a LOT nicer

asraa: I guess the thing was that you would have already had the canonicalized PEM bytes.

Instead, return a `signature.Verifier` then, because that'll directly be useful for you and be a single comomn interface for verification and a return time here.

pkg/types/rekord/v0.0.1/entry.go
```diff
@@ -418,3 +424,45 @@ func (v V001Entry) CreateFromArtifactProperties(ctx context.Context, props types
 
 	return &returnVal, nil
 }
+
+func (v V001Entry) Verifier() (*cx509.Certificate, crypto.PublicKey, openpgp.EntityList, error) {
+	switch f := *v.RekordObj.Signature.Format; f {
```
asraa, line 431: can we grab the `v.keyObj.Verifier` if you implement that common underlying PublicKey interface method?

pkg/types/cose/cose_test.go
Hayden-IO: No longer exported

pkg/types/entries.go
Hayden-IO: Updated based on conversation

Hayden-IO:
Still needs tests, but this looks a lot better I think @asraa 

Main changes:
* EntryImpl needs a Verifier() that returns a pki.PublicKey
* PublicKey now requires an Identities() that returns a list of subjects from either pgp keys, ssh keys, or certs, and for keys, returns either the PEM encoding (pkcs7/tuf/x509) or a specific encoding (minisign/pgp)

pkg/pki/pkcs7/pkcs7.go
```diff
@@ -213,3 +215,27 @@ func (k PublicKey) EmailAddresses() []string {
...
+
+	}
+	if len(k.certs) > 0 {
+		cert := k.certs[0]
+		pem, err := cryptoutils.MarshalPublicKeyToPEM(cert.PublicKey)
+		if err != nil {
+			return nil, err
+		}
+		identities = append(identities, string(pem))
+		identities = append(identities, px509.GetSubjectAlternateNames(cert)...)
```
asraa: could you comment here that this gets `EmailAddresses` and other SANs? Didn't realize this at the start

Is it worth using this method for `Subjects()` above (that's used to populate the redis index)

pkg/pki/ssh/ssh.go
```diff
@@ -106,3 +107,11 @@ func (k PublicKey) EmailAddresses() []string {
 func (k PublicKey) Subjects() []string {
 	return nil
 }
+
+// Identities implements the pki.PublicKey interface
+func (k PublicKey) Identities() ([]string, error) {
+	// an authorized key format with the comment
+	authorizedKey := string(ssh.MarshalAuthorizedKey(k.key)) + k.comment
+	// return the authorized key and comment (assuming it represents an identity)
+	return []string{authorizedKey, k.comment}, nil
```
asraa: what is k.comment is the empty string? is that possible?

pkg/pki/tuf/tuf.go
```diff
@@ -170,3 +175,20 @@ func (k PublicKey) EmailAddresses() []string {
 func (k PublicKey) Subjects() []string {
 	return nil
 }
+
+// Identities implements the pki.PublicKey interface
+func (k PublicKey) Identities() ([]string, error) {
```
asraa, line 175: I think this may be better off as the entire root.Keys struct (including the keytype/scheme) since that's a pretty common format used in secure-systems-lib. so long as you are comparing the struct via DeepEquals in the monitoring side.

pkg/types/cose/v0.0.1/entry.go
```diff
@@ -347,3 +347,7 @@ func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.A
 
 	return &returnVal, nil
 }
+
+func (v V001Entry) Verifier() (pki.PublicKey, error) {
+	return x509.NewPublicKey(bytes.NewReader(*v.CoseObj.PublicKey))
```
asraa, line 355: why not return `v.keyObj` for this and others?

asraa (commented):
Nice change! Some quick first comments

pkg/pki/ssh/ssh.go
bobcallaway: are we guaranteed that `k.comment` is an identity?

pkg/pki/pkcs7/pkcs7.go
Hayden-IO: Changed for Subjects(), and added comment.

pkg/pki/ssh/ssh.go
Hayden-IO: Adding a check that the comment looks like an email before returning it. 

There's no guarantee, but it's very likely.

pkg/pki/tuf/tuf.go
Hayden-IO: I've changed this to return the marshalled root.Keys map, lemme know if that sounds good

pkg/types/cose/v0.0.1/entry.go
Hayden-IO: I was concerned that there may be cases where they aren't set, because the use of `keyObj` was not uniform for all types.

pkg/pki/pkcs7/pkcs7.go
Hayden-IO: These actually aren't equivalent currently, it looks like EmailAddresses is searching the cert's subject rather than SAN. Do you know what's expected for a pkcs7 certificate? The example test cert includes an email in the subject rather than SAN.

Hayden-IO: To be safe, maybe search the SAN and subject?

Hayden-IO:
@asraa @bobcallaway It's ready for another review. I've added tests for all types.

There's one potential bug I've found that I'm filing issues for:
* For PGP, CanonicalValue re-encodes the PGP key such that the PEM encoding does not match the input 
* ~For JARs, the test JAR fails the test (that's commented out) because it says that a CERTIFICATE block, rather than a PKCS7 block, was found. I'm not sure if the JAR is invalid or if we need to allow different certificate formats. I also think there might be a bug in that we pass both the same pkcs7 signature to the NewPublicKey and NewSignature functions, unless this is WAI? (Because the PEM decoder will drop everything before the first PEM block it finds)~ EDIT: That was my error

pkg/pki/pkcs7/pkcs7.go
Hayden-IO: I search both now.

Hayden-IO:
@asraa @bobcallaway Bumping for review

pkg/pki/pkcs7/pkcs7.go
```diff
@@ -211,5 +213,31 @@ func (k PublicKey) EmailAddresses() []string {
 
 // Subjects implements the pki.PublicKey interface
 func (k PublicKey) Subjects() []string {
-	return k.EmailAddresses()
+	// combine identities in the subject and SANs
+	identities := k.EmailAddresses()
+	cert, err := x509.ParseCertificate(k.rawCert)
```
asraa: why not take `k.certs[0]`? Seems like that's the parsed k.rawCert: https://github.com/sigstore/rekor/blob/16915805c250dfc27108276d6598b14b177154e7/pkg/pki/pkcs7/pkcs7.go#L172

pkg/pki/pkcs7/pkcs7.go
```diff
@@ -211,5 +213,31 @@ func (k PublicKey) EmailAddresses() []string {
...
+	return identities
+}
+
+// Identities implements the pki.PublicKey interface
+func (k PublicKey) Identities() ([]string, error) {
+	// pkcs7 structure may contain both a key and certificate chain
+	var identities []string
+	if len(k.certs) > 0 {
+		cert := k.certs[0]
+		pem, err := cryptoutils.MarshalPublicKeyToPEM(cert.PublicKey)
```
asraa: for this you can take `k.key` too, which is `k.certs[0].PublicKey`, so I'm not sure you need this if block.

(get k.key marshalled PEM, and then add k.Subjects()`)

asraa: Do identity-checkers check for the PEM public key inside the cert? Or the PEM encoded cert?

pkg/pki/x509/sans.go
```diff
@@ -0,0 +1,39 @@
...
+
+package x509
+
+import (
+	"crypto/x509"
+
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+func GetSubjectAlternateNames(cert *x509.Certificate) []string {
```
asraa: not blocking - out of curiousity is this needed in cosign or sigstore/sigstore?

pkg/pki/x509/x509.go
```diff
@@ -213,10 +213,42 @@ func (k PublicKey) Subjects() []string {
...
+
+	var cert *x509.Certificate
+	if k.cert != nil {
+		cert = k.cert.c
+	} else if len(k.certs) > 0 {
+		cert = k.certs[0]
+	}
+
+	var identities []string
+	pem, err := cryptoutils.MarshalPublicKeyToPEM(cert.PublicKey)
```
asraa: same comment here: when people want a cert identity wouldn't they want to check the PEM encoded cert (not the extract pubkey)?

asraa (commented):
Thanks for revamping this! It was an easy review!

pkg/pki/pkcs7/pkcs7.go
Hayden-IO: done

pkg/pki/x509/sans.go
Hayden-IO: This came from Cosign, lemme add a TODO to use a shared version out of sigstore/sigstore

Hayden-IO: oh whoops, forgot i did move it! wil use that

pkg/pki/x509/x509.go
Hayden-IO: Absolutely, I think what I was thinking when I implemented this was that we would extract relevant info from the certificate. but we might as well return the certificate along with parsed values, give the caller as much info as possible. 

The other option would be to return only the certificate, but I want to continue to extract useful values in case the caller only cares about a key value or SAN. Does this sound good?

pkg/pki/pkcs7/pkcs7.go
Hayden-IO: added the PEM encoded certificate in addition to the key

Hayden-IO: cleaned up!

Hayden-IO:
@asraa ready for another review!

pkg/pki/x509/x509.go
asraa: SG! For ephemeral certs it's weird to think of pinning either the key value or the cert value - but things might change if people start using fulcio more with their own long lived keys (?).

pkg/pki/x509/x509.go
```diff
@@ -213,10 +213,47 @@ func (k PublicKey) Subjects() []string {
...
+		if err != nil {
+			return nil, err
+		}
+		return []string{string(pem)}, nil
+	}
+
+	var cert *x509.Certificate
+	if k.cert != nil {
+		cert = k.cert.c
+	} else if len(k.certs) > 0 {
```
asraa: do you want to fail after here if `cert` is still a nil pointer?

asraa (commented):
Just one final nit!

pkg/pki/x509/x509.go
Hayden-IO: Added!

asraa (approved):

PR #1253, opened 2022-12-21
Fuzzing: Add more bug detectors

AdamKorcz:

#### Summary

Adds more bug detectors to the Rekor fuzzers. Currently 4 bug detectors come with the "instrumentation" project. It is still early stage and is being added to OSS-Fuzz projects one by one. Containerd, Docker, Istio, Golang are other projects that this is integrated with. 

The "instrumentation" project adds a series of checks in the target code. It analyzes the project, modifies its AST and overwrites the source files with the updated AST. Since this is all done inside the OSS-Fuzz build container, the changes to the source tree are only applied for fuzz testing.

### DIFF
--- a/tests/oss_fuzz.sh
+++ b/tests/oss_fuzz.sh
@@ -14,6 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+cd $SRC
+git clone --depth=1 https://github.com/AdamKorcz/instrumentation
+cd instrumentation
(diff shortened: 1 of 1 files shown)


naveensrinivasan (commented):
What does the bug detector do? Could you please add some information as to what it does?

AdamKorcz:
> What does the bug detector do? Could you please add some information as to what it does?

Added to the PR message.

bobcallaway (approved):

PR #1258, opened 2022-12-29
types: refactor multiple fuzzers

AdamKorcz:

#### Summary

Adds a fuzzer for intoto v0.0.2

Does the following changes to multiple types fuzzers:

1. Silences the logger.
2. Creates the artifact file if the artifact bytes are `nil` for the properties.
3. Hardcodes the version to avoid unnecessary time in the semver dependency.
4. Moves the fuzzers to their v0.0.1 directory to initiate the relevant entry in the VersionMap. For example here:
https://github.com/sigstore/rekor/blob/26f44cd4ba51119da6b9557edcadabe3a97ae2af/pkg/types/helm/v0.0.1/entry.go#L46-L50

### DIFF
--- a/pkg/fuzz/fuzz_utils.go
+++ b/pkg/fuzz/fuzz_utils.go
@@ -1,0 +1,74 @@
+//
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fuzz
+
+import (
+	"net/url"
+	"os"
+	"path/filepath"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+
+	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/types"
+)
+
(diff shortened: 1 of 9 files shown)


bobcallaway (dismissed):

pkg/types/cose/fuzz_test.go
```diff
@@ -17,13 +17,18 @@ package cose
...
 
+	fuzzUtils "github.com/sigstore/rekor/pkg/fuzz"
 	"github.com/sigstore/rekor/pkg/types"
 )
 
+var initter sync.Once
+
 func FuzzCreateProposedEntry(f *testing.F) {
 	f.Fuzz(func(t *testing.T, version string) {
+		initter.Do(fuzzUtils.SetFuzzLogger)
```
asraa, line 31: should this be moved to cose v0.0.1 and populate the props with `CreateProps` like other types?

pkg/fuzz/fuzz_utils.go
```diff
@@ -0,0 +1,74 @@
...
+
+		artifactPath, err := filepath.Abs("ArtifactFile")
+		if err != nil {
+			return props, nil, err
+		}
+		artifactURL, err := url.Parse(artifactPath)
+		if err != nil {
+			return props, nil, err
+		}
+		props.ArtifactPath = artifactURL
```
asraa, line 55: In most cases, if `ArtifactBytes` are written, then we won't even look at the `ArtifactPath` (https://github.com/sigstore/rekor/blob/8fbe9e4a5424c8e8ea2fe668848526202fa46d87/pkg/types/rekord/v0.0.1/entry.go#L343-L345) -- maybe instead you can take some input from `ff` to either use `ArtifactBytes` directly or use it from `ArtifactPath` and clear `ArtifactBytes`?

The same goes for `SignatureBytes` and `SignaturePath`/`PublicKeyBytes` and `PublicKeyPaths`

asraa (approved):
Thanks! Feel free to address the review in a follow-up, they're minor.

pkg/fuzz/fuzz_utils.go
asraa: I misread this! If the ArtifactBytes are nil -- lgtm! (Just see about the Signature/PublicKey if there are any issues)

Issue #1164, opened 2022-11-02
502 error when posting intoto 0.0.2 entry w/o hash


bdehamer:
**Description**

If I try to upload an intoto/0.0.2 entry to Rekor without including a value for the `hash` field, I receive a 502 status code in response:

```
curl -X "POST" "https://rekor.sigstore.dev/api/v1/log/entries" \
     -H 'Accept: application/json' \
     -H 'Content-Type: application/json' \
     -d $'{
  "kind": "intoto",
  "apiVersion": "0.0.2",
  "spec": {
    "content": {
      "envelope": {
        "payload": "YUdWc2JHOHNJSGR2Y214a0lRPT0=",
        "payloadType": "text/plain",
        "signatures": [
          {
            "sig": "TUVRQ0lHZ3hXNWgxUmZoNklwcFcvdm9KV2xNMUNVdStsRzBWdGoxN21XM2pta1BwQWlBdEx2aTFMekxyc2xsSnd3YnJYMEppR0w2MjA0ZURmNUpad2lTaFF4Mk9vdz09",
[...]

Issue #1139, opened 2022-10-21
Signatures in DSSE envelopes for intoto/v0.0.2 are double encoded


kommendorkapten:
**Description**

When uploading an `intoto` object with version `0.0.2`, the retrieved entry has the signature double base64 encoded. 

An example entry with this: 

```
Created entry at index 5547337, available at: https://rekor.sigstore.dev/api/v1/log/entries/24296fb24b8ad77a5f9c879a81d619af36c272b25889758a7855a31c7e4b27d51d1b0b943d2aabc2
```

When retrieving the entry:

```json
...
    "content": {
      "envelope": {
        "payloadType": "application/vnd.in-toto+json",
        "signatures": [
          {
            "publicKey": "LS0...",
            "sig": "TUVVQ0lDZWNWcS9lckI3MU4xRm5PYVpnV0l1OHBnb1h4QXM2N212NUpXd0JIR05KQWlFQWxwRjY4N0lHNnF5L0NOTmYwZmE4OFdhWVR2RVJsOEhjeXhoVFgzQWZYTVk9"
          }
...
```

After base64 decode the signature, we get the real signature:

```shell
$ printf TUVVQ0lDZWNWcS9lckI3MU4xRm5PYVpnV0l1OHBnb1h4QXM2N212NUpXd0JIR05KQWlFQWxwRjY4N0lHNnF5L0NOTmYwZmE4OFdhWVR2RVJsOEhjeXhoVFgzQWZYTVk9 | base64 -D
MEUCICecVq/erB71N1FnOaZgWIu8pgoXxAs67mv5JWwBHGNJAiEAlpF687IG6qy/CNNf0fa88WaYTvERl8HcyxhTX3AfXMY=
```

which was uploaded:

```shell
$ grep 'sig"' envelope
                "sig": "MEUCICecVq/erB71N1FnOaZgWIu8pgoXxAs67mv5JWwBHGNJAiEAlpF687IG6qy/CNNf0fa88WaYTvERl8HcyxhTX3AfXMY="
```

**Version**

Rekor as deployed of 2022-10-21.

kommendorkapten:
I understand this is a known issue: https://sigstore.slack.com/archives/C01CX4E2K70/p1664483571881309, should I close this?

bobcallaway:
which client is being used to upload?

kommendorkapten:
I was using a local built `rekor-cli`, at the current `main` branch a few days ago. 
[...]

PR #1279, opened 2023-01-13
WIP: add new version of intoto type

bobcallaway:
- [ ] move tests out of e2e dir and into pkg-specific version
- [ ] add test coverage

#### Summary

This adds `v0.0.3` of the intoto type, aiming to address a few issues in v0.0.2:

- https://github.com/sigstore/rekor/issues/1139
- https://github.com/sigstore/rekor/issues/1164
- ensure that index keys and digests are computed only off of validated inputs
- more cleanly separating the schema between what should be proposed VS what is returned from the log

#### Release Note

Adds new default version of intoto type

FYI @pxp928 @bdehamer


asraa:
sorry, quick comment before I dive in: wdyt about renaming to a dsse type? i was talking with @laurentsimon and had a huge amount of confusion myself about what content/payloadHash referred to, and if it's a dsse type it could be better named as predicate/envelopeHash  

I think this came up in an issue before but I can't find a reference.

bobcallaway:
> sorry, quick comment before I dive in: wdyt about renaming to a dsse type? i was talking with @laurentsimon and had a huge amount of confusion myself about what content/payloadHash referred to, and if it's a dsse type it could be better named as predicate/envelopeHash
> 
> I think this came up in an issue before but I can't find a reference.

In https://github.com/sigstore/rekor/blob/224afcab4145900192f3f5354f7e0e1cf96a3c1b/pkg/types/intoto/v0.0.3/intoto_v0_0_3_schema.json#L12 I tried to make this clear that this would only accept DSSE envelopes that are signed intoto attestations.

I would presume that a more generically named `dsse` type could refer to any payload, but this type has been specific to intoto attestations?

kommendorkapten:
My two cents:

I think it would be a bit cleaner to have the type referring to the outer envelope, e.g. `dsse`. Then after the envelope is parsed, we can extract certain fields based on a set of payload types Rekor recognizes, such as in-toto statements (and the predicate type for build provenances like it is today). This is e.g. how `cose` does it: https://github.com/sigstore/rekor/blob/fd8e054af754dcb9e8861c0a9189669ee443e2d7/pkg/types/cose/v0.0.1/entry.go#L115.

The risk with this is of course that there can be a combinatorial explosion of "subtypes" in the `dsse` type, but with proper structure of the code I think it should be fine.
I guess with this way, it would be easier for a consumer to use. 

To start with we may only recognize `in-toto` statements, but any `dsse` enevelope could be uploaded.

pkg/types/intoto/v0.0.3/entry.go
```diff
@@ -0,0 +1,460 @@
...
+	}
+
+	// either we have just proposed content or the canonicalized fields
+	if intotoObj.ProposedContent == nil {
+		// then we need canonicalized fields, and all must be present (if present, they would have been validated in the above call to Validate())
+		if intotoObj.EnvelopeHash == nil || intotoObj.PayloadHash == nil || len(intotoObj.Signatures) == 0 {
+			return errors.New("either proposed content or envelope hash, payload hash, and signatures must be present")
+		}
+		v.IntotoObj = *intotoObj
+		return nil
```
asraa, line 189: Because this is incoming proposed content with canonicalized fields, shouldn't `verifyEnvelope` still apply here for robustness?

What if a user proposed a `ProposedEntry` with pre-canonicalized fields, let's say, because it wants to hide the `Envelope` content? It should still be verifiable, assuming verification over a hash.

But if it's clear that users MUST be publishing full envelope content to the log, fine with me.

asraa: Only hesitation here is that I am wondering if it's possible for a client to be setting the `EnvelopeHash`, `PayloadHash`, and `Signatures` of a bogus signed envelope and bypassing validation (not through the CLI, but through setting a proposed entry in the createentry handler)

asraa (commented):
Just some very minor comments, but overall this is so nicely organized, multiple signatures are covered, and so on.

PR #1288, opened 2023-01-17
put a reasonable size limit on ssh key reader

bobcallaway:
This PR limits the ssh public key size to 64K; this is covered under the external API request size limit, but adding an explicit test to  reduce risk in other usage.


asraa (dismissed):
nice! opening a tracking issue is probably a good idea

asraa (approved):

Hayden-IO (approved):

PR #1294, opened 2023-01-20
Handle possibly nil values gracefully

Hayden-IO:
When searching by entry, you could construct an entry without a required field, causing canonicalization to fail.

### DIFF
--- a/pkg/types/intoto/v0.0.1/entry.go
+++ b/pkg/types/intoto/v0.0.1/entry.go
@@ -193,6 +193,16 @@ func (v *V001Entry) Canonicalize(ctx context.Context) ([]byte, error) {
 	}
 	pkb := strfmt.Base64(pk)
 
+	// hash and payloadHash can be nil when providing an entry for /api/v1/log/retrieve
+	// that is not canonicalized or complete
+	// these fields are not required by the API since they are set during validation
+	if v.IntotoObj.Content.Hash == nil {
+		return nil, errors.New("hash must be set for intoto v0.0.1")
+	}
+	if v.IntotoObj.Content.PayloadHash == nil {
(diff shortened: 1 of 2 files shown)


asraa (commented):
Could you explain the context of this? 

I'm guessing it's that you're running `Canonicalize` on a client before running a Search by entry query. Where are you getting the entry though that you are canonicalizing, and why doesn't it have the ContentHash/PayloadHash? These seem required... (or is this a backwards compatible thing, where a while ago these types somehow skipped that... i can't remember)

Hayden-IO:
@bobcallaway @asraa PTAL

I would prefer to fix this by specifying two different schemas, one for input and one for a canonicalized output. This would allow us to set different required fields for each. This fix should be sufficient.

Hayden-IO:
@AdamKorcz This is an interesting vector for fuzzing. I suspect there's more instances of this. Happy to chat offline about the details.

asraa:
> I would prefer to fix this by specifying two different schemas, one for input and one for a canonicalized output. This would allow us to set different required fields for each. This fix should be sufficient.

I don't think we want a per-entry input schema, but rather an input schema that works across all types. Something like

ProposedEntry:
  * type: intoto
  * pk: pk
  * etc...

But it would be a breaking change :| so minimally let's fix it right now. I'll open up an issue which i'll flag as a v2 improvement

Hayden-IO:
Moving to draft as @bobcallaway is fixing this in a better way. Will close once we have merged the fixes for 0.0.1/2

Hayden-IO:
@bobcallaway Were you planning to add in the 0.0.2 changes too? Envelope is calculated server side, correct?

PR #1329, opened 2023-02-05
bump release builder image to use go1.20

cpanato:
#### Summary
- bump release builder image to use go1.20

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor
 
-go 1.20
+go 1.19
 
 require (
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor
 
-go 1.18
+go 1.20
 
 require (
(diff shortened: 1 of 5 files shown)


Hayden-IO (dismissed):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor
 
-go 1.18
+go 1.20
```
bobcallaway: What's the reasoning for making this the minimum version?

cpanato: Are we using this as a library in other places?
Do you want to bump it at least to 1.19? 1.18 will be deprecated soon

bobcallaway: I'd keep this at the lowest level possible, unless we need to use features from 1.20 that are deemed necessary. That way we are interoperable with the widest set of other projects (who compile us in via dependency). 

I just looked at https://go.dev/doc/devel/release - since 1.18 is out of support I'd vote yes to moving to 1.19

cpanato: pushed and set to 1.19

bobcallaway (approved):

go.mod
Hayden-IO: We are going to want to use ed25519ph for hashedrekord eventually, which comes in 1.20.

PR #1334, opened 2023-02-08
add feature to limit HTTP request body length to process

bobcallaway:

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -105,6 +105,8 @@ Memory and file-based signers should only be used for testing.`)
 
 	rootCmd.PersistentFlags().StringSlice("enabled_api_endpoints", operationIds, "list of API endpoints to enable using operationId from openapi.yaml")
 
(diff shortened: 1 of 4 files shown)


cmd/rekor-server/app/root.go
```diff
@@ -105,6 +105,8 @@ Memory and file-based signers should only be used for testing.`)
 
 	rootCmd.PersistentFlags().StringSlice("enabled_api_endpoints", operationIds, "list of API endpoints to enable using operationId from openapi.yaml")
 
+	rootCmd.PersistentFlags().Uint64("max_request_body_size", 0, "maximum size for HTTP request body, in bytes; set to 0 for unlimited")
```
Hayden-IO, line 108: Do you want to set a default limit rather than unlimited?

bobcallaway: For backward compatibility I didn't want to change behavior.

Hayden-IO (approved):

cmd/rekor-server/app/root.go
Hayden-IO: makes sense! maybe update the description to include a suggestion of a max value?

PR #1351, opened 2023-02-17
improve validation on intoto v0.0.2 type

bobcallaway:
This improves the validation of intoto v0.0.2 types when processed through the SearchLogQuery endpoint.


pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
```diff
@@ -93,7 +93,10 @@
                     ],
                     "readOnly": true
                 }
-            }
+            },
+            "required": [
```
Hayden-IO, line 97: To confirm, this shouldn’t break any existing callers, correct? The envelope should have always been specified?

Hayden-IO (approved):

Hayden-IO:
The new codecov tests may need to be configured or disabled, they’re failing in other projects too.

Hayden-IO:
https://docs.codecov.com/docs/commit-status to configure (this is actually a nice feature)

pkg/types/intoto/v0.0.2/intoto_v0_0_2_schema.json
bobcallaway: no, the envelope is required on upload anyway so this is actually more accurate of intent.

Hayden-IO (approved):

PR #1375, opened 2023-03-06
Update cloudbuild for cosign 2.0

Hayden-IO:
Missing identity flags, would have been noticed when we cut a release


cpanato (approved):
forgot to update that :( my bad

PR #1390, opened 2023-03-16
Add Rekor OID info.

wlynch:
@haydentherapper Finally circling back to this 🙏 

#### Summary

This defines OIDs for most log entry fields (though not all). This is primarily intended to be used by clients that want to represent Rekor data in different formats and/or cannot use bundles.

For example, gitsign cannot use the verification bundle since it uses PKCS7 signatures to keep compatibility with existing Git clients / signature formats. Since PKCS7 already defines a structure for payload hashes, it implements offline verification by deconstructing bundles into distinct OIDs then reconstructs them for verification. See https://github.com/sigstore/gitsign/pull/220 for more details.

#### Release Note

Adds OID definitions for LogEntry fields.

#### Documentation

✅

### DIFF
--- a/oid-info.md
+++ b/oid-info.md
@@ -9,20 +9,21 @@ Rekor reserves the `1.3.6.1.4.1.57264.3` OID root for all of its values.
 
 ## Directory
 
-| OID                       | Name                   | Description                                                                                                   |
-| ------------------------- | ---------------------- | ------------------------------------------------------------------------------------------------------------- |
-| 1.3.6.1.4.1.57264.3.1     | Integrated Time        | When the data was added to the log.                                                                           |
-| 1.3.6.1.4.1.57264.3.2     | Log ID                 | This is a SHA256 hash of the DER-encoded public key for the log at the time the entry was included in the log |
-| 1.3.6.1.4.1.57264.3.3     | Log Index              | The index of the entry in the transparency log.                                                               |
-| 1.3.6.1.4.1.57264.3.4     | Verification           | Log Entry Verification data.                                                                                  |
-| 1.3.6.1.4.1.57264.3.4.1   | Inclusion Proof        | Proof of inclusion on the transparency log.                                                                   |
-| 1.3.6.1.4.1.57264.3.4.1.1 | Checkpoint             | The checkpoint (signed tree head) that the inclusion proof is based on.                                       |
-| 1.3.6.1.4.1.57264.3.4.1.2 | Hashes                 | A list of hashes required to compute the inclusion proof, sorted in order from leaf to root.                  |
-| 1.3.6.1.4.1.57264.3.4.1.3 | Root Hash              | The hash value stored at the root of the merkle tree at the time the proof was generated.                     |
-| 1.3.6.1.4.1.57264.3.4.1.4 | Tree Size              | The size of the merkle tree at the time the inclusion proof was generated.                                    |
-| 1.3.6.1.4.1.57264.3.4.2   | Signed Entry Timestamp | Base64 encoded signature of the Inclusion Proof                                                               |
+| OID                       | Name                   | Tag Type               | Description                                                                                                   |
+| ------------------------- | ---------------------- | ---------------------- | ------------------------------------------------------------------------------------------------------------- |
+| 1.3.6.1.4.1.57264.3.1     | Integrated Time        | INTEGER                | When the data was added to the log.                                                                           |
+| 1.3.6.1.4.1.57264.3.2     | Log ID                 | UTF8STRING             | This is a SHA256 hash of the DER-encoded public key for the log at the time the entry was included in the log |
+| 1.3.6.1.4.1.57264.3.3     | Log Index              | INTEGER                | The index of the entry in the transparency log.                                                               |
+| 1.3.6.1.4.1.57264.3.4     | Verification           |                        | Log Entry Verification data.                                                                                  |
+| 1.3.6.1.4.1.57264.3.4.1   | Inclusion Proof        |                        | Proof of inclusion on the transparency log.                                                                   |
+| 1.3.6.1.4.1.57264.3.4.1.1 | Checkpoint             | UTF8STRING             | The checkpoint (signed tree head) that the inclusion proof is based on.                                       |
+| 1.3.6.1.4.1.57264.3.4.1.2 | Hashes                 | SEQUENCE OF UTF8STRING | A list of hashes required to compute the inclusion proof, sorted in order from leaf to root.                  |
+| 1.3.6.1.4.1.57264.3.4.1.3 | Root Hash              | UTF8STRING             | The hash value stored at the root of the merkle tree at the time the proof was generated.                     |
+| 1.3.6.1.4.1.57264.3.4.1.4 | Tree Size              | INTEGER                | The size of the merkle tree at the time the inclusion proof was generated.                                    |
+| 1.3.6.1.4.1.57264.3.4.2   | Signed Entry Timestamp | UTF8STRING             | Base64 encoded signature of the promise of inclusion.                                                         |
 
 ### Notes:
 
 - This is not an exhaustive list of values included in a LogEntry.
-- Log Index not included in Inclusion Proof since it's already defined in `1.3.6.1.4.1.57264.3.3`.
+- Log Index not included in Inclusion Proof since it's already defined in
+  `1.3.6.1.4.1.57264.3.3`.
--- a/oid-info.md
+++ b/oid-info.md
@@ -9,21 +9,6 @@ Rekor reserves the `1.3.6.1.4.1.57264.3` OID root for all of its values.
 
 ## Directory
 
-| OID                       | Name                   | Tag Type               | Description                                                                                                   |
-| ------------------------- | ---------------------- | ---------------------- | ------------------------------------------------------------------------------------------------------------- |
-| 1.3.6.1.4.1.57264.3.1     | Integrated Time        | INTEGER                | When the data was added to the log.                                                                           |
-| 1.3.6.1.4.1.57264.3.2     | Log ID                 | UTF8STRING             | This is a SHA256 hash of the DER-encoded public key for the log at the time the entry was included in the log |
-| 1.3.6.1.4.1.57264.3.3     | Log Index              | INTEGER                | The index of the entry in the transparency log.                                                               |
-| 1.3.6.1.4.1.57264.3.4     | Verification           |                        | Log Entry Verification data.                                                                                  |
-| 1.3.6.1.4.1.57264.3.4.1   | Inclusion Proof        |                        | Proof of inclusion on the transparency log.                                                                   |
-| 1.3.6.1.4.1.57264.3.4.1.1 | Checkpoint             | UTF8STRING             | The checkpoint (signed tree head) that the inclusion proof is based on.                                       |
-| 1.3.6.1.4.1.57264.3.4.1.2 | Hashes                 | SEQUENCE OF UTF8STRING | A list of hashes required to compute the inclusion proof, sorted in order from leaf to root.                  |
-| 1.3.6.1.4.1.57264.3.4.1.3 | Root Hash              | UTF8STRING             | The hash value stored at the root of the merkle tree at the time the proof was generated.                     |
-| 1.3.6.1.4.1.57264.3.4.1.4 | Tree Size              | INTEGER                | The size of the merkle tree at the time the inclusion proof was generated.                                    |
-| 1.3.6.1.4.1.57264.3.4.2   | Signed Entry Timestamp | UTF8STRING             | Base64 encoded signature of the promise of inclusion.                                                         |
-
-### Notes:
-
-- This is not an exhaustive list of values included in a LogEntry.
-- Log Index not included in Inclusion Proof since it's already defined in
-  `1.3.6.1.4.1.57264.3.3`.
+| OID                   | Name                 | Tag Type     | Description                                                                                                                                                        |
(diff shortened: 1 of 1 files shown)


oid-info.md
```diff
@@ -0,0 +1,28 @@
...
+## Description
+
+This document defines Rekor
+[OID values](https://github.com/sigstore/sigstore/blob/main/docs/oid-info.md).
+
+Rekor reserves the `1.3.6.1.4.1.57264.3` OID root for all of its values.
+
+## Directory
+
+| OID                       | Name                   | Description                                                                                                   |
```
asraa: drive-by: would it make sense to use a single OID for a "sigstore bundle" that is using the spec at protobuf-specs?

(That would avoid a whole mess of comments that we've sorted out there: e.g. what encoding is the Log ID, is Verification required, what encoding is the root hash, is the log index the global, etc)

oid-info.md
```diff
@@ -0,0 +1,28 @@
...
+| ------------------------- | ---------------------- | ------------------------------------------------------------------------------------------------------------- |
+| 1.3.6.1.4.1.57264.3.1     | Integrated Time        | When the data was added to the log.                                                                           |
```
Hayden-IO: Could you add a column for the [tag/type](https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/#tag) for each? I assume you'll just be using the universal class and not defining application specific types? As in, most will just be UTF8String or Integer?

oid-info.md
```diff
@@ -0,0 +1,28 @@
...
+| 1.3.6.1.4.1.57264.3.2     | Log ID                 | This is a SHA256 hash of the DER-encoded public key for the log at the time the entry was included in the log |
+| 1.3.6.1.4.1.57264.3.3     | Log Index              | The index of the entry in the transparency log.                                                               |
+| 1.3.6.1.4.1.57264.3.4     | Verification           | Log Entry Verification data.                                                                                  |
```
Hayden-IO: Could you clarify what is set for this field? Or is nothing set, and only its children are?

oid-info.md
```diff
@@ -0,0 +1,28 @@
...
+| 1.3.6.1.4.1.57264.3.4.1   | Inclusion Proof        | Proof of inclusion on the transparency log.                                                                   |
```
Hayden-IO: Same question here, is anything set for this OID?

oid-info.md
wlynch: My only concern is there would be duplicated data that may be in other locations (body/attestation) - for gitsign body is the main concern, but as long as it's a hashed checksum I guess it wouldn't be that much more?

Let me try playing around with this in my draft PR and see how much of a pain it is.

Hayden-IO: To confirm, do you mean to turn this into the sigstore bundle OID instead of Rekor?

I think what's here should be transferable to the protobuf-specs spec? Though I think some of the naming isn't aligned.

oid-info.md
```diff
@@ -0,0 +1,28 @@
...
+| 1.3.6.1.4.1.57264.3.4.1.1 | Checkpoint             | The checkpoint (signed tree head) that the inclusion proof is based on.                                       |
+| 1.3.6.1.4.1.57264.3.4.1.2 | Hashes                 | A list of hashes required to compute the inclusion proof, sorted in order from leaf to root.                  |
+| 1.3.6.1.4.1.57264.3.4.1.3 | Root Hash              | The hash value stored at the root of the merkle tree at the time the proof was generated.                     |
+| 1.3.6.1.4.1.57264.3.4.1.4 | Tree Size              | The size of the merkle tree at the time the inclusion proof was generated.                                    |
+| 1.3.6.1.4.1.57264.3.4.2   | Signed Entry Timestamp | Base64 encoded signature of the Inclusion Proof                                                               |
```
Hayden-IO:
```suggestion
| 1.3.6.1.4.1.57264.3.4.2   | Signed Entry Timestamp | Base64 encoded signature of the promise of inclusion                                                               |
```

Hayden-IO: Just want to be precise here, that the SET is only a promise and the above `3.4.1` is a stronger claim

oid-info.md
asraa: >  but as long as it's a hashed checksum I guess it wouldn't be that much more?

Yes, if it's not an attestation it would just be the checksum of the body.

Yes! I was suggesting turning it into a sigstore bundle OID

wlynch: Okay, tried it out using https://pkg.go.dev/github.com/sigstore/cosign/v2/pkg/cosign/bundle#RekorBundle as a proxy (let me know if there are major differences between this and protobuf-specs). I don't think we want to do this.

There's a few fields that end up being duplicated across the signature with the bundle:
- content hash
- signature
- cert chain
- rfc3161 signed timestamp 
- canonicalized_body (which if it acts like the cosign RekorBundle, includes the base64 encoded HashedRekord entry which re-encodes the hash, signature, and cert chain again)

This ends up ~doubling (and if we used the full protobuf-spec bundle perhaps tripling) the size of the signature to store data that we already have elsewhere in the PKCS7 signature.

What would be better is if there was a modified version of [`TransparencyLogEntry`](https://github.com/sigstore/protobuf-specs/blob/c8a23a43dc1c685d336d112864ba237a15061c53/protos/sigstore_rekor.proto#LL89C9-L89C29), but let the client re-construct the canonicalized_body from other data it already has.

Maybe we could do this by nil-ing out the field and encoding that? Otherwise we can stick the original plan of just defining OIDs for the individual values. Open to ideas here - let me know what you think!

(aside - long term I'd love to offer a `--- BEGIN SIGSTORE BUNDLE---` signature type and encode the proto bundle as is, but I don't think we have that luxury based on https://git-scm.com/docs/signature-format)

oid-info.md
wlynch: Done. Used UTF8STRING by default. We could technically use printable string for some of these, but I don't know if there's a good reason to prefer that over utf8.

oid-info.md
wlynch: Just a logical grouping based on the existing LogEntry type. Hopefully the tags make this clearer, but if there's anything else I should add let me know!

oid-info.md
wlynch: Nope.

oid-info.md
wlynch: Done.

oid-info.md
asraa: > What would be better is if there was a modified version of [TransparencyLogEntry](https://github.com/sigstore/protobuf-specs/blob/c8a23a43dc1c685d336d112864ba237a15061c53/protos/sigstore_rekor.proto#LL89C9-L89C29), but let the client re-construct the canonicalized_body from other data it already has.

Good point! if you had the TransparencyLogEntry with a nil canonicalized_body, that seems to be all the info needed for you to verify the Rekor signature for offline verification, right?

wlynch: I think so. If we go down this route, we should modify in the protobuf spec to make it clear this is an expected behavior / usage of TransparencyLogEntry. Do either of these options sound better to you?

1. We break out canonicalized_body from TransparencyLogEntry and create a new wrapper message that combines them.
2. We document the canonicalized_body field to say that it's valid for this field to be unset, in which case it's the responsibility of the client to construct it.

FWIW, it looks like canonicalized_body wasn't in the original version of TransparencyLogEntry - https://github.com/sigstore/protobuf-specs/pull/10. We don't have the same problem with Gitsign w.r.t. DSSE canonicalization, so being able to deterministically reconstruct the body shouldn't be an issue.

Hayden-IO (approved):
LGTM for the OIDs!

oid-info.md
asraa: I guess if @haydentherapper is OK with adding Rekor specific OID, that's fine - if something changes with the semantics then I worry this won't be maintained, but if GitSign is one of the only consumers that's OK

Hayden-IO: If there's a way to do this for the protos, definitely on board.

Another option is to move these OIDs back to Gitsign and not have this in Rekor, if Gitsign really is the only client.

Hayden-IO: The more I think about it, maybe that is the best way to proceed. GitSign uses `.3`, defines its own child OIDs, we concurrently define OIDs for the bundle, and then at some point, GitSign adopts the bundle and these new OIDs.

wlynch: What's the likelihood of detangling canonicalized_body from the rest of the TransparencyLogEntry in protobuf-specs? I think that would be my preference since that would probably be easiest to maintain long term. If you need me to open an issue elsewhere for discussion lmk!

For the OIDs, the initial idea was to put this in Rekor since we're describing Rekor types. Functionally it's fine if this lives in Gitsign, but I'd worry these may diverge over time.

oid-info.md
```diff
@@ -0,0 +1,29 @@
...
+| 1.3.6.1.4.1.57264.3.4.1   | Inclusion Proof        |                        | Proof of inclusion on the transparency log.                                                                   |
+| 1.3.6.1.4.1.57264.3.4.1.1 | Checkpoint             | UTF8STRING             | The checkpoint (signed tree head) that the inclusion proof is based on.                                       |
+| 1.3.6.1.4.1.57264.3.4.1.2 | Hashes                 | SEQUENCE OF UTF8STRING | A list of hashes required to compute the inclusion proof, sorted in order from leaf to root.                  |
+| 1.3.6.1.4.1.57264.3.4.1.3 | Root Hash              | UTF8STRING             | The hash value stored at the root of the merkle tree at the time the proof was generated.                     |
+| 1.3.6.1.4.1.57264.3.4.1.4 | Tree Size              | INTEGER                | The size of the merkle tree at the time the inclusion proof was generated.                                    |
+| 1.3.6.1.4.1.57264.3.4.2   | Signed Entry Timestamp | UTF8STRING             | Base64 encoded signature of the promise of inclusion.                                                         |
```
woodruffw: Small formatting nitpick, to make the ASN.1 types more visually distinct 🙂 

```suggestion
| 1.3.6.1.4.1.57264.3.1     | Integrated Time        | `INTEGER`                | When the data was added to the log.                                                                           |
| 1.3.6.1.4.1.57264.3.2     | Log ID                 | `UTF8STRING`             | This is a SHA256 hash of the DER-encoded public key for the log at the time the entry was included in the log |
| 1.3.6.1.4.1.57264.3.3     | Log Index              | `INTEGER`                | The index of the entry in the transparency log.                                                               |
| 1.3.6.1.4.1.57264.3.4     | Verification           |                        | Log Entry Verification data.                                                                                  |
| 1.3.6.1.4.1.57264.3.4.1   | Inclusion Proof        |                        | Proof of inclusion on the transparency log.                                                                   |
| 1.3.6.1.4.1.57264.3.4.1.1 | Checkpoint             | `UTF8STRING`             | The checkpoint (signed tree head) that the inclusion proof is based on.                                       |
| 1.3.6.1.4.1.57264.3.4.1.2 | Hashes                 | `SEQUENCE OF UTF8STRING` | A list of hashes required to compute the inclusion proof, sorted in order from leaf to root.                  |
| 1.3.6.1.4.1.57264.3.4.1.3 | Root Hash              | `UTF8STRING`             | The hash value stored at the root of the merkle tree at the time the proof was generated.                     |
| 1.3.6.1.4.1.57264.3.4.1.4 | Tree Size              | `INTEGER`                | The size of the merkle tree at the time the inclusion proof was generated.                                    |
| 1.3.6.1.4.1.57264.3.4.2   | Signed Entry Timestamp | `UTF8STRING`             | Base64 encoded signature of the promise of inclusion.                                                         |
```

oid-info.md
wlynch: Opened https://github.com/sigstore/protobuf-specs/issues/72 for more discussion

Hayden-IO (changes requested):
Un-lgtming just so we don't accidentally merge. Having a single OID and using the bundle format sounds good with me

wlynch:
Updated for the single OID based on https://github.com/sigstore/protobuf-specs/issues/72

I wasn't sure what the best tag type would be here (i.e. bit string vs octet string vs utf8string) - let me know if there's a better fit!

oid-info.md
wlynch: Done.

oid-info.md
```diff
@@ -0,0 +1,14 @@
...
+This document defines Rekor
+[OID values](https://github.com/sigstore/sigstore/blob/main/docs/oid-info.md).
+
+Rekor reserves the `1.3.6.1.4.1.57264.3` OID root for all of its values.
+
+## Directory
+
+| OID                   | Name                 | Tag Type     | Description                                                                                                                                                        |
+| --------------------- | -------------------- | ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| 1.3.6.1.4.1.57264.3.1 | TransparencyLogEntry | `BIT STRING` | Proto serialized [TransparencyLogEntry](https://github.com/sigstore/protobuf-specs/blob/4dbf10bc287d76f1bfa68c05a78f3f5add5f56fe/protos/sigstore_rekor.proto#L89). |
```
Hayden-IO: I'd go with utf8string, as it's a more modern type. https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/#bit-string-and-octet-string suggests BIT STRING or OCTET STRING are typically uses for nonces or digests

Hayden-IO (commented):
Thanks for making this update!

oid-info.md
wlynch: Done!

Hayden-IO (approved):
@bobcallaway This is ready to merge

PR #1407, opened 2023-03-28
remove goroutine usage from SearchLogQuery

bobcallaway:


Hayden-IO (approved):
Much cleaner! Looks good

PR #1408, opened 2023-03-28
drop log messages regarding attestation storage to debug

bobcallaway:


Hayden-IO (approved):

vaikas:
I think this is a fine change, but I don't think it's going to address the bug linked from here, since those were related to getting the attestations?
```
error fetching attestation by uuid
```

Hayden-IO:
that looks to be debug already - https://github.com/sigstore/rekor/blob/01a2321450286e395f543b92207e656bb8854aae/pkg/api/entries.go#L142

vaikas:
Yeah, but still showing up as error in the console, so I think that's what @k4leung4 was alluding to about that it's where it's being logged that's making it show up as an error?

Issue #1305, opened 2023-01-27
[v2] Separate proposed input types from the type-specific schemas


asraa:
**Description**

Clients should allow sending proposed entries as non strongly types entries, something like the following:
```
ProposedEntry:

type: intoto
pk: pk
etc...
```

that includes the proposed type it intends to be. Rekor server-side should create the type specific entry. 

This would provide clarity in the required fields of the server-side computed entries, and allow clients to not have to deal with type specific canonicalization. 

I think search, create, retrieve should all use this type of input data.

PR #1410, opened 2023-03-29
add new method to test insertability of proposed entries into log

bobcallaway:
#### Summary

Given the schema definitions for entry types, there is a fair bit of embedded logic that ensures that entry proposals have the correct mix of properties and not others. For code readability, this PR proposes an intermediate step towards an updated API that is a start towards addressing #1305 by adding a new method to the `EntryImpl` interface called `Insertable`, which ensures that the unmarshalled entry has the appropriate fields specified.

- [x] add unit tests for each type


pkg/types/entries.go
```diff
@@ -82,6 +83,12 @@ func CreateVersionedEntry(pe models.ProposedEntry) (EntryImpl, error) {
 		if !tf.(func() TypeImpl)().IsSupportedVersion(ei.APIVersion()) {
 			return nil, fmt.Errorf("entry kind '%v' does not support inserting entries of version '%v'", kind, ei.APIVersion())
 		}
+	} else {
+		return nil, fmt.Errorf("unknown kind '%v' specified", kind)
+	}
+
+	if ok, err := ei.Insertable(); !ok {
```
Hayden-IO, line 90: To confirm, this should not be a breaking change to existing entries, correct? It looks like this is only called on create - https://github.com/sigstore/rekor/blob/3282e1ab3e9b8f773e8a44861a31a3df9adf01ce/pkg/api/entries.go#L178

pkg/types/hashedrekord/v0.0.1/entry_test.go
```diff
@@ -297,9 +304,15 @@ func TestCrossFieldValidation(t *testing.T) {
 			if err != nil {
 				t.Errorf("unexpected err from Unmarshalling canonicalized entry for '%v': %v", tc.caseDesc, err)
 			}
-			if _, err := types.UnmarshalEntry(pe); err != nil {
+			ei, err := types.UnmarshalEntry(pe)
+			if err != nil {
 				t.Errorf("unexpected err from type-specific unmarshalling for '%v': %v", tc.caseDesc, err)
 			}
+			// hashedrekord is one of two types (rfc3161, hashedrekord) in that what is persisted is also insertable
```
Hayden-IO, line 311: What do you mean by this?

pkg/types/hashedrekord/v0.0.1/entry_test.go
```diff
@@ -284,6 +284,13 @@ func TestCrossFieldValidation(t *testing.T) {
 			t.Errorf("unexpected result in '%v': %v", tc.caseDesc, err)
```
Hayden-IO, line 284: Do we want to test a failure case for Insertable? Maybe at least for the primary three types, hashedrekord, rekord, intoto?

pkg/types/rfc3161/v0.0.1/entry.go
```diff
@@ -210,3 +210,15 @@ func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.A
 func (v V001Entry) Verifier() (pki.PublicKey, error) {
 	return nil, errors.New("Verifier() does not support rfc3161 entry type")
 }
+
+func (v V001Entry) Insertable() (bool, error) {
```
Hayden-IO, line 214: Should we specify writeOnly for TSR? https://github.com/sigstore/rekor/blob/main/pkg/types/rfc3161/v0.0.1/rfc3161_v0_0_1_schema.json#L15

More generally, do we want to be consistent with adding writeOnly tags to all types?

pkg/types/hashedrekord/v0.0.1/entry_test.go
bobcallaway: Proposed entries are potentially unmodified when canonicalized, so what is proposed may be exactly what is persisted. RFC3161 timestamps are also persisted as written, so the notion of “ensuring proposed entries are insertable but canonicalized entries are not” makes sense for some pluggable types but not others.

pkg/types/rfc3161/v0.0.1/entry.go
bobcallaway: It’s not writeOnly because it will be read out as well

pkg/types/hashedrekord/v0.0.1/entry_test.go
bobcallaway: Probably a good idea

pkg/types/entries.go
bobcallaway: Correct, it should only affect all new entries put into the log

pkg/types/hashedrekord/v0.0.1/entry_test.go
Hayden-IO: Makes sense, thanks

Hayden-IO (dismissed):

cpanato (dismissed):

pkg/types/intoto/v0.0.2/entry_test.go
```diff
@@ -499,3 +499,225 @@ func TestV002Entry_IndexKeys(t *testing.T) {
...
+				},
+				env: *env,
+			},
+			expectSuccess: false,
+		},
+		{
+			caseDesc: "missing content",
+			entry: V002Entry{
+				IntotoObj: models.IntotoV002Schema{
+					/*
```
Hayden-IO, line 695: Remove commented out test?

pkg/types/rekord/v0.0.1/entry_test.go
```diff
@@ -296,3 +296,194 @@ func TestUnspecifiedPKIFormat(t *testing.T) {
...
+		},
+		{
+			caseDesc: "missing signature content",
+			entry: V001Entry{
+				RekordObj: models.RekordV001Schema{
+					Data: &models.RekordV001SchemaData{
+						Content: strfmt.Base64([]byte("content")),
+					},
+					Signature: &models.RekordV001SchemaSignature{
+						//Content: &sig,
```
Hayden-IO, line 392: Here too

pkg/types/intoto/v0.0.2/entry_test.go
bobcallaway: Nope, that was intentional

pkg/types/rekord/v0.0.1/entry_test.go
bobcallaway: Nope, that was intentional

pkg/types/intoto/v0.0.2/entry_test.go
Hayden-IO: Ah I was reviewing on my phone, I see why

Hayden-IO (approved):

PR #1446, opened 2023-04-23
remove legacy encrypted cosign key

bobcallaway:
We are now using Fulcio-based ephemeral keys to sign rekor releases, so deleting the encrypted private key and Makefile target.


cpanato (approved):
thanks

PR #1461, opened 2023-05-02
Publish stable checkpoint periodically to Redis

### LINKED ISSUE #1067: Correctness issue: Change to batched entry upload to support witnessing
Witnesses monitor the consistency of the log, verifying that the log is append-only and immutable. Roughly, the verification process for a witness is:
* Persist a checkpoint (signed tree head)
* Periodically, request the current checkpoint
* Verify the signature on the checkpoint
* Request and verify a consistency proof between the persisted checkpoint and current checkpoint
* Persist the new checkpoint
* Publish a countersignature over the current checkpoint

Witnesses also help mitigate split-view attacks, where the log presents different views to different clients. As a consumer, I can use witnesses' countersignatures to verify that all witnesses have seen the same checkpoint. For a client that wants to verify an entry in the log and verify the log is consistent and no split-view attack is occurring, the client needs to:
* Verify the entry's inclusion proof
   * If presented with an SET (offline promise of inclusion), the client should request an inclusion proof
* Verify consistency by reaching quorum with a set of witnesses. This involves a) checking signatures for each countersignature from a witness, and b) verifying that the root hash that's been signed over is the same as the root hash of the inclusion proof.
   * Note that a client doesn't need to request a consistency proof if they trust the set of witnesses
   * It's up to the client how many witnesses constitutes quorum
   * Note that quorum could be reached via a gossip protocol too, where witnesses share signatures with one another.

This quorum mechanism requires that witnesses have seen the same root hash. This means that the root hash cannot frequently update. With the current state of Rekor, the root hash updates with almost every entry, because we've set Trillian's MMD (maximum merge delay, the time it takes for Trillian to process an entry for inclusion) to 0. This means that witnesses are very, very unlikely to ever witness the same checkpoint.

The proposed fix would be to move towards a model like CT, where entries are batched and processed periodically. Whatever this period is, we can set witnesses to poll more frequently than this period, giving them a chance to witness the same checkpoint. For example, if we set MMD to 5 minutes, we can have witnesses poll every minute or two.

The impact of this change is:
* The SET will no longer include a log index. This will be a **breaking change to clients**, because the SET signature is over the log index
* The SET will no longer include the integrated time. This timestamp will have to be the received/requested time, or a signed timestamp.
* The response on entry upload will no longer include the inclusion proof, also a **breaking client change**. I do not see this as a significant problem. Technically, an inclusion proof is a stronger statement than an SET (which is just a promise), but the log can always lie about the inclusion proof and present a split view. You still need to verify a) consistency with that inclusion proof's root hash, and b) find quorum. Persisting an inclusion proof saves one call to Rekor however, since for SET verification, you should also request an inclusion proof to confirm that the promise has been resolved and the entry is actually in Rekor.
* Entries can't be verified online immediately. This is not likely an issue for two reasons:
   * Most Sigstore clients have just been using SETs to "verify inclusion" and never querying the Rekor log
   * Artifacts are likely not going to be verified immediately after upload, but later when they're consumed. We just need to keep the MMD low enough (<1hr would be my recommendation).

Hayden-IO:
Currently, a log witness countersigns the latest checkpoint Rekor publishes. Rekor updates its checkpoint on every entry upload, which is extremely frequent. This means that two witnesses are very unlikely to countersign the same checkpoint. While gossiping, it will not be possible to reach a consensus on the same checkpoint, and therefore we can't mitigate split-view attacks.

This change publishes a checkpoint, a "stable checkpoint", every 5 minutes (configurable) to Redis. This runs as a goroutine, with a Redis key derived from the current time rounded to the nearest 5 minutes. We use set-if-not-exist for Redis, meaning you can run replicated instances of Rekor, with all instances writing to the same Redis key.

For a client that wants to gossip, this means waiting 5 minutes before a checkpoint is published that witnesses will countersign (Note that this is an area of active development and research too). The stable checkpoint can be accessed with a query parameter.

Fixes https://github.com/sigstore/rekor/issues/1067. There is still value in batching in terms of reliablity, but stable checkpoints solve the gossiping issue without a breaking change.


Hayden-IO:
@bobcallaway @asraa @jiggoha @mhutchinson @alcutter Looking for early feedback on the implementation while I'm writing tests

pkg/api/tlog.go
```diff
@@ -52,6 +53,39 @@ func GetLogInfoHandler(params tlog.GetLogInfoParams) middleware.Responder {
 		inactiveShards = append(inactiveShards, is)
 	}
 
+	stable := swag.BoolValue(params.Stable)
+	if stable {
+		ts := time.Now().Truncate(time.Duration(viper.GetUint("publish_frequency")) * time.Minute).UnixNano()
+		// key is treeID/timestamp, where timestamp is rounded down to the nearest X minutes
+		key := fmt.Sprintf("%d/%d", api.logRanges.ActiveTreeID(), ts)
```
Hayden-IO: Need to fix this, this should be "give me the latest", not by key (because the job might not have run for that time window yet)

pkg/witness/publish_checkpoint.go
```diff
@@ -0,0 +1,191 @@
...
+		log.Logger.Errorf("error signing checkpoint to publish: %v", err)
+		return
+	}
+
+	// encode checkpoint as hex to write to redis
+	hexCP := hex.EncodeToString(checkpoint)
+
+	// write checkpoint to Redis if key does not yet exist
+	// this prevents multiple instances of Rekor from writing different checkpoints in the same time window
+	ts := time.Now().Truncate(time.Duration(c.checkpointFreq) * time.Minute).UnixNano()
```
Hayden-IO, line 146: Update to only use as locking mechanism, don't need to store checkpoint too

Hayden-IO: Decided to leave this as is, as it gives us a history of checkpoints if need be.

pkg/api/tlog.go
```diff
@@ -53,6 +53,38 @@ func GetLogInfoHandler(params tlog.GetLogInfoParams) middleware.Responder {
 		inactiveShards = append(inactiveShards, is)
 	}
 
+	stable := swag.BoolValue(params.Stable)
+	if stable {
```
bobcallaway:
```suggestion
	if swag.BoolValue(params.Stable) {
```

pkg/api/tlog.go
```diff
@@ -53,6 +53,38 @@ func GetLogInfoHandler(params tlog.GetLogInfoParams) middleware.Responder {
...
+		key := fmt.Sprintf("%d/latest", api.logRanges.ActiveTreeID())
+		redisResult, err := redisClient.Get(params.HTTPRequest.Context(), key).Result()
+		if err != nil {
+			return handleRekorAPIError(params, http.StatusInternalServerError,
+				fmt.Errorf("error getting checkpoint from redis: %w", err), "error getting checkpoint from redis")
+		}
+		if redisResult == "" {
+			return handleRekorAPIError(params, http.StatusInternalServerError,
+				fmt.Errorf("no checkpoint found in redis: %w", err), "no checkpoint found in redis")
+		}
```
bobcallaway, line 68: Is this the desired behavior? It would change the default behavior given that you're defaulting to this being on.

pkg/witness/publish_checkpoint.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// on successfully obtaining the "lock" for the time window, update latest checkpoint
+	latestKey := fmt.Sprintf("%d/latest", c.treeID)
+	latestCtx, latestCancel := context.WithTimeout(c.ctx, 10*time.Second)
+	defer latestCancel()
+
+	// return value ignored, which is whether or not the entry was set
+	// no error is thrown if the key already exists
+	_, err = c.redisClient.Set(latestCtx, latestKey, hexCP, 0).Result()
+	if err != nil {
```
bobcallaway:
```suggestion
	if _, err = c.redisClient.Set(latestCtx, latestKey, hexCP, 0).Result(); err != nil {
```

pkg/witness/publish_checkpoint.go
```diff
@@ -0,0 +1,196 @@
...
+// means that a witness will not see a fresh checkpoint for an additional period.
+func (c *CheckpointPublisher) StartPublisher() {
+	tc := trillianclient.NewTrillianClient(context.Background(), c.logClient, c.treeID)
+	sTreeID := strconv.FormatInt(c.treeID, 10)
+
+	// publish on startup to ensure a checkpoint is available the first time Rekor starts up
+	c.publish(&tc, sTreeID)
+
+	ticker := time.NewTicker(time.Duration(c.checkpointFreq) * time.Minute)
+	go func() {
```
bobcallaway, line 91: you're leaking this goroutine as there's no way to quit it. 

You probably want to fix this, and also add the following into your unit test file to ensure it doesn't regress:
```go
import (
...
        "go.uber.org/goleak"
)

func TestMain(m *testing.M) {
        goleak.VerifyTestMain(m)
}
```

Hayden-IO: Fixed by adding a context and calling cancel on shutdown

pkg/api/tlog.go
Hayden-IO: Added a check that redisClient is not nil before entering this conditional. This would mean that if no redis client is initialized, stable=true does nothing, which I think is a reasonable default.

bobcallaway (approved):

PR #1469, opened 2023-05-04
Remove dependencies from pkg/util

mtrmac:
#### Summary
We are maintaining a Rekor client that calls `pkg/client`, and we care about the size of the binary

Because `pkg/client` depends on `pkg/util`, all dependencies of `pkg/util` end in the binary; recently that has grown quite a bit.

So, move things that aren’t shared between clients and servers into other sub packages, either directly into the only users, or into new packages.  See individual commit messages for details.

In our case, this saves 1.87 MB from a macOS binary, and should not affect the behavior of any users.

I certainly don’t have the full picture, so I’d be happy to adjust the final code locations and names however is necessary.

(It’s a bit tempting to completely eliminate the `pkg/util` subpackage, maybe by moving the truly shared parts into `pkg/clientserver`, and the test helpers into `pkg/test`; `util` is a name that attracts adding more and more dependencies over time. But this is enough for my purposes, and I’ll leave the design discussions of such a refactor to maintainers.)

#### Release Note
NONE

#### Documentation
N/A


Hayden-IO (approved):
Looks good, thanks!

bobcallaway (approved):

PR #1486, opened 2023-05-12
omit informational message if machine-parseable output has been requested

### LINKED ISSUE #1395: Please add "script-friendly output" flag to rekkor-cli
At present, `rekor-cli` is too verbose to be used in scripts becasue it produces too much extraneous output, even with `--format json` set, for example:

```
./go/bin/rekor-cli search --sha $(openssl dgst -sha256 blob | awk '{print $NF}') --format json
Found matching entries (listed by UUID):
{"UUIDs":["somehexgoeshere"]}
```

There should be way to output the relevant output _only_, i.e. no "Found matching entries (listed by UUID):" line.

Its the same with `rekor-cli get`, which outputs extraneous header lines and hides the actual json behind `Body:`:

> 
> LogID: foo
> Index: bar
> IntegratedTime: 2023-01-01T01:01:01Z
> UUID: foobar
> Body:

bobcallaway:
Fixes: #1395

### DIFF
--- a/cmd/rekor-cli/app/search.go
+++ b/cmd/rekor-cli/app/search.go
@@ -201,7 +201,9 @@ var searchCmd = &cobra.Command{
 			return nil, fmt.Errorf("no matching entries found")
 		}
 
-		fmt.Fprintln(os.Stderr, "Found matching entries (listed by UUID):")
+		if viper.GetString("format") != "json" {
+			fmt.Fprintln(os.Stderr, "Found matching entries (listed by UUID):")
+		}
 
 		return &searchCmdOutput{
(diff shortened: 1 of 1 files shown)


cpanato (approved):

Hayden-IO (approved):
Does this issue exist in rekor-CLI get? If so, we should leave the ticket open

bobcallaway:
> Does this issue exist in rekor-CLI get? If so, we should leave the ticket open

If I use `--format=json` on `rekor-cli get`, I don't see the behavior reported in #1395 - the `default` format isn't meant to be machine-parseable.

Issue #1278, opened 2023-01-13
EntryImpl.Verifier only returns a single public key, but should support returning multiple keys


bobcallaway:
**Description**

intoto envelopes support multiple signatures with potentially multiple unique public keys, so Verifier should return `[]pki.PublicKey` instead of a single one

@haydentherapper

Hayden-IO:
SG, I’ll take a look at this. Going to do some refactoring to make it easier to know what key type you’re working with too.

PR #1487, opened 2023-05-12
add dsse type

bobcallaway:
#### Summary
This adds a new pluggable type `dsse`, which adds entries for [DSSE](https://github.com/secure-systems-lab/dsse) JSON documents which are often used with in-toto attestations. There are a few significant differences between this `dsse` type and the already existing `intoto` types:

- The JSON schema more clearly separates the proposed entry from what is persisted to simplify client expectations
- The DSSE envelope is provided as a JSON string, rather than double-encoding values in base64
- The DSSE envelope is not stored by Rekor; this is to set the client expectation that attestation storage in Rekor should not be relied upon for attestation discovery. 

#### Release Note

* Adds a new entry type `dsse` which persisted information about DSSE envelopes without storing the envelope in Rekor

#### Documentation

- [x] Add docs on type to docs.sigstore.dev


cpanato (dismissed):

AdamKorcz:
Fuzzing-part looks good!

pkg/types/dsse/README.md
```diff
@@ -0,0 +1,26 @@
...
+**Recognized content types**
+
+- [in-toto
+  statements](https://github.com/in-toto/attestation/tree/main/spec#statement)
+  are recognized and parsed. The found subject hashes are indexed so
+  they can be searched for.
+
+**What data about the envelope is stored in Rekor**
+
+Only the hash of the payload, the hash of the DSSE envelope, the signature(s)
```
Hayden-IO: nit: Just for newcomers, worth defining that payload != DSSE envelope? particularly because payload is overloaded with DSSE

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
```diff
@@ -0,0 +1,96 @@
...
+    "$id": "http://rekor.sigstore.dev/types/dsse/dsse_v0_0_1_schema.json",
+    "title": "DSSE v0.0.1 Schema",
+    "description": "Schema for DSSE envelopes",
+    "type": "object",
+    "properties": {
+        "proposedContent": {
+            "type": "object",
+            "properties": {
+                "envelope": {
+                    "description": "DSSE envelope specified as a stringified JSON object; payloadType in the envelope MUST be set to 'application/vnd.in-toto+json'",
```
Hayden-IO: YAGNI, but I wonder if it'd useful to support more than just this payload type. Given we're only storing the hash of the envelope, should the server not care about payload type?

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
```diff
@@ -0,0 +1,96 @@
...
+                    "type": "string",
+                    "writeOnly": true
+                },
+                "publicKeys": {
```
Hayden-IO: One bit of feedback I've heard is it's confusing that this can either be a raw key or a certificate - should we rename this to "sigVerifiers" or something like that? Alternatively, specify that this can either be a raw public key or X.509 certificate

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
```diff
@@ -0,0 +1,96 @@
...
+            "items": {
+                "description": "a signature of the envelope's payload along with the public key for the signature",
+                "type": "object",
+                "properties": {
+                    "signature": {
+                        "description": "base64 encoded signature of the payload",
+                        "type": "string",
+                        "pattern": "^(?:[A-Za-z0-9+\\/]{4})*(?:[A-Za-z0-9+\\/]{2}==|[A-Za-z0-9+\\/]{3}=|[A-Za-z0-9+\\/]{4})$"
+                    },
+                    "publicKey": {
```
Hayden-IO: Edge case, but what happens if multiple public keys can verify one signature? As long as order is well-defined and the public keys from proposedContent are processed in the same order, this should be fine

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+			return result, fmt.Errorf("could not canonicalize key: %w", err)
+		}
+
+		keyHash := sha256.Sum256(canonKey)
+		result = append(result, "sha256:"+hex.EncodeToString(keyHash[:]))
+
+		result = append(result, keyObj.Subjects()...)
+	}
+
+	if v.DSSEObj.PayloadHash != nil {
```
Hayden-IO, line 94: Shouldn't payloadhash and envelopehash always be set? Is this just defensive?

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+		result = append(result, envelopeHashKey)
+	}
+
+	if v.env == nil {
+		log.Logger.Info("DSSEObj content or DSSE envelope is nil, returning partial set of keys")
+		return result, nil
+	}
+
+	switch v.env.PayloadType {
+	case in_toto.PayloadType:
```
Hayden-IO, line 110: On insert, are we verifying that this is the only payload type supported?

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+
+	// field validation
+	if err := dsseObj.Validate(strfmt.Default); err != nil {
+		return err
+	}
+
+	// either we have just proposed content or the canonicalized fields
+	if dsseObj.ProposedContent == nil {
+		// then we need canonicalized fields, and all must be present (if present, they would have been validated in the above call to Validate())
+		if dsseObj.EnvelopeHash == nil || dsseObj.PayloadHash == nil || len(dsseObj.Signatures) == 0 {
```
Hayden-IO, line 183: To confirm, because we have Insertable(), we know that the DSSEObj must have valid signatures at this point, correct? As in, it's not possible to insert a DSSE envelope with invalid signatures because Insertable checks that proposed content is set, and then we check the signatures if proposed content is set

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+	// if we're here, then we're trying to propose a new entry so we check to ensure client's aren't setting server-side computed fields
+	if dsseObj.EnvelopeHash != nil || dsseObj.PayloadHash != nil || len(dsseObj.Signatures) != 0 {
+		return errors.New("either proposedContent or envelopeHash, payloadHash, and signatures must be present but not both")
+	}
+
+	env := &dsse.Envelope{}
+	if err := json.Unmarshal([]byte(*dsseObj.ProposedContent.Envelope), env); err != nil {
+		return err
+	}
+
```
Hayden-IO, line 198: Should we check the payload type here?

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil {
+		return err
+	}
+
+	// we need to ensure we canonicalize the ordering of signatures
+	sortedSigs := make([]string, 0, len(sigToKeyMap))
+	for sig := range sigToKeyMap {
+		sortedSigs = append(sortedSigs, sig)
+	}
+	sort.Strings(sortedSigs)
```
Hayden-IO, line 218: To confirm, these sigs are base64 encoded already, correct? The DSSE spec says signatures must be base64 encoded, I assume the DSSE verifier checks this?

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+		allPubKeyBytes = append(allPubKeyBytes, props.PublicKeyBytes...)
+	}
+
+	if len(props.PublicKeyPaths) > 0 {
+		for _, path := range props.PublicKeyPaths {
+			if path.IsAbs() {
+				return nil, errors.New("dsse public keys cannot be fetched over HTTP(S)")
+			}
+
+			publicKeyBytes, err := os.ReadFile(filepath.Clean(path.Path))
```
Hayden-IO, line 323: Is there any threat of SSRF? path is controlled by the caller. Reading a random file should cause a failure on verification though

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("could not load verifier: %w", err)
+		}
+
+		dsseVfr, err := dsse.NewEnvelopeVerifier(&sigdsse.VerifierAdapter{SignatureVerifier: vfr})
+		if err != nil {
+			return nil, fmt.Errorf("could not use public key as a dsse verifier: %w", err)
+		}
+
+		accepted, err := dsseVfr.Verify(context.Background(), env)
```
Hayden-IO, line 381: should you pass the context from the caller? CreateFromArtifactProps has a context, but Unmarshal doesnt

pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -0,0 +1,420 @@
...
+
+	return verifierBySig, nil
+}
+
+func (v V001Entry) Verifier() (pki.PublicKey, error) {
+	if len(v.DSSEObj.Signatures) == 0 {
+		return nil, errors.New("dsse v0.0.1 entry not initialized")
+	}
+
+	//TODO: return multiple pki.PublicKeys
```
Hayden-IO: Can you link to https://github.com/sigstore/rekor/issues/1278 so I make sure to change this

pkg/types/dsse/v0.0.1/e2e_test.go
```diff
@@ -0,0 +1,723 @@
...
+			t.Errorf("unexpected error returned from rekor: %v", err.Error())
+		}
+
+		dsse_eimpl := eimpl.(*V001Entry)
+
+		if len(dsse_eimpl.DSSEObj.Signatures) != 2 {
+			t.Errorf("incorrect number of signatures returned in response: expected 2, got %d", len(dsse_eimpl.DSSEObj.Signatures))
+		}
+	}
+}
```
Hayden-IO, line 723: A test for rejecting non-intoto payload types? (pending other comment, if we want to enforce that)

pkg/types/dsse/v0.0.1/entry_test.go
```diff
@@ -0,0 +1,509 @@
...
+		{
+			env:  envelope(t, key, []byte(validPayload)),
+			name: "invalid key",
+			it: &models.DSSEV001Schema{
+				ProposedContent: createRekorEnvelope(envelope(t, key, []byte(validPayload)), [][]byte{[]byte("notavalidkey")}),
+			},
+			wantErr: true,
+		},
+		{
+			env:  multiSignEnvelope(t, []*ecdsa.PrivateKey{key, priv}, []byte(validPayload)),
```
Hayden-IO, line 214: Do we have a test for a mismatch between signatures and public keys, that not all keys can verify all signatures?

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
bobcallaway: the code doesn't actually enforce this, will remove the content.

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
bobcallaway: would `verifiers` be OK then?

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
bobcallaway: I think the first matching key would be used and the rest discarded.

pkg/types/dsse/v0.0.1/dsse_v0_0_1_schema.json
Hayden-IO: Yea, `verifiers` SGTM

pkg/types/dsse/v0.0.1/entry.go
bobcallaway: yes, just defensive in case it was called on a proposed entry

pkg/types/dsse/v0.0.1/entry.go
bobcallaway: no, not strictly

pkg/types/dsse/v0.0.1/entry.go
bobcallaway: Insertable is called *after* Unmarshal on the insert path.

pkg/types/dsse/v0.0.1/entry.go
bobcallaway: `verifyEnvelope` will do all of that and would fail above on line 210 if they can't be used to verify the signatures.

pkg/types/dsse/v0.0.1/entry.go
bobcallaway: this is only invoked from the CLI process so no SSRF concern

pkg/types/dsse/v0.0.1/entry.go
bobcallaway: I'd either have to set the request context as a variable on the struct as otherwise I'd have to make a change to the golang interface

pkg/types/dsse/v0.0.1/entry_test.go
bobcallaway: those are in e2e tests

pkg/types/dsse/v0.0.1/e2e_test.go
bobcallaway: won't enforce

pkg/types/dsse/v0.0.1/entry.go
bobcallaway: looking at this, it will just be an in-memory RSA/ECDSA/ED25519 verification so context isn't used anywhere.

pkg/types/dsse/v0.0.1/entry.go
Hayden-IO: OK, that should be good - If the object is missing the proposed content, Insertable fails. If the object has proposed content but an invalid signature, it'll fail unmarshalling.

Hayden-IO (dismissed):

bdehamer (dismissed):
Thanks for tackling this. Looking forward to using this in place of the `intoto` type!

codysoyland (dismissed):
Great stuff, thank @bobcallaway! 🎉

Hayden-IO (dismissed):

Hayden-IO (approved):

jul-sh:
> The DSSE envelope is not stored by Rekor; this is to set the client expectation that attestation storage in Rekor should not be relied upon for attestation discovery.

@bobcallaway to confirm, is this the desired expectation for for all rekor entries or specific to the DSSE type? 

Rekor's search functionality, as well as its UI for viewing in-toto SLSA provenances positioned it as a useful tool for attestation discovery.

PR #1490, opened 2023-05-16
Remove requirement of PayloadHash for intoto 0.0.1

Hayden-IO:
The payload hash was added later on, so old committed entries are missing the payload hash. This means that canonicalization will fail when searching for the entry if we require the payload hash.


bobcallaway:
Could you please verify this case is covered by the tests in harness and if not, add one explicitly for this? thx

Hayden-IO:
@bobcallaway added a test to demonstrate no failure without payload hash

bobcallaway (approved):

PR #1504, opened 2023-05-24
Add concurrency to backfill-redis

### LINKED ISSUE #1501: Add concurrency to backfill-redis script
**Description**

The `backfill-redis` script processes items sequentially, including a blocking API call to Rekor in each iteration. It might make sense to use a worker pool to parallelize the Rekor index lookup, with a configurable pool size defaulting to `1`. This could greatly reduce the time it takes to perform a backfill.

codysoyland:
This adds concurrency to the `backfill-redis` command, including:

- Configurable `-concurrency` flag to set worker pool size, defaulting to 1
- `-dry-run` flag to test without writing to Redis
- Signal handler to exit on SIGTERM

Fixes #1501

### DIFF
--- a/cmd/backfill-redis/main.go
+++ b/cmd/backfill-redis/main.go
@@ -20,8 +20,8 @@
 
 	To run:
 	go run cmd/backfill-redis/main.go --rekor-address <address> \
-	    --hostname <redis-hostname> --port <redis-port>
-		--start <first index to backfill> --end <last index to backfill>
+	    --hostname <redis-hostname> --port <redis-port> --concurrency <num-of-workers> \
+		--start <first index to backfill> --end <last index to backfill> [--dry-run]
 */
 
 package main
@@ -34,9 +34,12 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/go-openapi/runtime"
 	"github.com/redis/go-redis/v9"
+	"golang.org/x/sync/errgroup"
 	"sigs.k8s.io/release-utils/version"
 
 	"github.com/sigstore/rekor/pkg/client"
@@ -66,6 +69,8 @@ var (
 	endIndex      = flag.Int("end", -1, "Last index to backfill")
 	rekorAddress  = flag.String("rekor-address", "", "Address for Rekor, e.g. https://rekor.sigstore.dev")
 	versionFlag   = flag.Bool("version", false, "Print the current version of Backfill Redis")
+	concurrency   = flag.Int("concurrency", 1, "Number of workers to use for backfill")
+	dryRun        = flag.Bool("dry-run", false, "Dry run - don't actually insert into Redis")
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/backfill-redis/main.go
```diff
@@ -106,45 +111,57 @@ func main() {
...
-			keys, err := e.IndexKeys()
+		index := i // capture loop variable for closure
+		group.Go(func() error {
+			params := entries.NewGetLogEntryByIndexParamsWithContext(ctx)
+			params.SetLogIndex(int64(index))
+			resp, err := rekorClient.Entries.GetLogEntryByIndex(params)
 			if err != nil {
-				insertErrs = append(insertErrs, fmt.Errorf("error building index keys for %s: %v", uuid, err))
-				continue
+				log.Fatalf("retrieving log uuid by index: %v", err)
```
Hayden-IO, line 150: Can we change this to just print the error?

cmd/backfill-redis/main.go
```diff
@@ -106,45 +111,57 @@ func main() {
...
+				}
+			} else {
+				fmt.Printf("Completed log index %d\n", index)
 			}
-		} else {
-			fmt.Printf("Completed log index %d\n", i)
-		}
+			return nil
+		})
+	}
```
Hayden-IO, line 196: Could we also track which indices couldn't be backfilled and log those?

cmd/backfill-redis/main.go
codysoyland: We can, but I want to highlight that this is the current behavior (I wish the GitHub diff showed character-level diff so we can see that this whole block was indented and barely modified), and when I ran this on production (over ~50,000 entries), I never hit this. So IMO, I would rather retain this behavior -- bail if the Rekor call fails, since that likely indicates a config/networking problem. Let me know, I'm happy to change it if you disagree.

Hayden-IO: Fair point, I'm ok with keeping this as is.

cmd/backfill-redis/main.go
codysoyland: Not sure if this is what you had in mind, but I added a result summary in my last commit. It does make a simple script a bit more complex, so I'm totally fine removing/changing it.

I decided to separate "parse errors" from "insert errors" when collating results, as I think parse errors are sort of unactionable and insert errors might be recoverable.

The results appear at the end of the command output, for example:
```
...
Uploaded Redis entry 24296fb24b8ad77abe5d0de220d5e64d0b52178d175a5833a5e327928368a2c9beea87009046c16d, index 21320998, key sha256:ee2ae5b13687d4d443a206359ad7c24a31b753c2233d9d1cb04dbe88535d09db
Completed log index 21320998
Backfill complete
Failed to parse 2 entries: [21320576 21320640]
Failed to insert/remove 1 entries: [21320316]
````

Hayden-IO: This looks great!!

Hayden-IO (approved):
Thanks for cleaning this up!

PR #1511, opened 2023-05-26
Move github.com/sigstore/protobuf-specs users into a separate subpackage

mtrmac:
#### Summary

Nothing else in this repository currently uses sigstore/protobuf-specs, notably prodution users of pkg/client don't use it. So, move the new TLE utilities into a separate subpackage, to make clients smaller.

I’m not sure what the long-term plan is, but at least short-term avoiding these dependencies in clients seems valuable. Compare #1469.

#### Release Note
NONE

#### Documentation
N/A


mtrmac:
(Also the `github.com/sigstore/rekor/pkg/types` dependency adds all the individual type implementations and _their_ dependencies, and a middleware stack via `pkg/log`.)

bobcallaway:
It is anticipated that users of pkg/client will make use of the TLE, as we work towards wider adoption of the sigstore bundle format defined in protobuf-specs; but I understand your concern on space and the dependency graph. 

If you're really space constrained, I would recommend just writing your own client against the REST API rather than importing pkg/client.

bobcallaway (approved):

mtrmac:
Thanks, that’s good to know; I certainly don’t want to keep futilely pushing against. and disrupting, intentional and planned work.

A separately-maintained client is quite practical for the current API, we’ll need to see how that works for the new data types. Actually protobuf might not make a difference for us, it’s more likely the dependency stack of `rekor/pkg/types` (`pkg/log` depending on a large middleware suite, and type validation with an extensive dependency stack), right now primarily because the `APIVersion` field seems to be type-specific instead of in the base `models.ProposedEntry`.

@Luap99 FYI .

PR #1561, opened 2023-06-24
feat: Support publishing new log entries to Pub/Sub topics

jalseth:
#### Summary

Adds initial support publishing new log entries to Pub/Sub topics. Interested parties can subscribe to the topic in order to receive notifications when new entries are added.

This relates to #191. 

#### Release Note

* New features: Added support publishing new log entries to Pub/Sub topics
* Config changes: Added the `rekor_server.new_entry_publisher` flag to configure the Pub/Sub topic to send notifications to when a new log entry is added. This flag is optional.

#### Documentation

This is a TODO still. I would like feedback on the PR before writing the docs.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -31,6 +31,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/pubsub"
 	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/signer"
 	"github.com/sigstore/rekor/pkg/storage"
@@ -39,6 +40,8 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
+
+	_ "github.com/sigstore/rekor/pkg/pubsub/gcp" // Load GCP pubsub implementation
 )
 
 func dial(ctx context.Context, rpcServer string) (*grpc.ClientConn, error) {
@@ -63,6 +66,9 @@ type API struct {
 	signer     signature.Signer
 	// stops checkpoint publishing
 	checkpointPublishCancel context.CancelFunc
+	// Publishes notifications when new entries are added to the log. May be
+	// nil if no publisher is configured.
+	newEntryPublisher pubsub.Publisher
 }
 
 func NewAPI(treeID uint) (*API, error) {
@@ -112,6 +118,14 @@ func NewAPI(treeID uint) (*API, error) {
 
 	pubkey := cryptoutils.PEMEncode(cryptoutils.PublicKeyPEMType, b)
 
+	var newEntryPublisher pubsub.Publisher
+	if p := viper.GetString("rekor_server.new_entry_publisher"); p != "" {
+		newEntryPublisher, err = pubsub.Get(ctx, p)
+		if err != nil {
+			return nil, fmt.Errorf("init event publisher: %w", err)
+		}
+	}
+
 	return &API{
 		// Transparency Log Stuff
 		logClient: logClient,
@@ -121,6 +135,8 @@ func NewAPI(treeID uint) (*API, error) {
 		pubkey:     string(pubkey),
 		pubkeyHash: hex.EncodeToString(pubkeyHashBytes[:]),
 		signer:     rekorSigner,
+		// Utility functionality not required for operation of the core service
+		newEntryPublisher: newEntryPublisher,
 	}, nil
 }
 
@@ -165,5 +181,8 @@ func ConfigureAPI(treeID uint) {
 }
 
 func StopAPI() {
+	if api.newEntryPublisher != nil {
+		api.newEntryPublisher.Close()
+	}
 	api.checkpointPublishCancel()
 }
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -39,6 +39,7 @@ import (
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/generated/restapi/operations/entries"
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/pubsub"
 	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/trillianclient"
 	"github.com/sigstore/rekor/pkg/types"
@@ -290,7 +291,7 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 		RootHash:   swag.String(hex.EncodeToString(root.RootHash)),
 		LogIndex:   swag.Int64(queuedLeaf.LeafIndex),
 		Hashes:     hashes,
-		Checkpoint: stringPointer(string(scBytes)),
+		Checkpoint: swag.String(string(scBytes)),
 	}
 
 	logEntryAnon.Verification = &models.LogEntryAnonVerification{
@@ -301,6 +302,31 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	logEntry := models.LogEntry{
 		entryID: logEntryAnon,
 	}
+
+	if api.newEntryPublisher != nil {
+		// Publishing notifications should not block the API response.
+		go func() {
+			var subjects []string
+			verifier, err := entry.Verifier()
+			if err != nil {
+				log.ContextLogger(ctx).Warnf("Could not get verifier for log entry: %w", err)
+			} else {
+				subjects = verifier.Subjects()
+			}
+			event, err := pubsub.BuildNewEntryEvent(entryID, logEntryAnon, subjects)
+			if err != nil {
+				log.ContextLogger(ctx).Error(err)
+				return
+			}
+			err = api.newEntryPublisher.Publish(context.Background(), event)
+			if err != nil {
+				log.ContextLogger(ctx).Error(err)
+				return
+			}
+			log.ContextLogger(ctx).Debugf("Published new entry event: %+v", event)
+		}()
+	}
+
 	return logEntry, nil
 }
 
--- a/pkg/pubsub/gcp/gcp.go
+++ b/pkg/pubsub/gcp/gcp.go
@@ -1,0 +1,110 @@
+// Copyright 2023 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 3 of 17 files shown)


pkg/api/api.go
```diff
@@ -165,5 +178,28 @@ func ConfigureAPI(treeID uint) {
 }
 
 func StopAPI() {
+	if api.newEntryPublisher != nil {
+		api.newEntryPublisher.Close()
+	}
 	api.checkpointPublishCancel()
 }
+
+func selectEntryPublisher(ctx context.Context) (pubsub.Publisher, error) {
```
bobcallaway: I'd like to see this moved into `pkg/pubsub` and mimic a factory pattern, similar to how https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/kms.go does this for KMS providers.

pkg/pubsub/gcp/gcp.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, fmt.Errorf("create pubsub client for project %q: %w", projectID, err)
+	}
+	return &Publisher{
+		client: client,
+		topic:  topic,
+	}, nil
+}
+
+func (p *Publisher) Publish(ctx context.Context, entry models.LogEntryAnon) error {
+	data, err := json.Marshal(entry)
```
bobcallaway: instead of JSON marshalling `models.LogEntryAnon` directly, we should really convert this into the Sigstore bundle format using the two methods in https://github.com/sigstore/rekor/blob/main/pkg/tle/tle.go that can convert `models.LogEntryAnon` into the bundle format JSON.

pkg/pubsub/gcp/gcp.go
```diff
@@ -0,0 +1,69 @@
...
+
+func New(ctx context.Context, refStr string, opts ...option.ClientOption) (*Publisher, error) {
+	projectID, topic, err := parseRef(refStr)
+	if err != nil {
+		return nil, fmt.Errorf("parse ref: %w", err)
+	}
+	client, err := pubsub.NewClient(ctx, projectID, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("create pubsub client for project %q: %w", projectID, err)
+	}
```
bobcallaway, line 62: should we check for the existence of the topic at this point?

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+
+	"github.com/sigstore/rekor/pkg/generated/models"
+)
+
+// Publisher provides methods for publishing notifications of new entries in the
+// transparency log.
+type Publisher interface {
+	// Publish publishes a transparency log entry to the configured Pub/Sub topic.
+	Publish(ctx context.Context, entry models.LogEntryAnon) error
```
bobcallaway: I'd like to see the entryID passed down here and published as a field within the pubsub message

pkg/pubsub/gcp/gcp.go
bobcallaway: We had also talked about potentially defining a https://cloudevents.io/ schema for these messages, which would be nice to do from the start. Thoughts?

pkg/api/entries.go
```diff
@@ -301,6 +301,16 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	logEntry := models.LogEntry{
 		entryID: logEntryAnon,
 	}
+
+	if api.newEntryPublisher != nil {
+		// Publishing notifications should not block the API response.
+		go func() {
```
bobcallaway, line 308: I don't think the GCP publish() call blocks (per https://pkg.go.dev/cloud.google.com/go/pubsub#Topic.Publish)... should we simply define the interface as "must not block" and defer this down, or leave this as is? Curious to other folks' opinion

pkg/api/entries.go
```diff
@@ -301,6 +301,16 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
...
+			if err := api.newEntryPublisher.Publish(ctx, logEntryAnon); err != nil {
+				log.Logger.Error(err)
```
bobcallaway: please use `log.ContextLogger(ctx).Error(err)` instead

bobcallaway (commented):
Thanks so much @jalseth for the contribution! I put some thoughts inline; tagging @haydentherapper @vaikas @lukehinds for thoughts here too

pkg/api/entries.go
jalseth: client.Publish() does not block, but the subsequent [PublishResult.Get()](https://pkg.go.dev/cloud.google.com/go/internal/pubsub#PublishResult.Get) does block until success, error, or timeout. The timeout is 60s, so this call could be blocking up to 60s. The Publish() call does not return an error, so you won't know the status of the call until you add the blocking call.

I prefer to have the Publisher.Publish() interface defined as blocking (must return an error) and leave the option to make it non-blocking up to the caller as is done here.

pkg/pubsub/gcp/gcp.go
jalseth: That's a good idea. When rolling out new code or new flags the server should error immediately upon startup where possible.

pkg/pubsub/gcp/gcp.go
jalseth: Ack re using the TLE methods.

I wasn't aware of the Cloud Events spec, but it seems reasonable to follow. I had planned on adding some of these anyway in a future PR, I can tackle that then if it is OK with you. I prefer to keep individual PRs smaller and easier to review, atomically roll-back if needed, etc.

Edit: I saw the example and realized the attributes are encoded in the message rather than just metadata, which would change the message significantly. I guess we do need to decide this now rather than later.

pkg/api/entries.go
```diff
@@ -301,6 +301,16 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	logEntry := models.LogEntry{
 		entryID: logEntryAnon,
 	}
+
+	if api.newEntryPublisher != nil {
+		// Publishing notifications should not block the API response.
+		go func() {
+			if err := api.newEntryPublisher.Publish(ctx, logEntryAnon); err != nil {
```
bobcallaway: one other thought - should we be doing this on the same context as the HTTP call or on a clone of it? If the client closes the connection before the response from `PublishResult.Get()` is successfully read, could we have a situation where the the context was cancelled but we still want to publish the message?

pkg/pubsub/pubsub.go
jalseth: What consumer do you envision using this data, and for what purpose?

pkg/api/entries.go
jalseth: I think you're right, good call out!

pkg/api/api.go
jalseth: Ack, done.

pkg/pubsub/pubsub.go
bobcallaway: the entryID uniquely identifies the entry within a given log (aka its merkle leaf hash). It can be recomputed by the other data in the bundle, but since we know it at the time the message is created it would be nice to pass it along as-is. 

We use it in the REST api to fetch a specific entry, e.g. `https://rekor.sigstore.dev/api/v1/log/entries/{entryID}`

lukehinds: +1 entryID is going to be useful.

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,78 @@
...
+
+var providersMap = map[string]ProviderInit{}
+
+// Get returns a Publisher for the given resource string and hash function.
+// If no matching provider is found, Get returns a ProviderNotFoundError. It
+// also returns an error if initializing the Publisher fails. If no resource
+// is supplied, it returns a nil Publisher and no error.
+func Get(ctx context.Context, topicResourceID string) (Publisher, error) {
+	if topicResourceID == "" {
+		return nil, nil
```
lukehinds: should we not warn or err if a topicResourceID is blank / not provided?

jalseth: It should not error. Publishing notifications is an optional feature that operators of private instances (or some e2e test scenarios) may not want to enable.

I'm happy to add a warning log, will that work?

pkg/pubsub/pubsub.go
jalseth: (excuse my ignorance, still fairly new to the Rekor internals) What is the value in returning the `entryID` when the `LogIndex` is already returned as a part of `pb.TransparencyLogEntry`? There is an `/api/v1/log/entries?logIndex=<IDX>` which returns the same `'#/definitions/LogEntry'` according to the OpenAPI spec. I have confirmed this is handled using and `GetLogEntryByUUIDHandler` and `GetLogEntryByIndexHandler` respectively in `pkg/api/entries.go`.

bobcallaway: The `LogIndex` value does identify the entry and is used when computing an inclusion proof (we need to know whether there are 0 or 1 siblings of the leaf when computing the proof). Returning the entryID (which contains a shard identifier and the merkle leaf hash) could be considered an optimization, but is useful when verifying other calculations for definitively knowing which shard an entry is in, and ensuring the merkle leaf hash computation matches what is stored by the server.

pkg/pubsub/pubsub.go
bobcallaway: I think you could wrap the code in [pkg/api/api.go](https://github.com/sigstore/rekor/pull/1561/files/472860e355c486235f402ce1e3f050d75e5c325f#diff-67a6e36014be865053c43d38b2091bd2d3f17a91b117d97c0f5244cd7d71127d) with a `viper.IsSet("rekor_server.new_entry_publisher")` check. 

You could then return an error if `topicResourceID` is an empty string as @lukehinds is suggesting. I think "returning an error if `topicResourceID` is empty" is more consistent with the coding style we've used elsewhere in sigstore.

pkg/pubsub/gcp/gcp.go
jalseth: @bobcallaway I have defined a proto for this in https://github.com/sigstore/protobuf-specs/pull/85. I opted to use the JSON format [1] instead of the protobuf format [2] for the initial implementation for better compatibility with PubSub providers (ex AWS SNS) which require JSON for event filtering to work. We can consider also publishing the protobuf format to a second topic in the future if there's demand. Please take a look when you have time.

[1] https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/json-format.md
[2] https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/protobuf-format.md

pkg/pubsub/pubsub.go
jalseth: Thanks for the explanation. I've included entryID and it is used as the unique `id` in the CloudEvents field.

pkg/pubsub/pubsub.go
jalseth: Ack, done.

pkg/pubsub/gcp/gcp.go
Hayden-IO: What do you think about reusing the existing `TransparencyLogEntry` type? This would make each pub/sub message larger, but would prevent diverging message contents and allow for verification of each message (since each message will contain an inclusion proof).

Hayden-IO: Sorry, I missed that this would be encoded in `data`. Could we use the TLE type instead and annotate the type with JSON annotations if needed?

jalseth: @bobcallaway @haydentherapper I wasn't happy with the approach in this PR, so I have rewritten it on a new branch and opened a separate PR:  #1580. Apologies for the churn, but I think the new approach is much better.

jalseth:
Closing in favor of #1580

PR #1580, opened 2023-07-08
feat: Support publishing new log entries to Pub/Sub topics

jalseth:
#### Summary

Adds initial support publishing new log entries to Pub/Sub topics. Interested parties can subscribe to the topic in order to receive notifications when new entries are added.

This relates to #191.  Requires https://github.com/sigstore/protobuf-specs/pull/86 to be merged.

#### Release Note

* New features: Added support publishing new log entries to Pub/Sub topics
* Config changes:
    * Added the `rekor_server.new_entry_publisher` flag to configure the Pub/Sub topic to send notifications to when a new log entry is added. This flag is optional.
    * Addded the `rekor_server.publish_events_protobuf` flag to configure whether events are published using the CloudEvents protobuf format.
    * Addded the `rekor_server.publish_events_json` flag to configure whether events are published using the CloudEvents JSON format.

#### Documentation

This is a TODO still. I would like feedback on the PR before writing the docs.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -93,6 +93,10 @@ func init() {
 Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("rekor_server.signer-passwd", "", "Password to decrypt signer private key")
 
+	rootCmd.PersistentFlags().String("rekor_server.new_entry_publisher", "", "URL for pub/sub queue to send messages to when new entries are added to the log. Ignored if not set. Supported providers: [gcppubsub]")
+	rootCmd.PersistentFlags().Bool("rekor_server.publish_events_protobuf", false, "Whether to publish events in Protobuf wire format. Applies to all enabled event types.")
+	rootCmd.PersistentFlags().Bool("rekor_server.publish_events_json", false, "Whether to publish events in CloudEvents JSON format. Applies to all enabled event types.")
+
 	rootCmd.PersistentFlags().Uint16("port", 3000, "Port to bind to")
 
 	rootCmd.PersistentFlags().Bool("enable_retrieve_api", true, "enables Redis-based index API endpoint")
--- a/go.mod
+++ b/go.mod
@@ -51,6 +51,7 @@ require (
 )
 
 require (
+	cloud.google.com/go/pubsub v1.33.0
 	github.com/AdamKorcz/go-fuzz-headers-1 v0.0.0-20230618160516-e936619f9f18
 	github.com/cyberphone/json-canonicalization v0.0.0-20220623050100-57a0ce2678a7
 	github.com/go-redis/redismock/v9 v9.0.3
@@ -183,7 +184,7 @@ require (
 	golang.org/x/term v0.11.0 // indirect
 	golang.org/x/text v0.12.0 // indirect
 	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
-	google.golang.org/api v0.135.0 // indirect
+	google.golang.org/api v0.135.0
 	google.golang.org/appengine v1.6.7 // indirect
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -31,6 +31,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/pubsub"
 	"github.com/sigstore/rekor/pkg/sharding"
 	"github.com/sigstore/rekor/pkg/signer"
 	"github.com/sigstore/rekor/pkg/storage"
@@ -39,6 +40,8 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
+
+	_ "github.com/sigstore/rekor/pkg/pubsub/gcp" // Load GCP pubsub implementation
 )
 
 func dial(ctx context.Context, rpcServer string) (*grpc.ClientConn, error) {
@@ -63,6 +66,9 @@ type API struct {
 	signer     signature.Signer
 	// stops checkpoint publishing
 	checkpointPublishCancel context.CancelFunc
+	// Publishes notifications when new entries are added to the log. May be
+	// nil if no publisher is configured.
+	newEntryPublisher pubsub.Publisher
 }
 
 func NewAPI(treeID uint) (*API, error) {
@@ -112,6 +118,18 @@ func NewAPI(treeID uint) (*API, error) {
 
 	pubkey := cryptoutils.PEMEncode(cryptoutils.PublicKeyPEMType, b)
 
+	var newEntryPublisher pubsub.Publisher
+	if p := viper.GetString("rekor_server.new_entry_publisher"); p != "" {
+		if !viper.GetBool("rekor_server.publish_events_protobuf") && !viper.GetBool("rekor_server.publish_events_json") {
+			return nil, fmt.Errorf("%q is configured but neither %q or %q are enabled", "new_entry_publisher", "publish_events_protobuf", "publish_events_json")
+		}
+		newEntryPublisher, err = pubsub.Get(ctx, p)
+		if err != nil {
+			return nil, fmt.Errorf("init event publisher: %w", err)
+		}
+		log.ContextLogger(ctx).Infof("Initialized new entry event publisher: %s", p)
+	}
+
 	return &API{
 		// Transparency Log Stuff
 		logClient: logClient,
@@ -121,6 +139,8 @@ func NewAPI(treeID uint) (*API, error) {
 		pubkey:     string(pubkey),
 		pubkeyHash: hex.EncodeToString(pubkeyHashBytes[:]),
 		signer:     rekorSigner,
+		// Utility functionality not required for operation of the core service
+		newEntryPublisher: newEntryPublisher,
 	}, nil
 }
 
@@ -166,4 +186,8 @@ func ConfigureAPI(treeID uint) {
 
 func StopAPI() {
 	api.checkpointPublishCancel()
+
+	if api.newEntryPublisher != nil {
+		api.newEntryPublisher.Close()
+	}
 }
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -36,10 +36,14 @@ import (
 	"google.golang.org/genproto/googleapis/rpc/code"
 	"google.golang.org/grpc/codes"
 
+	"github.com/sigstore/rekor/pkg/events"
+	"github.com/sigstore/rekor/pkg/events/newentry"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/generated/restapi/operations/entries"
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/pubsub"
 	"github.com/sigstore/rekor/pkg/sharding"
+	"github.com/sigstore/rekor/pkg/tle"
 	"github.com/sigstore/rekor/pkg/trillianclient"
 	"github.com/sigstore/rekor/pkg/types"
 	"github.com/sigstore/rekor/pkg/util"
@@ -290,7 +294,7 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 		RootHash:   swag.String(hex.EncodeToString(root.RootHash)),
 		LogIndex:   swag.Int64(queuedLeaf.LeafIndex),
 		Hashes:     hashes,
-		Checkpoint: stringPointer(string(scBytes)),
+		Checkpoint: swag.String(string(scBytes)),
 	}
 
 	logEntryAnon.Verification = &models.LogEntryAnonVerification{
@@ -301,7 +305,64 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	logEntry := models.LogEntry{
 		entryID: logEntryAnon,
 	}
+
+	if api.newEntryPublisher != nil {
+		// Publishing notifications should not block the API response.
+		go func() {
+			verifiers, err := entry.Verifiers()
+			if err != nil {
+				incPublishEvent(newentry.Name, "", false)
+				log.ContextLogger(ctx).Errorf("Could not get verifiers for log entry %s: %v", entryID, err)
+				return
+			}
+			var subjects []string
+			for _, v := range verifiers {
+				subjects = append(subjects, v.Subjects()...)
+			}
+
+			pbEntry, err := tle.GenerateTransparencyLogEntry(logEntryAnon)
+			if err != nil {
+				incPublishEvent(newentry.Name, "", false)
+				log.ContextLogger(ctx).Error(err)
+				return
+			}
+			event, err := newentry.New(entryID, pbEntry, subjects)
+			if err != nil {
+				incPublishEvent(newentry.Name, "", false)
+				log.ContextLogger(ctx).Error(err)
+				return
+			}
+			if viper.GetBool("rekor_server.publish_events_protobuf") {
+				go publishEvent(ctx, api.newEntryPublisher, event, events.ContentTypeProtobuf)
+			}
+			if viper.GetBool("rekor_server.publish_events_json") {
+				go publishEvent(ctx, api.newEntryPublisher, event, events.ContentTypeJSON)
+			}
+		}()
+	}
+
 	return logEntry, nil
+}
+
+func publishEvent(ctx context.Context, publisher pubsub.Publisher, event *events.Event, contentType events.EventContentType) {
+	err := publisher.Publish(context.WithoutCancel(ctx), event, contentType)
+	incPublishEvent(event.Type().Name(), contentType, err == nil)
+	if err != nil {
+		log.ContextLogger(ctx).Error(err)
+	}
+}
+
+func incPublishEvent(event string, contentType events.EventContentType, ok bool) {
+	status := "SUCCESS"
+	if !ok {
+		status = "ERROR"
+	}
+	labels := map[string]string{
+		"event":        event,
+		"status":       status,
+		"content_type": string(contentType),
+	}
+	metricPublishEvents.With(labels).Inc()
 }
 
 // CreateLogEntryHandler creates new entry into log
--- a/pkg/events/newentry/new_entry.go
+++ b/pkg/events/newentry/new_entry.go
@@ -1,0 +1,47 @@
+// Copyright 2023 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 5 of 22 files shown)


jalseth:
@bobcallaway Please note the build will fail as-is due to dependency on https://github.com/sigstore/protobuf-specs/pull/86 which hasn't been merged yet. All tests pass running locally when using the not-yet-merged dependency: https://gist.github.com/jalseth/ebf866def89661d31fd87c93eed80ed2

cmd/rekor-server/app/root.go
```diff
@@ -93,6 +93,10 @@ func init() {
 Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("rekor_server.signer-passwd", "", "Password to decrypt signer private key")
 
+	rootCmd.PersistentFlags().String("rekor_server.new_entry_publisher", "", "URL for pub/sub queue to send messages to when new entries are added to the log. Ignored if not set.")
+	rootCmd.PersistentFlags().Bool("rekor_server.publish_new_entry_protobuf", false, "Whether to publish events for new log entries in Protobuf wire format. Requires new_entry_publisher to be configured.")
```
Hayden-IO: Can we allow either only protobuf or json, rather than both? Otherwise if both are configured, a monitor may think that an identity has shown up twice. Is there a use case where both would need to be enabled?

pkg/api/api.go
```diff
@@ -39,6 +40,8 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
+
+	_ "github.com/sigstore/rekor/pkg/pubsub/gcp" // Load GCP pubsub implementation
```
Hayden-IO, line 44: Any downside to always importing this even if pubsub is not enabled?

cmd/rekor-server/app/root.go
```diff
@@ -93,6 +93,10 @@ func init() {
 Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("rekor_server.signer-passwd", "", "Password to decrypt signer private key")
 
+	rootCmd.PersistentFlags().String("rekor_server.new_entry_publisher", "", "URL for pub/sub queue to send messages to when new entries are added to the log. Ignored if not set.")
```
Hayden-IO: Should we document that only GCP is supported currently?

pkg/api/entries.go
```diff
@@ -301,9 +306,64 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 	logEntry := models.LogEntry{
 		entryID: logEntryAnon,
 	}
+
+	if api.newEntryPublisher != nil {
+		// Publishing notifications should not block the API response.
```
Hayden-IO, line 310: We should think through how we'll detect errors. We've missed failed Redis requests before. We'll also need a script to handle backfills.

FWIW, even if this were blocking, we'd still need to handle backfills if the request to Trillian succeeds but the topic write fails.

@bobcallaway Thoughts on blocking vs non-blocking?

pkg/api/entries.go
```diff
@@ -301,9 +306,64 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
...
+				return
+			}
+			event, err := newentry.New(entryID, pbEntry, subjects)
+			if err != nil {
+				incPublishEvent(newentry.Name, "", false)
+				log.ContextLogger(ctx).Error(err)
+				return
+			}
+			if viper.GetBool("rekor_server.publish_new_entry_protobuf") {
+				go publishEvent(ctx, api.newEntryPublisher, event, events.ContentTypeProtobuf)
```
Hayden-IO, line 336: Given this is already in a go routine, do we need another?

pkg/api/entries.go
```diff
@@ -301,9 +306,64 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
...
 	return logEntry, nil
 }
 
+func publishEvent(ctx context.Context, publisher pubsub.Publisher, event *events.Event, contentType events.EventContentType) {
+	err := publisher.Publish(context.Background(), event, contentType)
+	incPublishEvent(event.Type().Name(), contentType, err == nil)
+	if err != nil {
+		log.ContextLogger(ctx).Error(err)
+	} else {
+		log.ContextLogger(ctx).Debugf("Published new entry event (%s): %s", contentType, prototext.Format(event.Message()))
```
Hayden-IO: I'd probably omit this, given it'll be very frequent and we shouldn't log the entire message.

pkg/events/types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+var (
+	registeredEventTypes = map[string]*EventType{}
+	mu                   sync.Mutex
+)
+
+// EventType describes the type of an event.
```
Hayden-IO, line 29: For context, is the design here so that we can support multiple event types in the future?

pkg/events/newentry/new_entry.go
```diff
@@ -0,0 +1,47 @@
...
+func init() {
+	empty := &rekor_pb.TransparencyLogEntry{}
+	ty = events.RegisterType(Name, Source, empty.ProtoReflect().Descriptor())
+}
+
+func New(id string, entry *rekor_pb.TransparencyLogEntry, subjects []string) (*events.Event, error) {
+	slices.Sort(subjects) // Must be sorted for consistency.
+	attrs := map[string]any{
+		"time":                   time.Unix(entry.GetIntegratedTime(), 0),
+		"rekor_entry_kind":       entry.GetKindVersion().GetKind(),
```
Hayden-IO, line 43: We'd also need version, since kind+version tells the caller how to parse the entry.

pkg/tle/tle.go
```diff
@@ -50,12 +50,22 @@ func GenerateTransparencyLogEntry(anon models.LogEntryAnon) (*rekor_pb.Transpare
 		inclusionProofHashes[i] = hashBytes
 	}
 
-	b, err := base64.StdEncoding.DecodeString(anon.Body.(string))
-	if err != nil {
-		return nil, fmt.Errorf("base64 decoding body: %w", err)
+	// Different call paths may supply string or []byte. If string, it is base64 encoded.
+	var body []byte
+	switch x := anon.Body.(type) {
```
Hayden-IO: nit, `t` rather than `x` for var name

pkg/tle/tle.go
```diff
@@ -50,12 +50,22 @@ func GenerateTransparencyLogEntry(anon models.LogEntryAnon) (*rekor_pb.Transpare
 		inclusionProofHashes[i] = hashBytes
 	}
 
-	b, err := base64.StdEncoding.DecodeString(anon.Body.(string))
-	if err != nil {
-		return nil, fmt.Errorf("base64 decoding body: %w", err)
+	// Different call paths may supply string or []byte. If string, it is base64 encoded.
```
Hayden-IO, line 53: I wonder if we should instead require a type here. Not a strong preference though. @bobcallaway?

Hayden-IO (commented):
Fantastic work on this!

pkg/events/types.go
jalseth: Yes, and not just for Rekor. Currently this package exists in Rekor only for convenience in making this change, but if for example Fulcio wanted to publish events in the future I imagine we'd move this package to a common library for reuse by all Sigstore projects.

pkg/events/newentry/new_entry.go
jalseth: The event includes the full TLE regardless, which must be deserialized to validate the entry anyway, and at that point you can get the version for the entry. This is extra metadata in the scope of the event that can be useful for filtering events before deserialization. I can imagine use cases for filtering on entry kind, but I don't imagine someone would need to filter on the kind's specific version.

All metadata we add here increases the throughput costs for publishing to PubSub (we pay for each byte) , so I think we should keep this lean until there's a FR to add additional fields.

pkg/api/entries.go
jalseth: Woops, forgot to remove that after fixing an issue when running the e2e tests.

pkg/api/entries.go
jalseth: Given the publish timeout is 60s, the intent was to publish as close to real-time as possible to give consumers as much of a heads up as possible if they are using these events to detect abnormal entries given that we can't make assumptions about which encoding consumers will use. In a practical sense a maximum delay of ~60s before getting to the JSON publishing probably doesn't really matter so I'm fine with removing these.

cmd/rekor-server/app/root.go
jalseth: The intent was to allow consumer flexibility. GCP Pub/Sub supports server-side filtering, and I was expecting each consumer to filter on the `datacontenttype` (mandatory field in the CloudEvents spec) of `application/protobuf` or `application/json`.

Thinking about this more, these should probably just be called `publish_events_{protobuf,json}` to apply to all events published by Rekor.

pkg/tle/tle.go
jalseth: I disagree. Assuming you mean `t` for `type`, the variable is not a specific type indicator, rather it is a value of a certain type. If you prefer, I can switch to `v` rather than `x` as I've done elsewhere in the PR.

pkg/api/api.go
jalseth: Nothing substantial. A small amount of memory will be used for each loaded PubSub provider.

cmd/rekor-server/app/root.go
jalseth: Sure, done.

pkg/tle/tle.go
Hayden-IO: Yep, you're correct, I misread this. Let's switch this to `v` though.

pkg/events/newentry/new_entry.go
Hayden-IO: I wonder if we should remove entry kind, to discourage filtering only by a kind that the searcher is aware of. For example, if an attacker steals credentials and knows that someone is filtering by a specific kind, they could upload a signed package for a different kind.

pkg/api/api.go
```diff
@@ -165,5 +185,8 @@ func ConfigureAPI(treeID uint) {
 }
 
 func StopAPI() {
+	if api.newEntryPublisher != nil {
+		api.newEntryPublisher.Close()
```
Hayden-IO, line 191: Does this wait until all transactions are complete?

jalseth: It would be up to each publisher implementation. The current GCP implementation does not wait.

pkg/events/newentry/new_entry.go
jalseth: With GCP PubSub, the cost savings can happen on both ends. Messages filtered server-sided are not sent to the client, and thus the consumer is not billed for the PubSub throughput (per byte again). The intent is to allow consumers to save cost by ignoring events they don't care about. I agree this could bite people, so the docs should have a large warning regarding this.

pkg/api/api.go
Hayden-IO: Can we add waiting? Minimizes the risk of currently processing messages from not being sent.

pkg/api/entries.go
jalseth: Having these start in different Go routines is now important to have the RLocks() called as soon as possible to help with safe shutdown.

pkg/api/api.go
jalseth: Done

cmd/rekor-server/app/root.go
```diff
@@ -93,6 +93,10 @@ func init() {
 Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("rekor_server.signer-passwd", "", "Password to decrypt signer private key")
 
+	rootCmd.PersistentFlags().String("rekor_server.new_entry_publisher", "", "URL for pub/sub queue to send messages to when new entries are added to the log. Ignored if not set. Supported providers: [gcppubsub]")
+	rootCmd.PersistentFlags().Bool("rekor_server.publish_events_protobuf", false, "Whether to publish events in Protobuf wire format. Applies to all enabled event types.")
+	rootCmd.PersistentFlags().Bool("rekor_server.publish_events_json", false, "Whether to publish events in CloudEvents JSON format. Applies to all enabled event types.")
```
bobcallaway, line 98: should this be set to a single flag with an enum value?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor
 
-go 1.20
+go 1.21
```
bobcallaway: do we need to go to 1.21 for this?

pkg/api/entries.go
```diff
@@ -301,9 +305,64 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
...
 		entryID: logEntryAnon,
 	}
+
+	if api.newEntryPublisher != nil {
+		// Publishing notifications should not block the API response.
+		go func() {
+			var subjects []string
+			verifiers, err := entry.Verifiers()
+			if err != nil {
+				log.ContextLogger(ctx).Warnf("Could not get verifier for log entry %s: %w", entryID, err)
```
bobcallaway: If you can't get the Verifiers for an entry, that feels like more than just a warning to me. Curious to get @haydentherapper 's opinion.

pkg/pubsub/gcp/publisher.go
```diff
@@ -0,0 +1,154 @@
...
+
+	msg := &pubsub.Message{
+		Data:       data,
+		Attributes: gcpAttrs(event, encoding),
+	}
+	res := p.client.Topic(p.topic).Publish(ctx, msg)
+	// TODO: Consider making the timeout configurable.
+	cctx, cancel := context.WithTimeout(ctx, pubsub.DefaultPublishSettings.Timeout)
+	defer cancel()
+	if _, err := res.Get(cctx); err != nil {
```
bobcallaway, line 116: nit: add a comment that explains that `res` is effectively a future, and will block until the prior `Publish` command succeeds?

pkg/tle/tle.go
bobcallaway: what call paths send a `[]byte`? I think that's more likely a bug we need to fix

bobcallaway (commented):
lgtm overall, a couple questions

pkg/tle/tle.go
jalseth: The new call path from `createLogEntry` in `pkg/api/entries.go`. The `Body` is set to `leaf_value` from the Trillian response which is defined as `[]byte` in the proto https://github.com/google/trillian/blob/6496dffd1199c5423f9741cb80e162b97c51c6ff/trillian_log_api.proto#L310

If the source is `[]byte` I don't see a purpose to b64 encoding it to string, just to have it decoded back to a byte slice.

go.mod
Hayden-IO: +1, let's not bump the Go version unless there's a 1.21 feature you're adding.

pkg/api/entries.go
Hayden-IO: I'm good with marking this as resolved, as the GCP Datastream approach will solve this.

pkg/api/entries.go
Hayden-IO: Agreed - It would mean that the entry or verifier is malformed, since it should always contain a valid verifier. We should log an error/metric and return.

pkg/events/newentry/new_entry.go
```diff
@@ -0,0 +1,47 @@
...
+		"rekor_signing_subjects": strings.Join(subjects, ","),
```
Hayden-IO, line 44: Is there any concept of string arrays, so that clients don't have to know subject delimiters?

pkg/events/newentry/new_entry_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	testCases := []struct {
+		desc     string
+		entryID  string
+		subjects []string
+		entry    *rekor_pb.TransparencyLogEntry
+
+		wantError bool
+		want      *events_pb.CloudEvent
```
Hayden-IO, line 55: Test for multiple subjects?

pkg/events/types.go
```diff
@@ -0,0 +1,100 @@
...
+		panic("event name must be set")
+	}
+	if source == "" {
+		panic("event source must be set")
+	}
+	if desc == nil {
+		panic("event descriptor must be set")
+	}
+	if _, ok := registeredEventTypes[name]; ok {
+		panic("event has already been registered: " + name)
```
Hayden-IO, line 67: Instead of a panic, should this be a warning? Seems low impact to try to register a type twice.

Hayden-IO (commented):
LGTM with just a few nits

go.mod
jalseth: The `maps` and `slices` packages were added to stdlib in Go 1.21, which this PR uses so I bumped this. Technically, they're still available under `golang.org/x/exp/{maps,slices}` which we can continue to use, I just need to fix up the imports if that is preferred.

pkg/pubsub/gcp/publisher.go
jalseth: Done, though `Publish()` itself does not block, the `Get` on the `PublishResult` (`res`) blocks. We need to call `Get()` so that we can know if the Publish call encountered and error or not.

cmd/rekor-server/app/root.go
jalseth: Please see https://github.com/sigstore/rekor/pull/1580#discussion_r1275533835. The intent was for the public Rekor instance to publish both, and let clients filter based on the format that works for them.

pkg/events/newentry/new_entry.go
jalseth: @haydentherapper Any further thoughts on this?

pkg/events/newentry/new_entry.go
jalseth: Technically, it would depend on the PubSub provider, but neither GCP or AWS support this. GCP is strictly limited to `map[string]string`.

pkg/api/entries.go
jalseth: Changed to Errorf

pkg/events/newentry/new_entry_test.go
jalseth: Done.

pkg/events/types.go
jalseth: If for some reason two callers were registering the same name but different types, it could lead to unexpected runtime errors. I think it is better to panic on start-up initialization in this case, as this is intended to be called from `init()` functions.

go.mod
Hayden-IO: I would use `x/exp` for now, I'd prefer to bump the go version in its own PR.

pkg/events/newentry/new_entry.go
Hayden-IO: I'm fine with keeping this with appropriate documentation.

pkg/events/newentry/new_entry.go
Hayden-IO: Hm, this might be an issue for filtering then, because there's no filter for "contains", only "hasPrefix", at least for GCP - https://cloud.google.com/pubsub/docs/subscription-message-filter#filtering_syntax

The good news is that for the common use case, a certificate should only contain a single subject. I found an issue tracking support for regex in filters, but there's been no progress for awhile, so this might just have to be a documented limitation unless you have any other suggestions.

jalseth: Yeah, it is an unfortunate limitation. I hope they can add `contains` at some point given that the max length of an attribute is 1024 bytes https://cloud.google.com/pubsub/quotas#resource_limits.

pkg/events/types.go
Hayden-IO: Thanks for the explanation, I'm good with keeping this as is

bobcallaway (approved):

Hayden-IO (approved):

pkg/tle/tle.go
jalseth: Discussed offline. We will leave this PR as-is for now and will follow up on this.

cmd/rekor-server/app/root.go
jalseth: Discussed offline

PR #1599, opened 2023-07-26
Use correct log index in inclusion proof

cdris:
#### Summary

Swaps to using the inclusion proof log index instead of the entry log index in the inclusion proof struct.


pkg/tle/tle.go
```diff
@@ -78,7 +78,7 @@ func GenerateTransparencyLogEntry(anon models.LogEntryAnon) (*rekor_pb.Transpare
 			SignedEntryTimestamp: anon.Verification.SignedEntryTimestamp,
 		},
 		InclusionProof: &rekor_pb.InclusionProof{
-			LogIndex: *anon.LogIndex,
+			LogIndex: *anon.Verification.InclusionProof.LogIndex,
```
bobcallaway: can you please add a comment to denote that the inclusion proof is relative to the specific shard, and therefore the value here might not match the global index on line 68

Hayden-IO (approved):

bobcallaway (approved):

PR #1611, opened 2023-08-03
Refactor Identities API

Hayden-IO:
Instead of providing raw strings that are up to the client to determine how to parse, return typed structs. Encoded keys are included and should be parsed based on the returned type.


Hayden-IO:
FYI: TUF requires a new root because go-tuf dropped support for hex-encoded keys, as they were not conformant to the TUF spec.

pkg/pki/tuf/tuf.go
```diff
@@ -119,6 +123,7 @@ func NewPublicKey(r io.Reader) (*PublicKey, error) {
 	db := verify.NewDB()
 	for id, k := range root.Keys {
 		if err := db.AddKey(id, k); err != nil {
+			fmt.Println("potato")
```
bobcallaway: left in by mistake?

bobcallaway (dismissed):
lgtm minus one errant 🥔 

also, since this changes a public API surface we should probably bump into a new minor release.

pkg/pki/tuf/tuf.go
Hayden-IO: I was hungry

Hayden-IO:
@bobcallaway Agreed on the minor bump.

bobcallaway (approved):

PR #1643, opened 2023-08-23
Add SECURITY.md file

david-a-wheeler:
The repo already clearly identifies how to report vulnerabilities. However, some people just look for a SECURITY.md file.

This commit adds a SECURITY.md file, with the same material as in the README, to increase the likelihood that panicked reporters will find the information they need.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,4 @@
+# Security
+
+Should you discover any security issues, please refer to sigstore's [security
+process](https://github.com/sigstore/.github/blob/main/SECURITY.md)


Hayden-IO (approved):

Hayden-IO:
Just realized that this isn't ideal because it overrides what's specified in https://github.com/sigstore/rekor/security/policy (see https://github.com/sigstore/cosign/security/policy for example)

david-a-wheeler:
I linked to: https://github.com/sigstore/.github/blob/main/SECURITY.md

Sounds like it should link to this instead: https://github.com/sigstore/rekor/security/policy

Correct? I'm happy to change the URL. I just want to make sure that people who look for SECURITY.md will find it.

Hayden-IO:
I'm not sure if that works. By specifying a security.md file, you override the policy set by the org.

david-a-wheeler:
Can we copy the policy contents into the SECURITY.md file, then?

The content would then appear in both places.

Hayden-IO:
I'd like to avoid that since it'll get out of date (eg if we update who's on the src, or the contact email). This is currently working as intended, to set one policy for the org.

Hayden-IO:
Going to close this - We want to avoid each repo setting its own security policy. If we find that users are unable to find the policy through GitHub's security policy link, we can revisit this.

PR #1682, opened 2023-09-14
Update .ko.yaml

loosebazooka:
Use debian12 base image

Open questions?
- [ ] do we need libssl? if not we should use `base-nossl-debian12:debug-nonroot`
- [ ] what's all this redirecting piping stuff going on? is that comment still relevant?

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # We need a shell for a lot of redirection/piping to work
-defaultBaseImage: gcr.io/distroless/base:debug-nonroot
+defaultBaseImage: gcr.io/distroless/base-debian12:debug-nonroot
 
 builds:
 - id: rekor-server
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -13,8 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# We need a shell for a lot of redirection/piping to work
-defaultBaseImage: gcr.io/distroless/base-debian12:debug-nonroot
+defaultBaseImage: gcr.io/distroless/static-debian12:nonroot
 
 builds:
 - id: rekor-server


loosebazooka:
cc: @jku

Hayden-IO:
I don't believe libssl is needed, as golang implements its own crypto and we don't use boringssl.

Hayden-IO:
Let's switch to static since we shouldn't need anything else.

.ko.yaml
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # We need a shell for a lot of redirection/piping to work
-defaultBaseImage: gcr.io/distroless/base:debug-nonroot
+defaultBaseImage: gcr.io/distroless/base-debian12:debug-nonroot
```
Hayden-IO: `gcr.io/distroless/static-debian12:nonroot` I think should work? We shouldn't need a shell.

jku: What about the comment above?

If the shell was only needed when entering the it for debugging, I imagine the comment would mention that instead... Is the comment obsolete?

Hayden-IO: good point, lemme test this locally. I’m not sure what is using the shell, any guesses @bobcallaway?

loosebazooka: @cpanato maybe knows?

Hayden-IO: Using this as the base image, seems like everything is working - I can start the log, upload and fetch an entry. 

Maybe the CLI needs a shell? Then we could just specify different base images if so.

bobcallaway: this is a copypasta job from cosign: https://github.com/sigstore/cosign/commit/8974bdc695ee5ce6ca5356c54101c829d144b23c

bobcallaway: not relevant for rekor IMO

Hayden-IO: OK, I'll follow up on the Cosign thread to double check if we need the shell. Just updated the base image, PTAL

bobcallaway (approved):

PR #1758, opened 2023-10-11
changelog for v1.3.2

bobcallaway:


cpanato (approved):
we might want to drop the dependabot bumps

PR #1838, opened 2023-11-20
update indexstorage interface to reduce roundtrips

bobcallaway:
This changes the interface for `indexstorage` in two ways:

1. Adds a `Shutdown()` method to clean up resources upon server shutdown
2. Takes `[]string` on both read and write paths

The redis implementation now pipelines requests into a single roundtrip between Rekor and Redis on both the read and write path (where possible).

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -195,7 +195,9 @@ func StopAPI() {
 	api.checkpointPublishCancel()
 
 	if api.newEntryPublisher != nil {
-		api.newEntryPublisher.Close()
+		if err := api.newEntryPublisher.Close(); err != nil {
+			log.Logger.Errorf("shutting down newEntryPublisher: %v", err)
+		}
 	}
 
 	if indexStorageClient != nil {


cpanato (dismissed):
lgtm

pkg/api/index.go
```diff
@@ -42,14 +42,20 @@ func SearchIndexHandler(params index.SearchIndexParams) middleware.Responder {
...
+	var lookupKeys []string
+
 	if params.Query.Hash != "" {
 		// This must be a valid hash
-		sha := util.PrefixSHA(params.Query.Hash)
-		resultUUIDs, err := indexStorageClient.LookupIndices(httpReqCtx, strings.ToLower(sha))
-		if err != nil {
-			return handleRekorAPIError(params, http.StatusInternalServerError, fmt.Errorf("index storage error: %w", err), indexStorageUnexpectedResult)
+		sha := strings.ToLower(util.PrefixSHA(params.Query.Hash))
+		if queryOperator == "or" {
```
Hayden-IO: Does "and" need to be handled too? I think the lookups would be the same for or vs and, it seems like NewCollection just processes them different I think

pkg/indexstorage/redis/redis_test.go
```diff
@@ -26,14 +26,14 @@ import (
 )
```
Hayden-IO: Can we add a test to confirm read/write works with multiple values?

pkg/api/index.go
bobcallaway: It's handled in the else clause on line 52. The way the Collector was implemented is that you have to pass it in the queries for each input parameter separately, where as OR is just a concatenation.

pkg/indexstorage/redis/redis_test.go
bobcallaway: The e2e tests already cover this

Hayden-IO (dismissed):

Hayden-IO (approved):

PR #1843, opened 2023-11-21
add mysql indexstorage backend

bobcallaway:
Adds a MySQL indexstorage backend as an alternative to Redis. 

This is built on top of https://github.com/sigstore/rekor/pull/1838

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -38,12 +38,15 @@ services:
       "--max_request_body_size=32792576",
       "--rekor_server.new_entry_publisher=gcppubsub://projects/test-project/topics/new-entry",
       "--rekor_server.publish_events_json=true",
+      "--search_index.storage_provider=mysql",
+      "--search_index.mysql.dsn=test:zaphod@tcp(mysql:3306)/test",
       ]
     ports:
       - "3000:3000"
       - "2112:2112"
     depends_on:
       - gcp-pubsub-emulator
+      - mysql
   gcp-pubsub-emulator:
     image: gcp-pubsub-emulator
     ports:
--- a/pkg/indexstorage/mysql/mysql.go
+++ b/pkg/indexstorage/mysql/mysql.go
@@ -30,14 +30,14 @@ import (
 const (
 	ProviderType = "mysql"
 
-	lookupStmt      = "SELECT EntryUUIDs FROM EntryIndex WHERE EntryKey IN (?)"
-	writeStmt       = "INSERT IGNORE INTO EntryIndex (EntryKey, EntryUUIDs) VALUES (:key, :uuid)"
+	lookupStmt      = "SELECT EntryUUID FROM EntryIndex WHERE EntryKey IN (?)"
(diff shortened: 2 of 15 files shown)


docker-compose.test.yml
```diff
@@ -38,12 +38,15 @@ services:
...
       "--rekor_server.publish_events_json=true",
+      "--search_index.storage_provider=mysql",
+      "--search_index.mysql.dsn=test:zaphod@tcp(mysql:3306)/test",
       ]
     ports:
       - "3000:3000"
       - "2112:2112"
     depends_on:
       - gcp-pubsub-emulator
+      - mysql
```
cpanato, line 49: where we define the mysql? maybe a dumb question 🤔

cpanato: ahh, i see, was a dumb question, please ignore

cpanato (dismissed):
lgtm

pkg/indexstorage/mysql/mysql.go
```diff
@@ -0,0 +1,134 @@
...
+		UNIQUE(EntryKey, EntryUUIDs)
+	)`
+)
+
+// IndexStorageProvider implements indexstorage.IndexStorage
+type IndexStorageProvider struct {
+	db *sqlx.DB
+}
+
+func NewProvider(dsn string) (*IndexStorageProvider, error) {
```
Hayden-IO: For a slightly cleaner interface, should this take an options struct so that this doesn't have to depend on viper?

pkg/indexstorage/mysql/mysql.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// this imports the mysql driver for go
+	_ "github.com/go-sql-driver/mysql"
+)
+
+const (
+	ProviderType = "mysql"
+
+	lookupStmt      = "SELECT EntryUUIDs FROM EntryIndex WHERE EntryKey IN (?)"
+	writeStmt       = "INSERT IGNORE INTO EntryIndex (EntryKey, EntryUUIDs) VALUES (:key, :uuid)"
```
Hayden-IO: JW, is IGNORE there to make this idempotent?

pkg/indexstorage/mysql/mysql.go
```diff
@@ -0,0 +1,134 @@
...
+	createTableStmt = `CREATE TABLE IF NOT EXISTS EntryIndex (
+		PK BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
+		EntryKey varchar(512) NOT NULL,
+		EntryUUIDs char(80) NOT NULL,
```
Hayden-IO: Should this be singular, since it's not repeated?

pkg/indexstorage/mysql/mysql.go
bobcallaway: yes - if we needed to backfill, i didn't want the insertion failure to throw an error.

pkg/indexstorage/mysql/mysql.go
bobcallaway: yes

Hayden-IO (approved):

cpanato (approved):

PR #1852, opened 2023-11-26
Add YAML tags to match the support indicated in the documentation

nkreiger:
#### Summary

Support is indicated for YAML, and the code attempts to do yaml first, however, based on the tags, only JSON is supported.

#### Release Note

(fix) add yaml tag

#### Documentation

No

### DIFF
--- a/pkg/sharding/ranges.go
+++ b/pkg/sharding/ranges.go
@@ -39,9 +39,9 @@ type LogRanges struct {
 type Ranges []LogRange
 
 type LogRange struct {
-	TreeID           int64  `json:"treeID"`
-	TreeLength       int64  `json:"treeLength"`
-	EncodedPublicKey string `json:"encodedPublicKey"`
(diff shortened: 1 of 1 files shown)


bobcallaway (approved):
the change is fine, but I can't seem to find any references to this in our documentation. Could you share a pointer?

nkreiger:
> the change is fine, but I can't seem to find any references to this in our documentation. Could you share a pointer?

@bobcallaway 

https://docs.sigstore.dev/logging/sharding/

specifically:

```
- treeID: $CURRENT_TREE_ID
  treeLength: $CURRENT_SHARD_LENGTH
  ```
 
 Is what I was following when I encountered this

bobcallaway:
ah, ok. We don't ever marshal this struct directly as YAML, so I think this fix is just effectively a documentation improvement IIUC

PR #1853, opened 2023-11-26
fix: Do not check for pubsub.topics.get on initialization

jalseth:
This permission is not strictly needed and is not included in the `pubsub.publisher` role.

https://cloud.google.com/pubsub/docs/access-control#pubsub.publisher


cpanato:
Dont you need that one to initialize the client?

bobcallaway (approved):

PR #1888, opened 2023-12-07
Remove timestamp from checkpoint

### LINKED ISSUE #1887: Remove timestamp from checkpoint
**Description**

We currently include the timestamp when the checkpoint was generated:

```
rekor.sigstore.dev - 3904496407287907110
4163431
TQBqpG78tgfdUdkAsSE3VMUMySUcNAXGwlYdnWovMjk=
Timestamp: 1701981305360721342

— rekor.sigstore.dev wNI9ajBFAiEA8v7iWeGsxEL5ysGdi1A2vu9lMSuCeRtSBOXzRCK+5ygCIBe0r5Au5PmbY1lbJv+1/x2KoLm/YPsb1X5JyY3OXUHH
```

Note the first line is the log identifier, the second is the log size, then the root hash, then extra data (the timestamp), then a new line and then the checkpoint signature.

No checkpoint verifier relies on the timestamp. It could be used to convey checkpoint freshness, that the checkpoint represents the latest snapshot of the log. However:
* this is not verifiable in any way
* if the log presented different views to different callers, say if the log wanted to act like it was "frozen" to one caller and serve an old checkpoint, that's mitigated through witnessing, not through checkpoint timestamps
* if a verifier were to require it be fresh, it means we cannot cache static checkpoints for old log shards.

I propose that we remove the timestamp from the checkpoint. We'll need to confirm that this is not a breaking change, that old clients can verify checkpoints without the timestamp present.

cc @mhutchinson

Hayden-IO:
Fixes https://github.com/sigstore/rekor/issues/1887. Verified that checkpoints are still verifiable with and without timestamps (since timestamps are just a part of the existing OtherContent, this is not a breaking change).

Looking over all of the Sigstore org, no project is relying on the timestamp.

### DIFF
--- a/pkg/util/checkpoint_test.go
+++ b/pkg/util/checkpoint_test.go
@@ -27,7 +27,6 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"testing"
-	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/sigstore/sigstore/pkg/signature"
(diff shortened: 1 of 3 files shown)


pkg/util/checkpoint_test.go
```diff
@@ -283,8 +310,8 @@ func TestSigningRoundtripCheckpoint(t *testing.T) {
 			if err != nil {
 				t.Fatalf("error creating signed checkpoint")
 			}
-			time := uint64(time.Now().UnixNano())
-			sth.SetTimestamp(time)
+			// time := uint64(time.Now().UnixNano())
+			// sth.SetTimestamp(time)
```
bobcallaway: should we just delete this?

bobcallaway (dismissed):

pkg/util/checkpoint_test.go
Hayden-IO: whoops, deleted

bobcallaway (approved):

mitar:
This seems to break node LTS 20 which comes with an older npm which does not have a change to account for this change. So it is breaking.

See:
* https://github.com/npm/cli/issues/7279
* https://github.com/nodemailer/nodemailer/issues/1634
* https://github.com/microsoft/playwright/issues/29798

PR #1929, opened 2023-12-21
Closes #1928 Allow for attestation upload for DSSE types

### LINKED ISSUE #1928: Allow attestation upload for DSSE type similar to the intoto type
As discussed in [Slack](https://sigstore.slack.com/archives/C01PZKDL4DP/p1702653623025059), we would like to have attestations uploaded to our attestation store (s3) similar to how the intoto type behaves. This is all because of the choice of cosign to abandon the upload of the intoto type and only have the dsse type.
This implementation does not break existing functionality so I don't think it requires reversioning of the DSSE type. This also still allows prevention of attestation upload in the public rekor instance.
I've gone ahead and raised a PR, but am open to any suggestions.

aalsabag:
#### Summary
This allows for attestation uploads to S3 if `--enable_attestation_storage` is set. This will now allow those using cosign to upload the attestations despite them being of type DSSE. Closes #1928 

#### Release Note
The DSSE type implements AttestationKey() and AttestationKeyValue()

#### Documentation
NONE


bobcallaway:
please use the intoto v0.0.2 type for this, the DSSE type by definition will not have the attestation attached.

aalsabag:
@bobcallaway so cosign has been updated to only support the DSSE type. I figured the approach above was better than updating cosign. My initial approach was to just read the type in cosign, and then have a conditional that uploads a dsse type or intoto type accordingly. I assume that's what you are suggesting as opposed to this?

bobcallaway:
yup, a flag in cosign (or other clients) to pick the attestation type is the right way to do this in my opinion.

Hayden-IO:
Note that we are likely to deprecate the intoto types in the near future as Rekor is not meant to be public storage for attestations. A flag would be fine in Cosign, but that would be for private deployments of the log.

aalsabag:
Ok I just implemented the [change in cosign](https://github.com/aalsabag/cosign/commit/f2e57b361b32deddfd43bef1920f2ef9e6d1a502) to check the flag, but I must say I'm a little confused again @haydentherapper. If we are deprecating the intoto type, then is this solution also not appropriate?

It is definitely the simpler solution than what I initially suggested, but my goal with the original solution was to still allow for backward compatability with the dsse type. From what I understand, the DSSE envelope is [not as rigid](https://github.com/secure-systems-lab/dsse/blob/master/envelope.md) as I had originally thought so long as it has a few required fields. Public rekor can still have `--enable-attestation-storage` set to false and people who use private instances would still have the ability to store attestations.

As I go through the slack channel, I see folks (running private instances) come up with strange solutions to store their attestations, but the functionality is there and it's all built in to the existing stack!

I'm good with either solution or any other suggestions for that matter, just need some clarification on what direction to head. I'm probably missing something dumb here so forgive my ignorance.

bobcallaway:
> Note that we are likely to deprecate the intoto types in the near future as Rekor is not meant to be public storage for attestations. 

@haydentherapper is referring to the public instance that we're running at rekor.sigstore.dev

aalsabag:
@bobcallaway ok I get that. 
Are you guys ok with [this basic change in cosign](https://github.com/aalsabag/cosign/commit/f2e57b361b32deddfd43bef1920f2ef9e6d1a502)? What would you like me to add to it?

I can add an additional conditional for the `rekor-server` to check if it's the public one.

Obviously I'll add this change to both `attest` and `attest-blob`

aalsabag:
@bobcallaway @haydentherapper just a small bump to follow up on this after the holidays.

bobcallaway:
> @bobcallaway ok I get that. Are you guys ok with [this basic change in cosign](https://github.com/aalsabag/cosign/commit/f2e57b361b32deddfd43bef1920f2ef9e6d1a502)? What would you like me to add to it?
> 
> I can add an additional conditional for the `rekor-server` to check if it's the public one.
> 
> Obviously I'll add this change to both `attest` and `attest-blob`

i'd rather it be a separate command line argument in cosign to set the upload type (to override the default behavior)

aalsabag:
> i'd rather it be a separate command line argument in cosign to set the upload type (to override the default behavior)

[Done](https://github.com/sigstore/cosign/pull/3466). Thanks for the quick feedback.

Issue #1724, opened 2023-10-02
Allow configurable client signing algorithms


tetsuo-cpp:
**Description**

_I've filed similar issues under Cosign and Fulcio. I realise there's a lot of overlap in maintainers, but wanted to make sure that we discuss each project that we plan to touch. Apologies if this feels a bit spammy._

Hi there! At Trail of Bits, we're looking at potentially implementing part of the [Configurable Crypto Algorithms](https://docs.google.com/document/d/1-wF1t6lmJO37BzStbyz8ZkDhNm86HyvKZSaDl7uyHxI) proposal (specifically Phase 1). I wanted to float this idea to each of the relevant Sigstore sub-projects so we can hash out the details in a more concrete way.

Across the Sigstore stack, we default to using ECDSA for signatures and SHA256 for hashing. There's more detail in the linked proposal but there are a number of motivations for wanting to customise the signatures that are generated, including paving the way for post-quantum signatures. The proposed design includes having a "supported algorithm" registry (perhaps this can go in the Protobuf specs) that outlines enumerates the approved signature/hash algorithm combinations. We specifically don't want to allow arbitrary mixing and matching of signature and hash algorithm to avoid some of the security pitfalls listed in the proposal.

I'm not super familiar with the Rekor side of things but I expect that there's a few places where the assumption that the client signing algorithm is ECDSA-SHA256 is hardcoded, so we'll need to make those more flexible. We can also add a flag like `--client-signing-algorithms=alg-1,alg-2` that can be used in case a given Rekor instance wants to constrain the list of supported algorithms to be more restrictive than the aforementioned registry.

tetsuo-cpp:
Taking some notes on where we'll have to make changes:

- `pkg/types/hashedrekord/v0.0.1/entry.go:validate`: This seems to support most key types however, there's an explicit error check for Ed25519. There's some discussion on why that check exists at https://github.com/sigstore/rekor/issues/851. So this path should work for all key types supported by the Go crypto library. We'll need to a), special case Ed25519 to support it and b), add a check that rejects any keys that don't belong to the "supported algorithms" registry.
- `pkg/signer/memory.go:NewMemory`: This creates an in-memory signer. Defaults to using ECDSA. This is unrelated to supporting configurable signing algorithms on the client-side, but we could do this at some point in the future.

Hayden-IO:
[...]

Issue #1325, opened 2023-02-03
Support ed25519ph for hashedrekord for Go 1.20


Hayden-IO:
**Description**

ed25519 signatures are not supported with the hashedrekord type, though they are supported with rekord. The reason is that ed25519 computes the digest as part of its algorithm, so the original artifact is needed to verify a signature. We return an error when an ed25519 key is used with hashedrekord - https://github.com/sigstore/rekor/blob/main/pkg/types/hashedrekord/v0.0.1/entry.go#L163-L166

Go 1.20 now supports ed25519ph, so only the artifact hash is needed to verify the signature.

PR #1945, opened 2024-01-11
Add support for ed25519ph user keys in hashedrekord

ret2libc:
#### Summary

Add support for Ed25519ph user keys in hashedrekord type, so we can test #1724 better with 2 options (ecdsa+sha256 and ed25519ph).

This change requires https://github.com/sigstore/sigstore/pull/1595 and https://github.com/sigstore/rekor/pull/1959 .

#### Release Note

* Added `x509.NewSignatureWithOpts` to support ED25519ph
* Allow ED25519 public keys when validating hashedrekord entries

### DIFF
--- a/pkg/types/hashedrekord/v0.0.1/entry.go
+++ b/pkg/types/hashedrekord/v0.0.1/entry.go
@@ -19,7 +19,6 @@ import (
 	"bytes"
 	"context"
 	"crypto"
-	"crypto/ed25519"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
@@ -148,7 +147,7 @@ func (v *V001Entry) validate() (pki.Signature, pki.PublicKey, error) {
 		return nil, nil, types.ValidationError(errors.New("missing signature"))
 	}
 	// Hashed rekord type only works for x509 signature types
-	sigObj, err := x509.NewSignature(bytes.NewReader(sig.Content))
+	sigObj, err := x509.NewSignatureWithOpts(bytes.NewReader(sig.Content), options.WithED25519ph())
 	if err != nil {
 		return nil, nil, types.ValidationError(err)
 	}
@@ -160,11 +159,6 @@ func (v *V001Entry) validate() (pki.Signature, pki.PublicKey, error) {
 	keyObj, err := x509.NewPublicKey(bytes.NewReader(key.Content))
 	if err != nil {
 		return nil, nil, types.ValidationError(err)
-	}
-
-	_, isEd25519 := keyObj.CryptoPubKey().(ed25519.PublicKey)
-	if isEd25519 {
-		return nil, nil, types.ValidationError(errors.New("ed25519 unsupported for hashedrekord"))
 	}
 
 	data := v.HashedRekordObj.Data
--- a/pkg/types/hashedrekord/v0.0.1/entry_test.go
+++ b/pkg/types/hashedrekord/v0.0.1/entry_test.go
@@ -41,6 +41,7 @@ import (
 	"github.com/sigstore/rekor/pkg/types"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
 	"go.uber.org/goleak"
 )
 
@@ -113,17 +114,17 @@ func TestCrossFieldValidation(t *testing.T) {
 		Type:  "PUBLIC KEY",
 	})
 
-	// testing lack of support for ed25519
-	invalidEdPubKey, _, err := ed25519.GenerateKey(rand.Reader)
+	// testing support ed25519
+	edPubKey, edPrivKey, err := ed25519.GenerateKey(rand.Reader)
 	if err != nil {
 		t.Fatal(err)
 	}
-	invalidDer, err := x509.MarshalPKIXPublicKey(invalidEdPubKey)
+	edDer, err := x509.MarshalPKIXPublicKey(edPubKey)
 	if err != nil {
 		t.Fatal(err)
 	}
-	invalidKeyBytes := pem.EncodeToMemory(&pem.Block{
-		Bytes: invalidDer,
+	edPubKeyBytes := pem.EncodeToMemory(&pem.Block{
+		Bytes: edDer,
 		Type:  "PUBLIC KEY",
 	})
 
@@ -141,6 +142,9 @@ func TestCrossFieldValidation(t *testing.T) {
 	sha384SigBytes, _ := sha384Signer.SignMessage(bytes.NewReader(dataBytes))
 	sha512Signer, _ := signature.LoadSigner(key, crypto.SHA512)
 	sha512SigBytes, _ := sha512Signer.SignMessage(bytes.NewReader(dataBytes))
+
+	edsha512Signer, _ := signature.LoadSignerWithOpts(edPrivKey, options.WithHash(crypto.SHA512), options.WithED25519ph())
+	edsha512SigBytes, _ := edsha512Signer.SignMessage(bytes.NewReader(dataBytes))
 
 	incorrectLengthHash := sha256.Sum224(dataBytes)
 	incorrectLengthSHA := hex.EncodeToString(incorrectLengthHash[:])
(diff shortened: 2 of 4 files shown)


ret2libc:
cc @woodruffw @tetsuo-cpp

I'm not sure if we need a new version for the change in hashedrekor which allows sha512 as hash algorithm and not only sha256.

woodruffw:
Thanks @ret2libc!

Yeah, I think we may need a `0.0.2` here, which means we'll _probably_ want to hold off on making the complete `hashedrekord` changes until we have a full list of hashes we want to support. But I'm not 100% sure about that.

CCing @haydentherapper for opinions + @laurentsimon since I know you've been looking at similar changes to Rekor/`hashedrekord` 🙂

pkg/types/hashedrekord/v0.0.1/hashedrekord_v0_0_1_schema.json
```diff
@@ -38,7 +38,7 @@
                         "algorithm": {
                             "description": "The hashing function used to compute the hash value",
                             "type": "string",
-                            "enum": [ "sha256" ]
+                            "enum": [ "sha256", "sha512" ]
```
Hayden-IO: This will be the first time we're adding a non-sha256 hash, so we'll want to double check and test we don't hardcode it elsewhere.

pkg/types/hashedrekord/v0.0.1/entry.go
```diff
@@ -160,11 +159,6 @@ func (v *V001Entry) validate() (pki.Signature, pki.PublicKey, error) {
 		return nil, nil, types.ValidationError(err)
 	}
 
-	_, isEd25519 := keyObj.CryptoPubKey().(ed25519.PublicKey)
```
Hayden-IO, line 165: Just for context, when I wrote up https://github.com/sigstore/rekor/issues/1325, my intention was to support ed25519ph only for hashedrekord, and ed25519 for all other types. This might mean making some changes to the 'pkg/pki/x509/x509' library to support loading in a custom verifier like on https://github.com/sigstore/rekor/blob/70d6663e297068383f708155f4017110c6f46713/pkg/pki/x509/x509.go#L87

ret2libc: What do you think about [this](https://github.com/sigstore/rekor/pull/1945/commits/2d93686760f8fe40ea552ca9895c2922bfcb7780#diff-1cad0eed4767a1ef5dbac388d3e45fc4fe1a3c4fb6187f572040b515d81660b8) way of customizing the verifier?

pkg/types/hashedrekord/v0.0.1/entry_test.go
```diff
@@ -93,7 +93,7 @@ func TestCrossFieldValidation(t *testing.T) {
 	h := sha256.Sum256(dataBytes)
 	dataSHA := hex.EncodeToString(h[:])
 
-	signer, _ := signature.LoadSigner(key, crypto.SHA256)
+	signer, _ := signature.LoadSigner(key, crypto.SHA256, signature.LoadDefaultSV, nil)
```
ret2libc: Probably we can avoid all these changes if we add a `LoadSignerWithOpts`/`LoadVerifierWithOpts`/`LoadSignerVerifierWithOpts` instead of changing the existing ones.

pkg/types/hashedrekord/v0.0.1/entry.go
```diff
@@ -230,10 +238,11 @@ func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.A
 		return nil, errors.New("only one public key must be provided")
 	}
 
+	hashAlgorithm, hashValue := util.UnprefixSHA(props.ArtifactHash)
```
ret2libc: this is to allow passing as `props.ArtifactHash` either a regular sha value, or a prefixed sha value, which can be used to determine what is the hash algorithm used.

Is this a breaking change? cc @haydentherapper

bobcallaway:
I threw [this together](https://github.com/sigstore/rekor/pull/1958) as the smallest change to add `sha384` and `sha512` to the rekor type. I'm happy to use your PR as the definitive change, but wanted to compare/contrast what was strictly needed to extend the enum, VS what was needed for the ed25519ph key types.

i don't think we need a 0.0.2 version of hashedrekord for this, since we're not actually changing the structure or canonicalization logic of the type.

go.mod
```diff
@@ -204,3 +204,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/sigstore/sigstore => github.com/trail-of-forks/sigstore v0.0.0-20240117103256-3095d93bafe8
```
bobcallaway: this would need to be removed before merging.

woodruffw: Yep! I believe this is here until https://github.com/sigstore/sigstore/pull/1595 gets merged (correct me if I'm wrong @ret2libc), so it probably makes sense to merge #1958 first and then we can rebase here 🙂

Hayden-IO (commented):
LGTM, this looks good.

While this might be obvious from the change since it only modifies hashedrekord, i wanted to double check that this won't affect other types. rekord also uses x509.NewSignature indirectly through a factory, but given that we've added a new signature function rather than modify the existing one, this has no impact on other types.

Let's merge https://github.com/sigstore/rekor/pull/1958 first (or split this into two) if that works for you, just to make it easier to see the two different changes.

ret2libc:
> I threw [this together](https://github.com/sigstore/rekor/pull/1958) as the smallest change to add `sha384` and `sha512` to the rekor type. I'm happy to use your PR as the definitive change, but wanted to compare/contrast what was strictly needed to extend the enum, VS what was needed for the ed25519ph key types.
> 
> i don't think we need a 0.0.2 version of hashedrekord for this, since we're not actually changing the structure or canonicalization logic of the type.

I have included the changes in your PR into mine, and mentioned you in the commit https://github.com/sigstore/rekor/pull/1945/commits/2577342834de752c2fd0e8b056fc2704186739fe . If that's not the proper way to "borrow" code let me know how I should note your help!

By the way, @haydentherapper I'm going to split this into SHA* additions and ED25519ph support.

Hayden-IO:
LGTM, thanks for splitting. Will approve once the sigstore/sigstore PR is merged.

Hayden-IO (commented):
LGTM, thanks for splitting this PR up. Will approve once we merge s/s and remove the replace.

go.mod
```diff
@@ -204,3 +204,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+// TODO: REMOVE ME
```
Hayden-IO: good to remove now

Hayden-IO (commented):
LGTM, thanks for your patience and for all the discussion on the underlying verifier interface. This came out really clean!

Can you update the test for this (https://github.com/sigstore/rekor/blob/main/pkg/types/hashedrekord/v0.0.1/entry_test.go#L116-L120 and https://github.com/sigstore/rekor/blob/main/pkg/types/hashedrekord/v0.0.1/entry_test.go#L195-L211), which I would have expected to be failing now (so it might not have been testing the right thing)?

Hayden-IO (dismissed):

Hayden-IO:
@ret2libc can you rebase?

ret2libc:
> @ret2libc can you rebase?

@haydentherapper rebased! Sorry for the delay

Hayden-IO:
No worries! Sorry, need one more rebase

go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/rs/cors v1.10.1
 	github.com/sassoftware/relic v7.2.1+incompatible
 	github.com/secure-systems-lab/go-securesystemslib v0.8.0
-	github.com/sigstore/sigstore v1.8.1
+	github.com/sigstore/sigstore v1.8.2-0.20240215213640-922069a3bc39
```
Hayden-IO: can you also bump to v1.8.2, we just cut a release

woodruffw: Sure, doing now!

Hayden-IO (dismissed):

bobcallaway (dismissed):

bobcallaway:
other than fixing the merge conflict, I think this is ready to go

Hayden-IO (approved):

PR #1957, opened 2024-01-17
Change Redis value for locking mechanism

Hayden-IO:
The lock value is never read. Only the key is used to determine if an instance has already created a checkpoint for a given period. Changing this to the smallest primitive value to reduce storage capacity needs.

### DIFF
--- a/pkg/witness/publish_checkpoint.go
+++ b/pkg/witness/publish_checkpoint.go
@@ -151,7 +151,9 @@ func (c *CheckpointPublisher) publish(tc *trillianclient.TrillianClient, sTreeID
 
 	// return value ignored, which is whether or not the entry was set
 	// no error is thrown if the key already exists
-	successNX, err := c.redisClient.SetNX(ctx, key, hexCP, 0).Result()
+	// use smallest value as it's unused
+	value := true
+	successNX, err := c.redisClient.SetNX(ctx, key, value, 0).Result()
 	if err != nil {
 		c.reqCounter.With(
 			map[string]string{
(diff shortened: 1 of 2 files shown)


pkg/witness/publish_checkpoint.go
```diff
@@ -151,7 +151,9 @@ func (c *CheckpointPublisher) publish(tc *trillianclient.TrillianClient, sTreeID
 
 	// return value ignored, which is whether or not the entry was set
 	// no error is thrown if the key already exists
-	successNX, err := c.redisClient.SetNX(ctx, key, hexCP, 0).Result()
+	// use smallest value as it's unused
+	value := true
+	successNX, err := c.redisClient.SetNX(ctx, key, value, 0).Result()
```
cmurphy, line 156: Does line 185 need to be changed too? Or why does `/latest` need to have the actual checkpoint value when the timestamped key doesn't?

Hayden-IO: `/latest` stores the most recent stable checkpoint and is read when requesting a stable checkpoint through `api/v1/log?stable=true` - https://github.com/sigstore/rekor/blob/ca115c94322d9b7759138ef12438118465408896/pkg/api/tlog.go#L57-L59

This key is only used to prevent multiple instances from writing different checkpoints, in the case of clock skew. If one instance were behind, this prevents it from overwriting the checkpoint. Its value is never used.

cmurphy (approved):

bobcallaway (approved):

PR #1958, opened 2024-01-17
add support for sha384 and sha512 to hashedrekord type

bobcallaway:

### DIFF
--- a/pkg/types/hashedrekord/v0.0.1/entry.go
+++ b/pkg/types/hashedrekord/v0.0.1/entry.go
@@ -18,6 +18,7 @@ package hashedrekord
 import (
 	"bytes"
 	"context"
+	"crypto"
 	"crypto/ed25519"
 	"crypto/sha256"
 	"encoding/hex"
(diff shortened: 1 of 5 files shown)


pkg/types/hashedrekord/v0.0.1/entry.go
```diff
@@ -178,11 +179,21 @@ func (v *V001Entry) validate() (pki.Signature, pki.PublicKey, error) {
...
+	default:
+		alg = crypto.SHA256
+	}
+
 	decoded, err := hex.DecodeString(*hash.Value)
 	if err != nil {
 		return nil, nil, err
 	}
-	if err := sigObj.Verify(nil, keyObj, options.WithDigest(decoded)); err != nil {
+	if err := sigObj.Verify(nil, keyObj, options.WithDigest(decoded), options.WithCryptoSignerOpts(alg)); err != nil {
```
Hayden-IO, line 196: nit for another file: Should we add a comment that the digest in https://github.com/sigstore/rekor/blob/fc28ac1bc2ecae172371df6b63efe48daee863ff/pkg/pki/x509/x509.go#L87 is a default? I was looking through the code and was initially uncertain if that needed to be modified too.

Hayden-IO (approved):

pkg/types/hashedrekord/v0.0.1/entry.go
bobcallaway: it is a default, so we should probably add that comment. the `WithCryptoSignerOpts()` already provides a way to override the algorithm.

ret2libc: I think I was missing this in my PR #1945 .

bobcallaway:
Closed in deference to #1959

PR #499, opened 2021-11-16
api/index/retrieve: allow searching on indices with sha1 hashes

mikhailswift:
Closes #498

#### Summary

Adds the ability to search for indices with sha1 hashes. Currently
rekor custom types can store indices with formats other than
sha256:<hash>.  Particularly the in-toto type can do this.

One particular use case of interest is indexing log entries by git
commit hash, which largely still use sha1.
#### Ticket Link

Fixes

#### Release Note

```release-note
* Added ability to search indices with sha1 hashes
```

PR #1959, opened 2024-01-18
Added support for sha384/sha512 hash algorithms in hashedrekords

ret2libc:
#### Summary

Add support for SHA384/512 in hashedrekord type. This PR has been split from https://github.com/sigstore/rekor/pull/1945 and borrows code from https://github.com/sigstore/rekor/pull/1958 (from @bobcallaway ).

#### Release Note

* Extend hashedrekord v0.0.1 schema to support sha256, sha384, and sha512 in `data.hash.algorithm`
* Changed `hashedrekord.CreateFromArtifactProperties` to accept a prefixed-checksum (e.g. `sha512:<value>`) to determine the type of hash used.
* Added `util.UnprefixSHA` API to split a string into the crypto.Hash being used and its value

### DIFF
--- a/pkg/types/hashedrekord/v0.0.1/e2e_test.go
+++ b/pkg/types/hashedrekord/v0.0.1/e2e_test.go
@@ -25,15 +25,11 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"os"
-	"strings"
 	"testing"
 	"time"
 
-	"github.com/go-openapi/strfmt"
-	"github.com/go-openapi/swag"
 	"github.com/sigstore/rekor/pkg/client"
 	"github.com/sigstore/rekor/pkg/generated/client/entries"
-	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/types"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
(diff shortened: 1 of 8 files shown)


ret2libc:
cc @haydentherapper @bobcallaway @woodruffw @tetsuo-cpp

pkg/types/hashedrekord/v0.0.1/entry.go
```diff
@@ -230,10 +253,11 @@ func (v V001Entry) CreateFromArtifactProperties(_ context.Context, props types.A
 		return nil, errors.New("only one public key must be provided")
 	}
 
+	hashAlgorithm, hashValue := util.UnprefixSHA(props.ArtifactHash)
```
ret2libc, line 256: this is to allow passing as props.ArtifactHash either a regular sha value, or a prefixed sha value, which can be used to determine what is the hash algorithm used.

Is this a breaking change? cc @haydentherapper @bobcallaway

pkg/util/sha.go
```diff
@@ -39,3 +40,34 @@ func PrefixSHA(sha string) string {
...
+
+	if len(components) == 2 {
+		prefix := components[0]
+		sha = components[1]
+
+		switch prefix {
+		case "sha1":
+			return crypto.SHA1, sha
+		case "sha256":
+			return crypto.SHA256, sha
```
bobcallaway, line 57: could you add sha384 here since we're adding it into the JSON schema, and in the length check below too please?

pkg/types/hashedrekord/v0.0.1/entry.go
```diff
@@ -178,17 +180,38 @@ func (v *V001Entry) validate() (pki.Signature, pki.PublicKey, error) {
...
 	}
 
 	return sigObj, keyObj, nil
 }
 
+func getDataHashAlgorithm(hashAlgorithm crypto.Hash) string {
+	switch hashAlgorithm {
+	case crypto.SHA256:
+		return models.HashedrekordV001SchemaDataHashAlgorithmSha256
+	case crypto.SHA512:
```
bobcallaway, line 208: can you add sha384 in this list?

pkg/util/sha.go
ret2libc: Done!

ret2libc: Added in all the 3 switches (in `PrefixSHA` as well)

pkg/types/hashedrekord/v0.0.1/entry.go
ret2libc: Done!

pkg/types/hashedrekord/v0.0.1/entry.go
bobcallaway: I don't think this is a breaking change, since all existing callers essentially presumed it needed to be a sha256 value

pkg/util/sha.go
```diff
@@ -33,9 +34,46 @@ func PrefixSHA(sha string) string {
...
+		case "sha512":
+			return crypto.SHA512, sha
+		default:
+			return crypto.Hash(0), ""
+		}
+	}
+
+	switch len(sha) {
+	case 40:
+		return crypto.SHA1, sha
```
bobcallaway, line 69: I know this is probably because the function above included `sha1`, but I think we should remove this (and the stanza above). There's no other use of that AFAIK in sigstore, and I wouldn't want to perpetuate this.

bobcallaway: actually, i think it is used to search for entries from gitsign, im working to confirm 

@wlynch - any thoughts?

bobcallaway: https://github.com/sigstore/rekor/pull/499

bobcallaway: nm, i think we need to leave this - but it would be nice to add a test case that calls `CreateFromArtifactProperties` with a proposed sha1 value to ensure that it doesn't somehow end up in the log.

Hayden-IO (commented):
LGTM, can you add the one test bob mentioned?

pkg/util/sha.go
ret2libc: Added it in https://github.com/sigstore/rekor/pull/1959/commits/6903f5f774fb0fed31211bee18193d17cf6c11ab . Please let me know if this is what you meant, otherwise I can adjust it.

pkg/types/hashedrekord/v0.0.1/e2e_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	re.HashedRekordObj.Signature.PublicKey = &models.HashedrekordV001SchemaSignaturePublicKey{}
+	publicKeyBytes := [][]byte{pubBytes}
+
+	re.HashedRekordObj.Signature.PublicKey.Content = strfmt.Base64(publicKeyBytes[0])
+	re.HashedRekordObj.Data.Hash = &models.HashedrekordV001SchemaDataHash{
+		Algorithm: swag.String("sha1"),
+		Value:     swag.String("a17c9aaa61e80a1bf71d0d850af4e5baa9800bbd"),
+	}
+
```
bobcallaway: Could you switch this to use the same pattern as above where you call `CreateFromArtifactProperties`? That is the call path that is at risk (as the JSON Schema validation will fail the API request)

woodruffw: Doing this now!

woodruffw: Done in 0f1f8e3: I've moved this out of e2e since it no longer needs access to a runtime Rekor instance.

ret2libc: The reason why I did not use `CreateFromArtifactProperties` is that the test would fail in `CreateFromArtifactProperties` on the client side and and not on the rekor server side. Thus I wanted to test that the server did not accept it rather than just checking the client.

woodruffw: > The reason why I did not use `CreateFromArtifactProperties` is that the test would fail in `CreateFromArtifactProperties` on the client side and and not on the rekor server side. Thus I wanted to test that the server did not accept it rather than just checking the client.

That makes sense, but I think @bobcallaway is pointing out that the server side is already checked by virtue of the JSON Schema Validation -- we could include both checks, though (with the non-e2e doing `CreateFromArtifactProperties` and the e2e doing it manually).

pkg/util/sha.go
wlynch: We don't use this for gitsign today - gitsign does a sha256 over the commit content + sha1 for rekor.

(though it would be a neat way to remove some of the chattiness gitsign has today with it's sha -> rekor entry lookup 🤔 ) 

Since it looks like this is already done + tests were added, I say keep it. We can revisit removing it later.

bobcallaway (approved):

PR #1974, opened 2024-01-23
Added --client-signing-algorithms flag

ret2libc:
#### Summary

Add `--client-signing-algorithms` flag to rekor-server to restrict the set of client keys accepted by a Rekor instance. See #1724 .

This work depends on https://github.com/sigstore/sigstore/pull/1601


ret2libc:
```
--- FAIL: TestJAR (0.07s)
    e2e_test.go:33: Using config file:%!(EXTRA string=/tmp/rekor_test.9ofp49.rekor.yaml)
        [POST /api/v1/log/entries][400] createLogEntryBadRequest  &{Code:400 Message:error processing entry: getting verifiers: jar v0.0.1 entry not initialized}
        warning: GOCOVERDIR not set, no coverage data emitted
```

Apparently the e2e test does not fill the Signature part of the JARModel:
```golang
	if v.JARModel.Signature == nil || v.JARModel.Signature.PublicKey == nil || v.JARModel.Signature.PublicKey.Content == nil {
		return nil, errors.New("jar v0.0.1 entry not initialized")
	}
```

Shall this be handled or is the signature always expected to be present?

go.mod
```diff
@@ -63,7 +62,8 @@ require (
 	github.com/jmoiron/sqlx v1.3.5
 	github.com/redis/go-redis/v9 v9.4.0
 	github.com/sassoftware/relic/v7 v7.6.1
-	github.com/sigstore/protobuf-specs v0.2.1
+	github.com/sigstore/protobuf-specs v0.3.0-beta.2
+	github.com/sigstore/sigstore v1.6.4
```
bobcallaway: that's a pretty significant downgrade in the version of sigstore/sigstore - was that intended?

Hayden-IO:
Signature should be present, it's required to validate a new jar entry - https://github.com/sigstore/rekor/blob/d596e9dead8cc355a2f4b2d2900550ef2cdf08d8/pkg/types/jar/v0.0.1/entry.go#L248

cmd/rekor-server/app/root.go
```diff
@@ -133,10 +137,21 @@ Memory and file-based signers should only be used for testing.`)
...
+	for _, keyAlgorithm := range api.AllowedClientSigningAlgorithms {
+		keyFlag, err := signature.FormatSignatureAlgorithmFlag(keyAlgorithm)
+		if err != nil {
+			panic(err)
+		}
+		keyAlgorithmTypes = append(keyAlgorithmTypes, keyFlag)
+	}
+	sort.Strings(keyAlgorithmTypes)
+	keyAlgorithmHelp := fmt.Sprintf("signing algorithm to use for signing/hashing (allowed %s)", strings.Join(keyAlgorithmTypes, ", "))
+	rootCmd.PersistentFlags().StringSlice("client-signing-algorithms", keyAlgorithmTypes, keyAlgorithmHelp)
```
bobcallaway, line 158: just checking that your intent is to default to allowing all types.

go.mod
```diff
@@ -204,3 +204,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/sigstore/sigstore => github.com/trail-of-forks/sigstore v0.0.0-20240219090738-536a0415e570
```
bobcallaway: still needs removed before merge

cmd/rekor-server/app/root.go
ret2libc: Yes, I think by default we want to allow everything that was already allowed. (there were no real restrictions before).

pkg/api/entries.go
```diff
@@ -192,12 +198,82 @@ func GetLogEntryByIndexHandler(params entries.GetLogEntryByIndexParams) middlewa
...
 
+func checkEntryAlgorithms(entry types.EntryImpl) (bool, error) {
+	verifiers, err := entry.Verifiers()
+	if err != nil {
+		return false, fmt.Errorf("getting verifiers: %w", err)
+	}
+	// Default to SHA256 if no artifact hash is specified
+	artifactHashValue := crypto.SHA256
+	// Get artifact hash from entry
+	artifactHash, err := entry.ArtifactHash()
```
bobcallaway: what if `err != nil`?

pkg/api/entries.go
```diff
@@ -192,12 +198,82 @@ func GetLogEntryByIndexHandler(params entries.GetLogEntryByIndexParams) middlewa
...
 	}
+
+	areEntryAlgorithmsAllowed, err := checkEntryAlgorithms(entry)
+	if err != nil {
+		return nil, handleRekorAPIError(params, http.StatusBadRequest, err, fmt.Sprintf(validationError, err))
+	}
+	if !areEntryAlgorithmsAllowed {
+		return nil, handleRekorAPIError(params, http.StatusBadRequest, errors.New("entry algorithms are not allowed"), fmt.Sprintf(validationError, "entry algorithms are not allowed"))
+	}
+
```
bobcallaway, line 297: this feels like it should be inside `types.CreateVersionedEntry` instead of in the API layer; wdyt?

pkg/types/jar/v0.0.1/entry.go
```diff
@@ -323,12 +323,30 @@ func (v *V001Entry) CreateFromArtifactProperties(ctx context.Context, props type
 	}
 	re.JARModel.Archive.Content = (strfmt.Base64)(artifactBytes)
 
-	if err := re.validate(); err != nil {
+	keyObj, sigObj, err := re.fetchExternalEntities(ctx)
```
bobcallaway: I'm not following what these changes have to do with this PR?

pkg/api/entries.go
ret2libc: Maybe, but it uses `api.algorithmRegistry` and nothing in pkg/types use `api`. Would that be fine anyway?

pkg/api/entries.go
```diff
@@ -192,12 +200,101 @@ func GetLogEntryByIndexHandler(params entries.GetLogEntryByIndexParams) middlewa
...
+func checkEntryAlgorithms(entry types.EntryImpl) (bool, error) {
+	// Only check algorithms for hashedrekord entries
+	switch entry.(type) {
+	case *hashedrekord.V001Entry:
+		break
+	default:
+		return true, nil
+	}
+
```
ret2libc, line 250: @bobcallaway @haydentherapper shall we start with hashedrekord entries only? If we want to add support for other types I guess we need to change the schemas similarly to what we did for hashedrekord.

pkg/types/jar/v0.0.1/entry.go
ret2libc: removed, probably some older stuff I think

pkg/api/entries.go
ret2libc: refactored this a bit, let me know if it's better!

pkg/api/entries.go
woodruffw: My 0.02c is that it's fine to start here: my understanding is that the long term plan is to remove a lot of the non-hashed entries with Rekor v2 anyways, plus this flag is primarily for test deployment purposes anyways 🙂.

bobcallaway:
code looks fine. I would like to see some e2e tests that show it works before merging.

ret2libc:
@bobcallaway I added an e2e test, though I don't see it running in the CI. Could it be because it's not in `main` yet?

bobcallaway:
> @bobcallaway I added an e2e test, though I don't see it running in the CI. Could it be because it's not in `main` yet?

It requires `build` so it won't show up in the list until that is done since it's a new job.

https://github.com/sigstore/rekor/actions/runs/13202747393/job/36858854058?pr=1974

bobcallaway (dismissed):

ret2libc:
> > @bobcallaway I added an e2e test, though I don't see it running in the CI. Could it be because it's not in `main` yet?
> 
> It requires `build` so it won't show up in the list until that is done since it's a new job.
> 
> https://github.com/sigstore/rekor/actions/runs/13202747393/job/36858854058?pr=1974

Great! Seems to be green :) Let me know what else I can do to make the PR better/ready

bobcallaway (approved):

pkg/api/entries.go
Hayden-IO: late reply, but yes, I agree, starting with just hashedrekord is totally reasonable

PR #1998, opened 2024-02-06
Add TLS support for Redis Client implementation

### LINKED ISSUE #1999: Missing TLS support for Redis Client
**Description**

Right now, Rekors Redis client implementation does not allow enforcing TLS. This is an issue, because a lot of Redis cluster implementations require TLS for the client connection setup and don't allow non-TLS clients. A good example of this is AWS ElastiCache, when MultiAZ is enabled on ElastiCache Redis, it will require TLS based client connections and will close any non-TLS connections.

mihkelparna1:
#### Summary
Adds support for using TLS when establishing a Redis Client. This becomes a mandatory step when using systems like Elasticache that enforce client TLS when set up to run in a Multi-AZ way. 

#### Release Note

- Adds TLS configuration by default to the Redis client implementation, this can be disabled by using ```insecureSkipVerify``` flag, closes https://github.com/sigstore/rekor/issues/1999 issue


pkg/api/api.go
```diff
@@ -179,7 +180,10 @@ func ConfigureAPI(treeID uint) {
 			Addr:     fmt.Sprintf("%v:%v", viper.GetString("redis_server.address"), viper.GetUint64("redis_server.port")),
 			Password: viper.GetString("redis_server.password"),
 			Network:  "tcp",
-			DB:       0, // default DB
+			TLSConfig: &tls.Config{
+				InsecureSkipVerify: viper.GetBool("redis_server.insecure-skip-verify"),
```
bobcallaway: I believe you'll need to define this flag in `cmd/rekor-server/app/root.go` in order for this to work

mihkelparna1: Like this - https://github.com/sigstore/rekor/pull/1998/commits/fe9b0a507a2ef61451d4b2b0f23c61dc509fab20 ?

cmd/rekor-server/app/root.go
```diff
@@ -115,6 +115,7 @@ Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("redis_server.address", "127.0.0.1", "Redis server address")
 	rootCmd.PersistentFlags().Uint16("redis_server.port", 6379, "Redis server port")
 	rootCmd.PersistentFlags().String("redis_server.password", "", "Redis server password")
+	rootCmd.PersistentFlags().Bool("redis_server.insecure-skip-verify", false, "Whether to skip TLS verification or not")
```
bobcallaway: nit: I'd add something in the description to note that this is only for connections to Redis

bobcallaway: also, setting this default to `false` would represent a breaking change for clients (which all would be connecting without TLS today) - is that correct?

mihkelparna1: Yes, it would be a breaking change, we can change it to 'true', but that feels like going against the standard approach when adding flags of this nature where you would expect an explicit opt-out from a more 'secure' way, open to feedback on this though, might just be my understanding of it. I pushed a commit to change the description to mention that this applies to the redis endpoint, not sure if it's clear enough.

bobcallaway: I think we'd just need to be sure the next release bumped the minor version of Rekor with a clear note that if folks are using Redis that they'd need to specify the flag - tagging @haydentherapper for his thoughts

Hayden-IO: I wouldn't include insecure in the flag name and instead call this "enable-tls" or something similar, which would avoid a breaking change. I don't think "insecure" needs to be mentioned because skipping TLS verification is not necessarily insecure depending on the setup - a redis instance could be behind a VPC in a production setup for example.

mihkelparna1: If we rename skipVerify to enable-tls then disabling the breaking change would mean that we would need to set the default values to 'true' for it. That seems a bit odd. InsecureSkipVerify is a pretty standard way of defining this feature and widely adopted. We can maybe just change the default behaviour to 'true' as a middle-ground?

Hayden-IO: Sorry, I'm not following how adding an enable-tls flag would be a breaking change. Given this feature is adding support for TLS, we could add a flag for enabling TLS set to false by default, and for those who need TLS, they can set the new flag to true. This is just the inverse of the current proposal to have an insecure flag set to true by default, it seems like better optics to have an enable flag set to false rather than an insecure flag set to true.

mihkelparna1: Sorry, misunderstood, was under the impression we were still only talking about the insecureFlag and not TLS in general. 

Pushed a commits ( https://github.com/sigstore/rekor/pull/1998/commits/53238d4b0c708ddf987a538d85d2310614672651 & https://github.com/sigstore/rekor/pull/1998/commits/a0c712d4187954963055f5c3bb691b5e24326159 ) that introduces the 'enable-tls' flag under the 'redis_server' group ( in addition to InsecureSkipVerify - Added explanation to the description that mentions it's only applicable when redis enable-tls is set ).

Had to extract the client creation into a separate function to wrap it with an if statement based on the 'enable-tls' flag value, not sure if that is the best way to do this.

cmd/backfill-redis/main.go
```diff
@@ -210,6 +208,30 @@ func main() {
...
+			DB:        0, // default DB
+		})
+	}
+	return redis.NewClient(&redis.Options{
+		Addr:     fmt.Sprintf("%s:%s", *redisHostname, *redisPort),
+		Password: *redisPassword,
+		Network:  "tcp",
+		DB:       0, // default DB
+	})
+}
```
bobcallaway: minor nit here and in api.go as well: could you reduce the amount of duplicated code here?

for example:
```
opts := &redis.Options{
	Addr:      fmt.Sprintf("%s:%s", *redisHostname, *redisPort),
	Password:  *redisPassword,
	Network:   "tcp",
	DB:        0, // default DB
}

if *enableTLS {
	opts.TLSConfig = &tls.Config{
		InsecureSkipVerify: *insecureSkipVerify,
	}
}

return redis.NewClient(opts)
```

bobcallaway:
I think this should be ready to merge once the last comment is addressed - thanks for all the help on this :)

cmd/backfill-redis/main.go
mihkelparna1: Pushed a commit ( https://github.com/sigstore/rekor/pull/1998/commits/191f24a961b8a2516286bca0241ea78df3df8fcf ) to reduce duplicate code.

bobcallaway (dismissed):

Hayden-IO (changes requested):
This also needs to be set in https://github.com/sigstore/rekor/blob/d596e9dead8cc355a2f4b2d2900550ef2cdf08d8/pkg/indexstorage/redis/redis.go#L45, which is for the search index. The change you made here is just for the checkpoints we write to redis.

mihkelparna1:
> This also needs to be set in
> 
> https://github.com/sigstore/rekor/blob/d596e9dead8cc355a2f4b2d2900550ef2cdf08d8/pkg/indexstorage/redis/redis.go#L45
> 
> , which is for the search index. The change you made here is just for the checkpoints we write to redis.

Something along the lines of this -> https://github.com/sigstore/rekor/pull/1998/commits/3b7485d056d94714704829590f586ba2a40630c9 ?

Hayden-IO (approved):
Thanks!

PR #2023, opened 2024-02-27
Update storage.sql script

cmurphy:
Update the storage.sql setup script to match the script in trillian 1.6.0[1]. This prevents the following error when running the script against the most revent MySQL version in GCP:

   You have an error in your SQL syntax; check the manual that
   corresponds to your MySQL server version for the right syntax to use
   near 'IF NOT EXISTS TreeHeadRevisionIdx ON TreeHead(TreeId,
   TreeRevision)' at line 1

[1] https://github.com/google/trillian/blob/v1.6.0/storage/mysql/schema/storage.sql


Hayden-IO (approved):
From looking at MySQL docs, this fails because create index doesn’t support an existence check, correct?

cmurphy:
@haydentherapper correct

Issue #1771, opened 2023-10-16
Failed to install rekor-cli latest version (1.3.2), broken dependencies


csarta:
**Description**
Hello there, I've tried to install rekor v1.3.2 following the documentation, running:
`go install -v github.com/sigstore/rekor/cmd/rekor-cli@latest`

But I get the following errors:
```
go/pkg/mod/github.com/letsencrypt/boulder@v0.0.0-20221109233200-85aa52084eaf/core/objects.go:15:2: unrecognized import path "gopkg.in/square/go-jose.v2": reading https://gopkg.in/square/go-jose.v2?go-get=1: 502 Bad Gateway
	server response: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/square/go-jose.git/info/refs?service=git-upload-pack: write tcp 10.131.9.20:52252->140.82.121.4:443: write: broken pipe
go/pkg/mod/github.com/sigstore/rekor@v1.3.2/pkg/types/cose/v0.0.1/entry.go:37:2: reading github.com/veraison/go-cose/go.mod at revision v1.2.0: unknown revision v1.2.0
```
I've noticed that project go-jose has moved from `square/go-jose` to `go-jose/go-jose`.
I've also noticed that project go-cose latest version is v1.1.0
Trying to install rekor v1.3.1 works correctly.

**Version**
1.3.2

Hayden-IO:
Hey, I'm not able to replicate this, so I believe this was a one-off failure. Please try to install again.

Thanks for pointing out the deprecated package.

Hayden-IO:
It looks like the square package is pulled in through a dependency, so we aren't directly using it.

lrascao:
also seeing this issue, `rekor` is directly importing `go-cose` `1.2.0` which has been [retracted](https://pkg.go.dev/github.com/veraison/go-cose@v1.2.0)
https://github.com/veraison/go-cose/issues/152

Hayden-IO:
Thanks @lrascao, created https://github.com/sigstore/rekor/pull/2030

PR #2030, opened 2024-03-06
Bump go-cose to non-retracted version

Hayden-IO:
Ref: https://github.com/sigstore/rekor/issues/1771


Hayden-IO:
There should be no functional difference, beyond fixing local installs.

cpanato (approved):

PR #2042, opened 2024-03-13
Add index performance script

cmurphy:
Add terraform configuration and scripts to set up rekor standalone on GCP, perform a series of insert and search operations, use Prometheus to gather metrics, and plot the results with gnuplot.

The scripts added here are for comparing mysql and redis as index storage backends. Other types of performance measurement scripts could be added here in the future.

To get a realistic sense of query speed for searches, a large data set is needed. Rather than using the rekor API to insert real data, fake data is generated and uploaded directly to the backend before searching it.

Different types of searches are performed: searches where there should be many results, searches where there should be few results, and searches where there should be no results. The goal is not to compare the latency of these different searches, but to take the overall average to compare across backends.

Depends on https://github.com/sigstore/scaffolding/pull/1036

### DIFF
--- a/scripts/performance/index-storage/index-performance.sh
+++ b/scripts/performance/index-storage/index-performance.sh
@@ -1,0 +1,385 @@
+#!/bin/bash
+#
+# Copyright 2024 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+
+cleanup() {
+    code=$?
+    if [ $code -ne 0 ] ; then
+        echo "An error occurred. Waiting 30 seconds to start cleanup, press ^C to cancel cleanup."
+        sleep 30
+    fi
+    $@
+    exit $code
+}
+
+INSERT_RUNS=${INSERT_RUNS:-1000}
+SEARCH_ENTRIES=${SEARCH_ENTRIES:-100000}
+INDEX_BACKEND=${INDEX_BACKEND:-redis}
+REGION=${REGION:-us-west1}
+
+setup_bastion() {
+    echo "Configuring the bastion..."
+    sudo apt install kubernetes-client google-cloud-sdk-gke-gcloud-auth-plugin git redis-tools gnuplot prometheus minisign jq -y
+    if ! which hyperfine >/dev/null ; then
+        local tag=$(curl -H "Accept: application/json" -L https://github.com/sharkdp/hyperfine/releases/latest | jq -r .tag_name)
+        wget -O /tmp/hyperfine_${tag:1}_amd64.deb https://github.com/sharkdp/hyperfine/releases/download/${tag}/hyperfine_${tag:1}_amd64.deb
+        sudo dpkg -i /tmp/hyperfine_${tag:1}_amd64.deb
+    fi
+    if ! which helm >/dev/null ; then
+        curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+    fi
+    if ! which rekor-cli >/dev/null ; then
+        wget -O /tmp/rekor-cli-linux-amd64 https://github.com/sigstore/rekor/releases/latest/download/rekor-cli-linux-amd64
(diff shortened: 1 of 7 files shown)


cmurphy:
Example output: https://gist.github.com/cmurphy/71683066aa94084e6795cd406fd1eab4

I've not added any workflows to run this script regularly, but it could be useful in the future to run this automatically to track performance trends. It would not be appropriate to gate on this, since an individual performance run could be better or worse from one to another based on a wide variety of factors.

I'm using this to do work on the public instance and it was suggested that it could be useful to include in this repo. I'm also fine with keeping it separate if this seems too niche.

scripts/performance/index-storage/index-performance.sh
```diff
@@ -0,0 +1,374 @@
...
+
+INSERT_RUNS=${INSERT_RUNS:-1000}
+SEARCH_ENTRIES=${SEARCH_ENTRIES:-100000}
+INDEX_BACKEND=${INDEX_BACKEND:-redis}
+REGION=${REGION:-us-west1}
+
+setup_bastion() {
+    echo "Configuring the bastion..."
+    sudo apt install kubernetes-client google-cloud-sdk-gke-gcloud-auth-plugin git redis-tools gnuplot prometheus minisign -y
+    which hyperfine >/dev/null || ( wget -O /tmp/hyperfine_1.16.1_amd64.deb https://github.com/sharkdp/hyperfine/releases/download/v1.16.1/hyperfine_1.16.1_amd64.deb && sudo dpkg -i /tmp/hyperfine_1.16.1_amd64.deb )
```
Hayden-IO: Can we download `latest` for each of these? There's some risk of breaking changes to the script, but that should be detectable at runtime.

scripts/performance/index-storage/terraform/main.tf
```diff
@@ -0,0 +1,134 @@
...
+}
+
+module "rekor" {
+    source = "git::https://github.com/sigstore/scaffolding.git//terraform/gcp/modules/rekor?ref=36131f6dd57a0b1d9eaad0f7dd17b11c3ab0eeee"
+
+    region = var.region
+    project_id = var.project
+    cluster_name = "rekor"
+
+    attestation_bucket = "cmurphy-sigstore-attestations"
```
Hayden-IO: Can you remove attestation bucket, since that isn't required until intoto types are in use?

scripts/performance/index-storage/terraform/main.tf
```diff
@@ -0,0 +1,134 @@
...
+    oslogin = {
+        enabled = true
+        enabled_with_2fa = false
+    }
+    instance_os_login_members = {
+        bastion = {
+            instance_name = module.bastion.name
+            zone = module.bastion.zone
+            members = [
+                "serviceAccount:ga-206@colleenmurphy-testing-410318.iam.gserviceaccount.com",
```
Hayden-IO: Should make these variables, or at least mention in docs they'd need to be updated.

scripts/performance/index-storage/terraform/main.tf
```diff
@@ -0,0 +1,134 @@
...
+
+module "network" {
+    source = "git::https://github.com/sigstore/scaffolding.git//terraform/gcp/modules/network?ref=36131f6dd57a0b1d9eaad0f7dd17b11c3ab0eeee"
+
+    region = var.region
+    project_id = var.project
+    cluster_name = "rekor"
+}
+
+module "bastion" {
```
Hayden-IO, line 29: Just wondering, is a bastion needed for the tests?

Hayden-IO (commented):
This looks great! I think it's fine to include the example in this repo of how to run this on GCP. Only question is if we should try to keep the various dependencies up to date or not.

scripts/performance/index-storage/terraform/main.tf
cmurphy: Oops, did not mean to leave this in here

scripts/performance/index-storage/terraform/main.tf
cmurphy: The bastion is needed to install the kubernetes services from. It's also easier to run the tests from here instead of setting up a gateway and public URL (and in my case org policies prevented me from doing so).

Hayden-IO: Sounds good, just saw you noted that in the readme.

Hayden-IO (dismissed):

scripts/performance/index-storage/index-performance.sh
```diff
@@ -0,0 +1,374 @@
+#!/bin/bash -e
```
cpanato: can we use
```
#!/usr/bin/env bash

set -o errexit
```

cpanato: to make more explicit

cmurphy: done

scripts/performance/index-storage/index-performance.sh
cmurphy: done

Hayden-IO (approved):

PR #2079, opened 2024-04-12
Run e2e tests on mysql and redis index backends

cmurphy:
Ensure both index storage backend paths are functionally tested.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -90,10 +90,22 @@ jobs:
       - name: install gocovmerge
         run: make gocovmerge
 
-      - name: CLI
+      - name: CLI mysql index
         run: ./tests/e2e-test.sh
-      - name: Refactor-e2e # this will a WIP to move all the tests to respective packages
+        env:
+          INDEX_BACKEND: mysql
+      - name: CLI redis index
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -90,10 +90,22 @@ jobs:
...
+      - name: CLI mysql index
         run: ./tests/e2e-test.sh
-      - name: Refactor-e2e # this will a WIP to move all the tests to respective packages
+        env:
+          INDEX_BACKEND: mysql
+      - name: CLI redis index
+        run: ./tests/e2e-test.sh
+        env:
+          INDEX_BACKEND: redis
+      - name: Refactor-e2e mysql index # this will a WIP to move all the tests to respective packages
```
cmurphy, line 101: It looks like this refactor was started two years ago and never finished, I'm not sure it's worth duplicating these steps or which step should be preferred.

cpanato (approved):
thanks

bobcallaway (approved):

.github/workflows/main.yml
bobcallaway: this seems reasonable to me at first glance

PR #2081, opened 2024-04-12
Add MySQL support to backfill script

cmurphy:
Rename backfill-redis to backfill-index and add support for when MySQL
is used as the index storage backend.

Some Redis-specific parameters are renamed to clearly differentiate them
from MySQL parameters. MySQL connection parameters mirror the
rekor-server paramters, using a single --dsn flag instead of separate
host, port, password, etc flags.

To do:

- [x] Rename backfill-redis and combine both backfill scripts into one?
- [x] Functional tests for backfill-redis
- [x] Functional tests for backfill-mysql

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -118,6 +118,33 @@ jobs:
           files: /tmp/rekor-merged.cov,/tmp/pkg-rekor-merged.cov
           flags: e2etests
 
+  backfill:
+    runs-on: ubuntu-latest
+    needs: build
+
+    steps:
+      - uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
+      - name: Extract version of Go to use
+        run: echo "GOVERSION=$(cat Dockerfile|grep golang | awk ' { print $2 } ' | cut -d '@' -f 1 | cut -d ':' -f 2 | uniq)" >> $GITHUB_ENV
+      - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+        with:
+          go-version: ${{ env.GOVERSION }}
+      - name: Install backfill test dependencies
+        run: |
+          go install ./cmd/rekor-cli
+          sudo add-apt-repository ppa:savoury1/minisign && sudo apt-get update && sudo apt-get install minisign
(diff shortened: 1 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -118,6 +118,37 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: build
+
+    steps:
+      - uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
+      - name: Extract version of Go to use
+        run: echo "GOVERSION=$(cat Dockerfile|grep golang | awk ' { print $2 } ' | cut -d '@' -f 1 | cut -d ':' -f 2 | uniq)" >> $GITHUB_ENV
+      - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+        with:
+          go-version: ${{ env.GOVERSION }}
```
Hayden-IO, line 131: Can we use `go-version-file: go.mod` instead? Dockerfile and go.mod should stay in sync.

cmd/backfill-index/main.go
```diff
@@ -87,11 +116,23 @@ func main() {
 		os.Exit(0)
 	}
 
-	if *redisHostname == "" {
-		log.Fatal("address must be set")
+	provider := providerUnset
```
Hayden-IO, line 119: nit, should we check that both mysqlDSN and redisHostname aren't set?

docker-compose.backfill-test.yml
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright 2022 The Sigstore Authors.
```
Hayden-IO: 2024

Hayden-IO (commented):
LGTM, fantastic work on the very thorough tests!

.github/workflows/main.yml
cmurphy: I can, but that would make this job inconsistent with every other workflow in rekor, fulcio, and cosign, for example line 89 above. Do you want this changed everywhere?

Hayden-IO: Let's keep it consistent, we can change it in a follow up.

Hayden-IO (approved):

PR #2120, opened 2024-05-15
Add script to delete indexes from Redis

cmurphy:
Add a script to check what indexes are in the EntryIndex table in a MySQL database and deletes those indexes from the Redis database. This is meant to be done after Rekor is migrated to using --search_index.storage_provider=mysql and the MySQL table has been backfilled with all the indexes. Deleting the data allows us to downsize the Redis instance.


Hayden-IO (approved):
Nice work!

PR #2137, opened 2024-06-03
Unset DisableKeepalive for backfill HTTP client

cmurphy:
Disabling Keep-Alive, as done by the default transport setting in the hashicorp cleanhttp package, seems to conflict with a network setting between the public good Rekor instances and the bastion and results in GET requests stalling or timing out after processing a few entries. This change adds an option to the rekor client to unset the DisableKeepalive setting and has the backfill script utilize that option. Other uses of the rekor client will see no change.

### DIFF
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -24,10 +24,11 @@ import (
 type Option func(*options)
 
 type options struct {
-	UserAgent   string
-	RetryCount  uint
-	InsecureTLS bool
-	Logger      interface{}
+	UserAgent           string
+	RetryCount          uint
+	InsecureTLS         bool
+	Logger              interface{}
+	NoDisableKeepalives bool
 }
 
 const (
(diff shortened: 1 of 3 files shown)


Hayden-IO (dismissed):

cmurphy:
@haydentherapper I added another change, this was still hitting a rate limit after 1000 or so requests.

Hayden-IO (approved):
SG, we might want to bypass the LB for the backfill (note the playbook is out of date, there is no longer a global rate limit, so you aren't affecting other clients)

pkg/client/options.go
```diff
@@ -24,10 +24,11 @@ import (
...
 type options struct {
-	UserAgent   string
-	RetryCount  uint
-	InsecureTLS bool
-	Logger      interface{}
+	UserAgent           string
+	RetryCount          uint
+	InsecureTLS         bool
+	Logger              interface{}
+	NoDisableKeepalives bool
```
bobcallaway, line 31: Should this be `EnableKeepalives`?

cmurphy: The field we're un-setting is `DisableKeepalives`. `EnableKeepalives` isn't really correct because it's not enabling `Keepalives`. The double negative seems more intuitive to me.

PR #2146, opened 2024-06-11
Add flags to backfill script

cmurphy:
## Add retry options and make all configurable

Add options to the rekor client to set the minimum and maximum retry
time, and make all rekor client options configurable for the backfill
script.

## Add ability to set custom headers in backfill

Add flags to the backfill script to allow passing custom headers to the
Rekor service. The Rekor infrastructure may be configured to respond
in certain ways to certain headers.

### DIFF
--- a/cmd/backfill-index/main.go
+++ b/cmd/backfill-index/main.go
@@ -161,7 +161,7 @@ func main() {
 		log.Fatalf("creating index client: %v", err)
 	}
 
-	rekorClient, err := client.GetRekorClient(*rekorAddress, client.WithNoDisableKeepalive(true), client.WithRetryCount(10))
+	rekorClient, err := client.GetRekorClient(*rekorAddress, client.WithNoDisableKeepalives(true), client.WithRetryCount(10))
 	if err != nil {
 		log.Fatalf("creating rekor client: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


cmd/backfill-index/main.go
```diff
@@ -161,7 +188,12 @@ func main() {
 		log.Fatalf("creating index client: %v", err)
 	}
 
-	rekorClient, err := client.GetRekorClient(*rekorAddress, client.WithNoDisableKeepalive(true), client.WithRetryCount(10))
+	opts := []client.Option{client.WithNoDisableKeepalives(!*rekorDisableKeepalives)}
```
Hayden-IO, line 191: Bob had mentioned last time, but the double negatives are a bit confusing, especially since we negate the input. Do you think it would be cleaner to switch these to `DisableKeepAlive`?

Hayden-IO (approved):

cmd/backfill-index/main.go
cmurphy: I tried that on the first iteration of this PR, the problem is in `GetRekorClient`, if we had this:

```
defaultTransport.DisableKeepAlives = o.DisableKeepalives
```
as opposed to [conditioning on the negative](https://github.com/sigstore/rekor/pull/2146/files#diff-5b4dad70b1086e28145daaaef5434d221386a6cadd8dcae4f01a9e0eaa8cceaeR41), then if the option is unset it actually reverses the default behavior and several unit tests fail. The double negative helps distinguish whether the setting was unset or explicitly false.

PR #2162, opened 2024-06-26
update builder images to use go1.22.4

cpanato:
#### Summary
- update builder images to use go1.22.4

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor
 
-go 1.21
+go 1.22
 
 require (
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2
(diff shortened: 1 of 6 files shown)


Hayden-IO (commented):
Are we updating https://github.com/sigstore/rekor/blob/main/go.mod#L3 as well?

cpanato:
> Are we updating https://github.com/sigstore/rekor/blob/main/go.mod#L3 as well?

oh let me do that

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor
 
-go 1.21
+go 1.22
```
Hayden-IO, line 3: can you go mod tidy as well? i assume this will add the toolchain directive

cpanato: i did, and did not updated

Hayden-IO (approved):

Issue #2163, opened 2024-06-30
Add option to enable TLS for communication with Trillian


fghanmi:
**Description**
Currently, the communication between Rekor and Trillian server does not support Transport Layer Security. This exposes the system to potential security risks such as data interception and man-in-the-middle attacks. To enhance the security and integrity of services communications, it is imperative to introduce an option to enable TLS - Rekor would ensure the verification of Trillian's certificates.

bobcallaway:
I think this is complete now?

PR #2164, opened 2024-06-30
Add TLS support for Trillian server

fghanmi:
#### Summary
This pull request introduces support for enabling TLS in communications with the Trillian server. By adding a new command-line flag `--tls-ca-cert` and implementing the necessary logic to handle TLS certificates, this update enhances the security of Rekor.

#### Release Note

- Feature: Added support for TLS in communication with the Trillian server.
- New Flag: --tls-ca-cert to specify the CA certificate file path for secure connections.
- Behavior: If the --tls-ca-cert flag is not provided, the system will default to insecure connections.
- Security: This update significantly enhances the security of data in transit by enabling TLS.

Resolves Issue: https://github.com/sigstore/rekor/issues/2163

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -117,7 +117,7 @@ Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("redis_server.password", "", "Redis server password")
 	rootCmd.PersistentFlags().Bool("redis_server.enable-tls", false, "Whether to enable TLS verification when connecting to Redis endpoint")
 	rootCmd.PersistentFlags().Bool("redis_server.insecure-skip-verify", false, "Whether to skip TLS verification when connecting to Redis endpoint, only applicable when 'redis_server.enable-tls' is set to 'true'")
-	rootCmd.PersistentFlags().String("tls_ca_cert", "", "Certificate file to use for secure connections with Trillian server")
+	rootCmd.PersistentFlags().String("trillian_log_server.tls_ca_cert", "", "Certificate file to use for secure connections with Trillian server")
 	rootCmd.PersistentFlags().Bool("trillian_log_server.tls", false, "Use TLS when connecting to Trillian Server")
 
 	rootCmd.PersistentFlags().Bool("enable_attestation_storage", false, "enables rich attestation storage")
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -117,6 +117,8 @@ Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("redis_server.password", "", "Redis server password")
 	rootCmd.PersistentFlags().Bool("redis_server.enable-tls", false, "Whether to enable TLS verification when connecting to Redis endpoint")
 	rootCmd.PersistentFlags().Bool("redis_server.insecure-skip-verify", false, "Whether to skip TLS verification when connecting to Redis endpoint, only applicable when 'redis_server.enable-tls' is set to 'true'")
+	rootCmd.PersistentFlags().String("tls_ca_cert", "", "Certificate file to use for secure connections with Trillian server")
+	rootCmd.PersistentFlags().Bool("trillian_log_server.tls", false, "Use TLS when connecting to Trillian Server")
 
 	rootCmd.PersistentFlags().Bool("enable_attestation_storage", false, "enables rich attestation storage")
 	rootCmd.PersistentFlags().String("attestation_storage_bucket", "", "url for attestation storage bucket")
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -22,12 +22,15 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/google/trillian"
 	"github.com/redis/go-redis/v9"
 	"github.com/spf13/viper"
 	"golang.org/x/exp/slices"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/sigstore/rekor/pkg/indexstorage"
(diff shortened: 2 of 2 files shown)


cpanato (dismissed):
lgtm

cmd/rekor-server/app/root.go
```diff
@@ -117,6 +117,7 @@ Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("redis_server.password", "", "Redis server password")
 	rootCmd.PersistentFlags().Bool("redis_server.enable-tls", false, "Whether to enable TLS verification when connecting to Redis endpoint")
 	rootCmd.PersistentFlags().Bool("redis_server.insecure-skip-verify", false, "Whether to skip TLS verification when connecting to Redis endpoint, only applicable when 'redis_server.enable-tls' is set to 'true'")
+	rootCmd.PersistentFlags().String("tls-ca-cert", "", "Certificate file to use for secure connections with Trillian server")
```
bobcallaway: should we also add another boolean parameter `trillian_log_server.tls`, which would indicate the use of TLS with trust delegated to the system trust store?

pkg/api/api.go
```diff
@@ -47,7 +50,25 @@ import (
 
 func dial(rpcServer string) (*grpc.ClientConn, error) {
 	// Set up and test connection to rpc server
-	creds := insecure.NewCredentials()
+	var creds credentials.TransportCredentials
+	tlsCACertFile := viper.GetString("tls-ca-cert")
+	if tlsCACertFile == "" {
+		creds = insecure.NewCredentials()
+	} else {
```
bobcallaway: would love to see some tests for this within the docker-compose setup

cmd/rekor-server/app/root.go
```diff
@@ -117,6 +117,8 @@ Memory and file-based signers should only be used for testing.`)
...
+	rootCmd.PersistentFlags().Bool("trillian_log_server.tls", false, "Use system trust store for TLS")
```
```suggestion
	rootCmd.PersistentFlags().Bool("trillian_log_server.tls", false, "Use TLS when connecting to Trillian")
```

bobcallaway (commented):
lgtm minus lint issue, thanks for adding this!

cmd/rekor-server/app/root.go
fghanmi: @bobcallaway, Actually, trillian_log_server.tls means to use the local CA certificates on the system (as you suggested: "trust delegated to the system trust store")
So, TLS would be enabled if trillian_log_server.tls == true OR tls_ca_cert != "" 
This is what you meant, right ?

bobcallaway: right. `insecure.NewCredentials()` disables transport-level security on the connection, whereas this new boolean flag would enable it with SSL but not require a user to specify a server certificate

tests/tls/ca.crt
```diff
@@ -0,0 +1,29 @@
+-----BEGIN CERTIFICATE-----
```
bobcallaway:
```Validity
            Not Before: Jul  2 22:26:18 2024 GMT
            Not After : Aug 31 22:26:18 2024 GMT
```

Can we either make this expire quite far out in the future or generate the cert ephemerally on each test run?

pkg/api/api.go
```diff
@@ -47,7 +50,33 @@ import (
...
 	// Set up and test connection to rpc server
-	creds := insecure.NewCredentials()
+	var creds credentials.TransportCredentials
+	tlsCACertFile := viper.GetString("tls_ca_cert")
+	useSystemTrustStore := viper.GetBool("trillian_log_server.tls")
+
+	switch {
+	case useSystemTrustStore:
+		creds = credentials.NewTLS(&tls.Config{
+			ServerName: rpcServer,
```
bobcallaway, line 60: it's unlikely, but possible that `rpcServer` will not simply be a `hostname/IP address` or `hostname:port` combination, as that string can be a value as defined by https://github.com/grpc/grpc/blob/master/doc/naming.md

pkg/api/api.go
```diff
@@ -47,7 +50,33 @@ import (
...
+			MinVersion: tls.VersionTLS12,
+		})
+	case tlsCACertFile == "":
+		creds = insecure.NewCredentials()
+	default:
```
bobcallaway: for readability i'd swap the logic here. make the default behavior `insecure.NewCredentials()` similar to before, and `case tlsCACertFile != "":` for the new case

fghanmi:
@bobcallaway I see that the e2e tests are failing, are they all using ./docker-compose.yaml resources ? 
Meaning, should I enable TLS on all the /docker-compose*.yaml files ? ( since the trillian logserver in docker-compose.yaml is TLS enabled)

fghanmi:
**Regarding the build failure: CI / issue-872-e2e:**
The script is using an old rekor-server image https://github.com/sigstore/rekor/blob/main/tests/issue-872-e2e-test.sh#L71
that does not have the new tag `--tls_ca_cert` and thus, it fails.
@bobcallaway  is it possible to use an image with the new updates ?

bobcallaway:
> **Regarding the build failure: CI / issue-872-e2e:** The script is using an old rekor-server image https://github.com/sigstore/rekor/blob/main/tests/issue-872-e2e-test.sh#L71 that does not have the new tag `--tls_ca_cert` and thus, it fails. @bobcallaway is it possible to use an image with the new updates ?

no, that specific test is for a regression that was started at that back version. 

I'll need to take a closer look at the dependency structure of those various docker-compose files to better advise you.

fghanmi:
> > **Regarding the build failure: CI / issue-872-e2e:** The script is using an old rekor-server image https://github.com/sigstore/rekor/blob/main/tests/issue-872-e2e-test.sh#L71 that does not have the new tag `--tls_ca_cert` and thus, it fails. @bobcallaway is it possible to use an image with the new updates ?
> 
> no, that specific test is for a regression that was started at that back version.
> 
> I'll need to take a closer look at the dependency structure of those various docker-compose files to better advise you.

We can create another trillian server `trillian-log-server-no-tls` and use it as the trillian server for the `rekor-server-issue-872-v060` that uses an older image. (I've tested it)
What do you think ?

cmd/rekor-server/app/root.go
```diff
@@ -117,6 +117,8 @@ Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("redis_server.password", "", "Redis server password")
 	rootCmd.PersistentFlags().Bool("redis_server.enable-tls", false, "Whether to enable TLS verification when connecting to Redis endpoint")
 	rootCmd.PersistentFlags().Bool("redis_server.insecure-skip-verify", false, "Whether to skip TLS verification when connecting to Redis endpoint, only applicable when 'redis_server.enable-tls' is set to 'true'")
+	rootCmd.PersistentFlags().String("tls_ca_cert", "", "Certificate file to use for secure connections with Trillian server")
```
```suggestion
	rootCmd.PersistentFlags().String("trillian_log_server.tls_ca_cert", "", "Certificate file to use for secure connections with Trillian server")
```

bobcallaway:
Let's do this for now; remove the test coverage from this PR and please open an issue on the repo to add back coverage later.

fghanmi:
> Let's do this for now; remove the test coverage from this PR and please open an issue on the repo to add back coverage later.

Done.
Issue created: https://github.com/sigstore/rekor/issues/2188

cmd/rekor-server/app/root.go
bobcallaway: Given that both flags relate to just the Trillian connection, let’s update this. Otherwise lgtm and we’ll merge.

bobcallaway (approved):

PR #2203, opened 2024-08-08
Fix index scripts

cmurphy:
Fix the cleanup-index script for indexes with capitalized keys
(e.g., workload identities for GitHub repositories with capital letters in the name).

Add a new script alongside the backfill script to copy indices directly
from Redis to MySQL. This is created as a separate script from the
backfill script because they are necessary for different purposes:
backfill is appropriate for when there is no data in any search index
backend and the only source of data is the Rekor API, where copy is
apprpriate when there is index data in the Redis database to pull
directly from. This enables the script to copy data that the backfill
script would not be able to recreate because the data needed to generate
the indexes is not always persisted to the Rekor API.

### DIFF
--- a/tests/copy-index-test.sh
+++ b/tests/copy-index-test.sh
@@ -33,24 +33,28 @@ source $(dirname "$0")/index-test-utils.sh
 
 trap cleanup EXIT
 
-remove_keys() {
+make_intoto_entries() {
     set -e
-    for i in $@ ; do
-        local rekord=$(rekor-cli --rekor_server $REKOR_ADDRESS get --log-index $i --format json)
-        local uuid=$(echo $rekord | jq -r .UUID)
-        local sha=sha256:$(echo $rekord | jq -r .Body.RekordObj.data.hash.value)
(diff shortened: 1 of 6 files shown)


tests/copy-index-test.sh
```diff
@@ -0,0 +1,158 @@
...
+
+source $(dirname "$0")/index-test-utils.sh
+
+trap cleanup EXIT
+
+make_intoto_entries() {
+    set -e
+    for v in 0.0.1 0.0.2 ; do
+        rekor-cli --rekor_server $REKOR_ADDRESS upload \
+            --type intoto:${v} \
```
Hayden-IO: Can we also add a test for the `dsse` type?

Hayden-IO (dismissed):
Looks great! Just one suggested test case to add

tests/copy-index-test.sh
cmurphy: done

Hayden-IO (approved):

PR #2214, opened 2024-08-21
Add public key hash check in Signed Note verification

### LINKED ISSUE #389: Signed Note verification question
When verifying SignedNotes, we never check that the loaded public key matches the hash. 

I think the only purpose is for easy lookup and early rejection in case none of the known public keys match the hash in the `note.Signature`. Shouldn't be a bad thing, right? WDYT?
(e.g. this would fail: a valid signature for a loaded public key, but the note.Signature specifies a different public key.)

https://github.com/sigstore/rekor/blob/07c8e8fda381c18af6d86dca273d4494bfd4cb5e/pkg/util/signed_note.go#L75

@bobcallaway

hojoungjang:
* Resolves sigstore/rekor#389

#### Summary
This PR adds a public key hash check in `SignedNote.Verify` method.
The hash stored in `note.Signature` is compared against the verifier's public key hash. If the hash mismatches, the verification fails.

#### Release Note
* `SignedNote.Sign` is updated to use `GetPublicKeyHash` function.
* `GetPublicKeyHash` is added to calculate sha256 hash of the input public key (copied the existing logic in `SignedNote.Sign` method)
* `SignedNote.Verify` is updated to compare hash stored in `note.Signature` is and the verifier's public key hash.

### DIFF
--- a/pkg/util/signed_note.go
+++ b/pkg/util/signed_note.go
@@ -18,6 +18,7 @@ package util
 import (
 	"bufio"
 	"bytes"
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/ed25519"
 	"crypto/rsa"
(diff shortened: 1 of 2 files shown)


pkg/util/signed_note.go
```diff
@@ -190,3 +199,13 @@ func SignedNoteValidator(strToValidate string) bool {
 	s := SignedNote{}
 	return s.UnmarshalText([]byte(strToValidate)) == nil
 }
+
+func GetPublicKeyHash(publicKey crypto.PublicKey) (uint32, error) {
```
Hayden-IO: Can you make this an unexported function `getPublicKeyHash`?

hojoungjang:
@haydentherapper Thanks! I have addressed your comment.

Hayden-IO (approved):
Thanks!

hojoungjang:
Thanks for the review. I will leave this as it is for now since I don't have the permission to merge. Please let me know if there is anything I can do.

PR #2228, opened 2024-09-25
Add support for signing with Tink keyset

Hayden-IO:
This allows deployers to securely sign in-memory while mitigating key exfiltration, since the key is encrypted at rest and loaded into memory at server startup.

Requires providing a path to an encrypted Tink keyset and the Key Encryption Key, a KMS URI for decrypting the keyset.

Heavily pulls from Fulcio's Tink implementation.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -98,9 +98,11 @@ func init() {
 	rootCmd.PersistentFlags().String("rekor_server.address", "127.0.0.1", "Address to bind to")
 
 	rootCmd.PersistentFlags().String("rekor_server.signer", "memory",
-		`Rekor signer to use. Valid options are: [awskms://keyname, azurekms://keyname, gcpkms://keyname, hashivault://keyname, memory, <filename containing PEM-encoded private key>].
+		`Rekor signer to use. Valid options are: [awskms://keyname, azurekms://keyname, gcpkms://keyname, hashivault://keyname, memory, tink, <filename containing PEM-encoded private key>].
 Memory and file-based signers should only be used for testing.`)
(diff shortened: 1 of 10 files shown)


cpanato (approved):
thanks
lgtm

cmd/rekor-server/app/root.go
```diff
@@ -98,9 +98,11 @@ func init() {
 	rootCmd.PersistentFlags().String("rekor_server.address", "127.0.0.1", "Address to bind to")
 
 	rootCmd.PersistentFlags().String("rekor_server.signer", "memory",
-		`Rekor signer to use. Valid options are: [awskms://keyname, azurekms://keyname, gcpkms://keyname, hashivault://keyname, memory, <filename containing PEM-encoded private key>].
+		`Rekor signer to use. Valid options are: [awskms://keyname, azurekms://keyname, gcpkms://keyname, hashivault://keyname, memory, tink, <filename containing PEM-encoded private key>].
 Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("rekor_server.signer-passwd", "", "Password to decrypt signer private key")
+	rootCmd.PersistentFlags().String("rekor_server.tink_kek_uri", "", "Key encryption key for decrypting Tink keyset. Valid options are [aws-kms://keyname, gcp-kms://keyname]")
```
bobcallaway, line 104: could we use `awskms` and `gcpkms` to match above?

Hayden-IO: No unfortunately, these are the values hardcoded in Tink.

Hayden-IO: https://github.com/tink-crypto/tink-go-gcpkms/blob/main/integration/gcpkms/gcp_kms_client.go#L33C2-L33C11

bobcallaway (approved):

PR #2262, opened 2024-10-30
add application/x-sigstore-tle as response type

### LINKED ISSUE #2218: Add TLE response media type
**Description**

We should support returning a response formatted as a `TransparencyLogEntry` (TLE) per the specification. This will be used by Sigstore clients who currently have to transform the Rekor response into the spec-compliant struct.

We should be able to implement this as a custom response type in OpenAPI - https://swagger.io/docs/specification/2-0/describing-responses/

bobcallaway:
Fixes: #2218

### DIFF
--- a/e2e-test.sh
+++ b/e2e-test.sh
@@ -31,7 +31,7 @@ docker build -t gcp-pubsub-emulator -f Dockerfile.pubsub-emulator .
 docker kill $(docker ps -q) || true
 
 echo "starting services"
-${docker_compose} up -d --build --force-recreate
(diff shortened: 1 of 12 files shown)


bobcallaway:
after reviewing, there's a problem in that we cannot take a TLE and convert back into a `models.LogEntry` because we don't have the tree ID to correctly set the entryUUID. While most tests still pass, others expect the full entryUUID and therefore I think this will significantly break clients. Given the flaky nature of go-openapi's `Accept` header support, and this fact, I think we'll have to punt this to the v2 api.

Issue #2279, opened 2024-11-20
Avoid reprocessing data in backfill-index.


JasonPowr:
**Description**

Currently, Backfill-index reprocesses data that has already been indexed during previous runs. This leads to unnecessary processing and inefficiency.

#### Proposed Solution
Add a way to store the index of the last processed key so we only need to process new entries.

PR #2280, opened 2024-11-20
Avoid reprocessing data in backfill-index

JasonPowr:
Issue: https://github.com/sigstore/rekor/issues/2279

##### Motivation
This pr helps to solve the issue of reprocessing already indexed data during repeated runs of the backfill-index job by adding a --enable-redis-index-resume flag to backfill-index. When enabled, backfill-index saves the last processed index in Redis and resumes from it in subsequent runs, processing only new entries. 

#### Release Note
Added a --enable-redis-index-resume flag to backfill-index, enabling jobs to resume from the last processed index and process only new entries for improved efficiency.

### DIFF
--- a/cmd/backfill-index/main.go
+++ b/cmd/backfill-index/main.go
@@ -38,6 +38,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -90,6 +91,10 @@ const (
 	providerMySQL
 )
 
+const (
+	redisLastProcessedIndexKey = "last_processed_index"
+)
+
 type indexClient interface {
 	idempotentAddToIndex(ctx context.Context, key, value string) error
 }
@@ -127,6 +132,7 @@ var (
 	redisPassword           = flag.String("redis-password", "", "Password for Redis authentication")
 	redisEnableTLS          = flag.Bool("redis-enable-tls", false, "Enable TLS for Redis client")
 	redisInsecureSkipVerify = flag.Bool("redis-insecure-skip-verify", false, "Whether to skip TLS verification for Redis client or not")
+	enableRedisIndexResume  = flag.Bool("enable-redis-index-resume", false, "Enable resuming from the last processed index stored in Redis. When enabled, the '--start' flag becomes optional for the Redis provider.")
 	mysqlDSN                = flag.String("mysql-dsn", "", "MySQL Data Source Name")
 	startIndex              = flag.Int("start", -1, "First index to backfill")
 	endIndex                = flag.Int("end", -1, "Last index to backfill")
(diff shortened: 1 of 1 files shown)


Hayden-IO (commented):
Can you use `--start`?

JasonPowr:
> Can you use `--start`?

Hi @haydentherapper, normally yes, however we tend to run the backfill-index on a schedule e.g a cronjob  and keeping '--start' update to date in this context tends to be awkward, we either have to manually do it or store the value for '--start' in an awkward way which could be overwritten or modified.

JasonPowr:
@haydentherapper When you get a chance can I get a review on this :), thanks

Hayden-IO:
What is the use case for running this periodically? It’s expected to be run exceptionally, when there is an outage that’s caused a loss of index entries.

How would you handle —end? If this is a cronjob, are you always processing up the last entry?

JasonPowr:
> What is the use case for running this periodically?

It was recommend to us to run the cronjob periodically, when we were first getting started with sigsore, as if there are errors writting to the index for any reason, we can make sure that it always manages to stay up to date.

> How would you handle —end? If this is a cronjob, are you always processing up the last entry?
Essentially we are curling the rekor api and using the treeSize as the --end value 

Hope that helps to clear things up :)

Hayden-IO:
> It was recommend to us to run the cronjob periodically

This shouldn't be needed. Failures would be exceptional, e.g. if Redis goes down, and they can be monitored for in the logs as well.

Technically, I don't have an issue with this PR, I just don't think it's necessary. I'm happy to merge this if you still need to be reprocessing data.

JasonPowr:
Apologies @haydentherapper linter was giving me trouble :)

> Technically, I don't have an issue with this PR, I just don't think it's necessary. I'm happy to merge this if you still need to be reprocessing data.

Thank you that would be great :) and thanks for you input also :)

Hayden-IO (approved):

cmurphy:
This is setting the last filled index to the `--end` index, which is already known to the script caller since it must be provided. Why not just have the cron job set this value itself when the script completes, and retrieve it to provide to `--start` when calling the script again?

JasonPowr:
> Why not just have the cron job set this value itself when the script completes, and retrieve it to provide to --start when calling the script again?

@cmurphy So something like that is definitely possible, however I feel like storing it in this manner is much more reliable, it also makes it more difficult to modify if someone wanted to, given that it could be locked behind a auth, as opposed to written to some env var or file somewhere

cmurphy:
It could just write to the redis instance itself? With redis-cli and the auth credentials you're already passing to the script.

JasonPowr:
@cmurphy So I did test this and this does work, I appreciate the suggestion never thought of doing it this way :) 

However, I do still think this pull request would be useful to anyone else who may have a similar use case to us, instead of everyone using backfill-index coming up with there own solution, we could all use this common one instead, wdyt?

cmurphy:
> this pull request would be useful to anyone else who may have a similar use case to us

I'm not sure this is a very common use case. When Rekor is working properly, it populates the indexes itself. This script was only ever meant to be used in case of outages or migrating the entire database.

JasonPowr:
Hey Guys :) thanks for the comments on this pr, after some discussions we will be going with the solution proposed in this comment, https://github.com/sigstore/rekor/pull/2280#issuecomment-2532591794, so I am going to close this pr. Thanks again :)

PR #2322, opened 2025-01-09
Update types.md

sgpinkus:
Add link to HashedRekord


Hayden-IO (approved):
Thanks! Note that this doc is out of date and is missing a number of new types, and also conflates entry types (e.g. hashedrekord) with signature verifiers (e.g. ssh, x509)

PR #2323, opened 2025-01-09
chore: relax go directive to permit 1.22.x

dnwe:
#### Summary

Setting this to 1.23.4 requires all consumers to be building with the absolute latest 1.23.4 or newer release of Go 1.23 and to update their own go.mod accordingly — this seems unnecessarily restrictive for a library module, particularly as the code itself doesn't currently use any modern language constructs and builds fine even with older Go versions.

Instead set the go directive to 1.22.0 and use the toolchain directive to recommend the latest 1.22.x when building locally.

Note: this also upgrades google/trillian to v1.7.1 and downgrades sigs.k8s.io/release-utils to v0.8.4, which are both required in order to allow the go directive to be relaxed here.

#### Release Note

NONE

#### Documentation

NONE


dnwe:
See also [this comment](https://github.com/sigstore/fulcio/issues/1899#issuecomment-2572116404) around standardising the sigstore repos, and https://github.com/sigstore/sigstore/pull/1878

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sigstore/rekor
 
-go 1.23.4
+go 1.22.0
```
Hayden-IO, line 3: Thanks for looking into this @dnwe!

Given that 1.24 is coming out next month, and from past announcements, it looks like Go backports security fixes only one version, can we use 1.23.0 instead, to avoid having to bump the Go version next month?

I'm not certain what the best practice is here to keep up with the latest bug/sec fixes. I'm assuming the minor version in go.mod should match what's pulled in setup-go and toolchain, which is why I suggested this.

Hayden-IO: If I'm understanding go versioning correctly, libraries should target the minimum version needed based on standard library functionality (e.g. for sigstore/sigstore), while binaries should target the highest to pick up the latest sec/bug fixes (e.g. fulcio, rekor, cosign). Though having both a library and a CLI/service in the same repo, like in Cosign or Rekor, would mean we need to target a higher version. 

_I think?_

dnwe: For cli/services the toolchain directive can be used to indicate “build with the latest go” rather than having the go directive higher

Hayden-IO: OK, I think I get it now. So across all Go Sigstore libraries, CLIs, and services, we need to:

1. Set go.mod version to 1.22.0 (or higher if needed due to a dependency). Maybe this could be 1.21.0 or lower as well, but if there's no issue sticking with 1.22 for now, I'd keep it given the new for-loop language feature.
2. Set the `toolchain` directive to be the latest 1.23.x patch release and keep it up to date.
3. Hardcode actions/setup-go to be '1.23', which will pull the latest patch release of 1.23.x. (there was an issue I found about using the toolchain directive to derive version, but it doesn't seem to be implemented yet)
4. Update the Dockerfile container to be the latest patch release of Go (already automated with dependabot)
5. Update cloudbuild container reference to the latest patch release (ghcr.io/gythialy/golang-cross)
6. When 1.24 comes out, update setup-go and toolchain references, and the container references.

We'll update go.mod to >= 1.23 in the future only if we have a language feature we need.

Does this seem correct?

dnwe: Yes! exactly right 🥇 

One slight twweak RE: "Hardcode actions/setup-go to be '1.23'" I believe rekor actually syncs setup-go against what is defined in the Dockerfile, which then tracks against what dependabot has updated the Dockerfile to:

```yaml
      - name: Extract version of Go to use
        run: echo "GOVERSION=$(cat Dockerfile|grep golang | awk ' { print $2 } ' | cut -d '@' -f 1 | cut -d ':' -f 2 | uniq)" >> $GITHUB_ENV

      - uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
        with:
          check-latest: true
 ```

You might prefer to keep that model as it is a good way of syncing your test/CI runs against what's shipping in the Dockerfile. It also means you'll get somewhat more reproducible builds and you'll have a concrete pull request whenever you bump a minor version

Comment (dnwe) [reply go.mod:3]: (Aside: raised https://github.com/sigstore/rekor/pull/2325 to slightly simplify that GOVERSION extraction)

File: go.mod [hunk 2]
```diff
@@ -1,6 +1,8 @@
...
+
+toolchain go1.22.10
```
Hayden-IO: Can we set this to `1.23.4?`

Hayden-IO (commented):
LGTM, just one question about toolchain version, thanks!

go.mod
dnwe: Sure

Hayden-IO (approved):
Thank you! @cpanato any other comments?

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sigstore/rekor
 
-go 1.23.4
+go 1.22.0
+
+toolchain go1.23.4
```
cpanato, line 5: this will make/force to download Go1.23.4 and then you will not build with go1.22 anymore

so i think here we need to be in go1.23.4 and the toolchain request that

hack/tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/rekor/hack/tools
 
-go 1.23.2
+go 1.22.0
```
cpanato, line 3: why we are downgrading and not keep in go1.23?

cpanato (changes requested):
have some comments

cpanato:
go1.22 will be deprecated very soon, go1.24 is around the corner

go.mod
dnwe: I think you're possibly misunderstanding here? The `go` directive specifies the api compatibility of the module and the `toolchain` directive specifies a suggested/recommended toolchain to use when building. Only the former has an effect on Go code that imports rekor as a module. The latter only applies as a hint to the rekor build itself and even then it is advisory and can be superseded in various ways.

hack/tools/go.mod
dnwe: We're not downgrading, we are permissively declaring "this go.mod file and the module source code is API compatible with the Go standard library and toolchain from go version 1.22.0 onwards"

dnwe:
> go1.22 will be deprecated very soon, go1.24 is around the corner

go1.24 is still a month away and we _can_ trivially bump the go directive again when it releases if you feel that is desirable — although it is somewhat unnecessary. The idea here is to cut a versioned release that people using go1.22.x can upgrade to

go.mod
cpanato: @dnwe thanks for the clarification but my understanding was that it will require updating go when building, and if any downstream projects import rekor they will be required to use the go 1.23 as well and not 1.22

But I think I did not understand that. I think I am totally wrong. This isn't very clear to me.

dnwe: Yeah for anything importing rekor toolchain is completely ignored, that’s only used if people go install rekor standalone or clone and build the repo (and even then only if their GOTOOLCHAIN contains auto)

Hayden-IO: From what I've learned, the motivation to set `toolchain` is so that local development matches what we build the release binaries with, given we're keeping the Dockerfile Go container up to date (to pull in the latest security patches), and we'll keep `toolchain` in sync.

We should set the go.mod version to be as low as possible, based on if we need features from a newer version of Go. This is for libraries that depend on Rekor, which they themselves can choose to build with either the newest Go binary or one that matches the go.mod.

Hayden-IO:
@cpanato, since we're not using any features from 1.23.x, we can set go.mod to 1.22.x. At some point, we'll update only once we need a feature from the standard library (for example, support for iterators), or once a dependency is bumped up to >= 1.23. Security fixes are pulled in based on the Go binary we build with.

I laid out this plan here - https://github.com/sigstore/rekor/pull/2323#discussion_r1909643825

go.mod
cpanato: thank you

cpanato:
thank you, for all clarifications

go.mod
kaovilai: +1 toolchain does not cascade to imports, go directive can stay 1.22 alongside using a newer toolchain to build.

cpanato (approved):

dnwe:
@cpanato please can you start the process to cut a v1.3.8 tag of rekor so we can start pulling this fix into the consuming modules like sigstore-go?

cpanato:
> @cpanato please can you start the process to cut a v1.3.8 tag of rekor so we can start pulling this fix into the consuming modules like sigstore-go?

add a changelog https://github.com/sigstore/rekor/pull/2331 lets wait for @haydentherapper to we sync

PR #2330, opened 2025-01-15
Support per-shard signing keys

Hayden-IO:
This change enables key rotation with a per-shard
signing key configuration. The LogRanges structure
now holds both active and inactive shards, with the
LogRange structure containing a signer, encoded
public key and log ID based on the public key.

This change is backwards compatible. If no signing
configuration is specified, the active shard
signing configuration is used for all shards.

Minor change: Standardized log ID vs tree ID, where
the former is the pubkey hash and the latter is the
ID for the Trillian tree.

### DIFF
--- a/cmd/rekor-cli/app/get.go
+++ b/cmd/rekor-cli/app/get.go
@@ -115,7 +115,7 @@ var getCmd = &cobra.Command{
 				}
 				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
 				if err != nil {
-					return nil, fmt.Errorf("retrieving rekor public key: %v", err)
+					return nil, fmt.Errorf("retrieving rekor public key: %w", err)
 				}
 				// verify log entry
 				e = entry
@@ -154,7 +154,7 @@ var getCmd = &cobra.Command{
 				}
 				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
 				if err != nil {
-					return nil, fmt.Errorf("retrieving rekor public key: %v", err)
+					return nil, fmt.Errorf("retrieving rekor public key: %w", err)
 				}
 
 				if err := compareEntryUUIDs(params.EntryUUID, k); err != nil {
--- a/cmd/rekor-cli/app/get.go
+++ b/cmd/rekor-cli/app/get.go
@@ -92,11 +92,6 @@ var getCmd = &cobra.Command{
 		if logIndex == "" && uuid == "" {
 			return nil, errors.New("either --uuid or --log-index must be specified")
 		}
-		// retrieve rekor pubkey for verification
-		verifier, err := loadVerifier(rekorClient)
-		if err != nil {
-			return nil, fmt.Errorf("retrieving rekor public key")
-		}
 
 		if logIndex != "" {
 			params := entries.NewGetLogEntryByIndexParams()
@@ -113,6 +108,15 @@ var getCmd = &cobra.Command{
 			}
 			var e models.LogEntryAnon
 			for ix, entry := range resp.Payload {
+				// retrieve rekor pubkey for verification
+				treeID, err := sharding.TreeID(ix)
+				if err != nil {
+					return nil, err
+				}
+				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
+				if err != nil {
+					return nil, fmt.Errorf("retrieving rekor public key")
+				}
 				// verify log entry
 				e = entry
 				if err := verify.VerifyLogEntry(ctx, &e, verifier); err != nil {
@@ -143,6 +147,16 @@ var getCmd = &cobra.Command{
 
 			var e models.LogEntryAnon
 			for k, entry := range resp.Payload {
+				// retrieve rekor pubkey for verification
+				treeID, err := sharding.TreeID(k)
+				if err != nil {
+					return nil, err
+				}
+				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
+				if err != nil {
+					return nil, fmt.Errorf("retrieving rekor public key")
+				}
+
 				if err := compareEntryUUIDs(params.EntryUUID, k); err != nil {
 					return nil, err
 				}
--- a/cmd/rekor-cli/app/get.go
+++ b/cmd/rekor-cli/app/get.go
@@ -115,7 +115,7 @@ var getCmd = &cobra.Command{
 				}
 				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
 				if err != nil {
-					return nil, fmt.Errorf("retrieving rekor public key")
+					return nil, fmt.Errorf("retrieving rekor public key: %v", err)
 				}
 				// verify log entry
 				e = entry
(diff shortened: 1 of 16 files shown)


Hayden-IO:
Ready for review. I'm going to keep playing around with making CodeQL happy but I might just dismiss the alert, it's a FP. (Edit: Dismissed)

cmd/rekor-cli/app/get.go
```diff
@@ -143,6 +147,16 @@ var getCmd = &cobra.Command{
...
 			var e models.LogEntryAnon
 			for k, entry := range resp.Payload {
+				// retrieve rekor pubkey for verification
+				treeID, err := sharding.TreeID(k)
+				if err != nil {
+					return nil, err
+				}
+				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
+				if err != nil {
+					return nil, fmt.Errorf("retrieving rekor public key")
```
bobcallaway: worth wrapping the original `err`?

cmd/rekor-cli/app/get.go
```diff
@@ -113,6 +108,15 @@ var getCmd = &cobra.Command{
...
 			var e models.LogEntryAnon
 			for ix, entry := range resp.Payload {
+				// retrieve rekor pubkey for verification
+				treeID, err := sharding.TreeID(ix)
+				if err != nil {
+					return nil, err
+				}
+				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
+				if err != nil {
+					return nil, fmt.Errorf("retrieving rekor public key")
```
bobcallaway: worth wrapping the original `err`?

pkg/api/entries.go
```diff
@@ -88,19 +88,24 @@ func logEntryFromLeaf(ctx context.Context, signer signature.Signer, _ trilliancl
...
 
 	virtualIndex := sharding.VirtualLogIndex(leaf.GetLeafIndex(), tid, ranges)
+	logRange, err := ranges.GetLogRangeByTreeID(tid)
+	if err != nil {
+		return nil, err
+	}
+
 	logEntryAnon := models.LogEntryAnon{
-		LogID:          swag.String(api.pubkeyHash),
+		LogID:          swag.String(logRange.PemPubKey),
```
```suggestion
		LogID:          swag.String(logRange.LogID),
```

bobcallaway (commented):
a couple comments; i want to think through the rollout implications of this in an environment where you have multiple instances of rekor running in parallel where they each might see a different signing config concurrently.

cmd/rekor-cli/app/get.go
```diff
@@ -115,7 +115,7 @@ var getCmd = &cobra.Command{
 				}
 				verifier, err := loadVerifier(rekorClient, strconv.FormatInt(treeID, 10))
 				if err != nil {
-					return nil, fmt.Errorf("retrieving rekor public key")
+					return nil, fmt.Errorf("retrieving rekor public key: %v", err)
```
bobcallaway:
```suggestion
					return nil, fmt.Errorf("retrieving rekor public key: %w", err)
```

cmd/rekor-cli/app/get.go
```diff
@@ -154,7 +154,7 @@ var getCmd = &cobra.Command{
...
```
bobcallaway:
```suggestion
					return nil, fmt.Errorf("retrieving rekor public key: %w", err)
```

Hayden-IO:
> a couple comments; i want to think through the rollout implications of this in an environment where you have multiple instances of rekor running in parallel where they each might see a different signing config concurrently.

tl;dr - I see no issues.

Walking through a few examples:

### Adding missing signing configurations to prod

Note that the active shard configuration is created dynamically, it is not specified in `trillian_log_server.sharding_config`. It is composed from the tree ID from `trillian_log_server.tlog_id` and the signer from `rekor_server.signer`. The inactive shards come from `trillian_log_server.sharding_config`.

When no signing config in `trillian_log_server.sharding_config` is specified, the server falls back to the active key. It's safe to have some instances with no signing config specified and some with it specified - in both cases, the key material is the same.

### Rolling out a new shard

I've reviewed the sharding procedure and there should be no issue. The only change is when we [create an active tree](https://github.com/sigstore/public-good-instance/blob/main/playbooks/sharding-prod.md#create-a-new-active-tree), the sharding config for the `CURRENT_TREE_ID` **must** specify a signing config if the key is being rotated.

1. Spin Rekor down to 1 instance. This prevents race conditions.
1. `rekor.server.config.treeID` is set to `""`, and the `createtree` job is forced. `rekor.server.sharding.contents` is updated to include the `CURRENT_TREE_ID` **AND** the signing config, which for prod will be a KMS key.
1. The `createjob` will populate the configmap `rekor-config.treeID`.
1. When Rekor starts up, helm will set the server arg `trillian_log_server.tlog_id` from the env var `TREE_ID` which comes from `rekor-config.treeID`. Rekor will dynamically create the active shard config from `tlog_id` and `signer` as noted above. Rekor will also have the signing config for the now-inactive shard.
1. Update `rekor.server.config.treeID` to the new tree ID, set `createtree` to false. Next time Rekor spins up, the configmap `rekor-config.treeID` will come from `rekor.server.config.treeID` rather than being set by `createtree`. [^1]
1. Spin Rekor up to 3 instances.
1. Set `updateTree` to run, which freezes `CURRENT_TREE_ID`. This just calls Trillian, Rekor has no knowledge if an inactive shard is frozen or not.
1. Disable `updateTree`.

### Rolling out a new shard without spinning Rekor down to 1 instance

This would cause issues outside of signing, so this is moot. While some instances continue to write to the old tree, new instances would write to the new tree. As new instances start up, they would have different tree lengths of inactive shards, which would lead to incorrect calculations on global log indices.

For the signing config, there would be no issue. Inactive shards would be specified to use the old signing key, while the new shard would use the KMS key passed as a server arg.

### Updating a signing config without sharding

This would never happen, you can't rotate the key of an active or inactive tree without breaking verification.

[^1]: Somewhat frighteningly, if you don't set `rekor.server.config.treeID`, Rekor [creates a new tree](https://github.com/sigstore/rekor/blob/1e9cdae679740658984c849aeaf3c4ad4532d714/pkg/api/api.go#L126-L134). This is likely used during testing. I think we should gate this behind a flag, because if we accidentally forget to set `config.treeID`, Rekor will create a new shard. Though, I'm not sure the `createtree` job is actually necessary, you could just let Rekor generate a new tree, then grab the tree ID and update the configmap value.

pkg/sharding/ranges.go
```diff
@@ -200,15 +258,11 @@ func (l *LogRanges) PublicKey(activePublicKey, treeID string) (string, error) {
...
-			// assume the active public key if one wasn't provided
-			return activePublicKey, nil
+			return i.PemPubKey, nil
 		}
 	}
-	if tid == int(l.active) {
-		return activePublicKey, nil
+	if tid == int(l.GetActive().TreeID) {
+		return l.active.PemPubKey, nil
 	}
```
bobcallaway, line 266: non-blocking comment: if you submit another commit here, move this check above the search across inactive shards? no need to search the array if you know you're requesting the active treeID

bobcallaway (approved):

pkg/sharding/ranges.go
Hayden-IO: Will fix in the follow up PR

PR #2332, opened 2025-01-16
Cache checkpoint for inactive shards

Hayden-IO:
#### Summary

On service startup, Rekor will sign checkpoints for the
inactive shards, since inactive tree lengths do not
change.

The calls to CreateAndSignCheckpoint that were not
updated are because the checkpoint is being signed
only for the active shard, e.g. on entry upload
and when returning the active shard checkpoint.

Any deployment using KMS will see a reduction in KMS
calls.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -145,7 +145,7 @@ func NewAPI(treeID uint) (*API, error) {
 		tc := trillianclient.NewTrillianClient(ctx, logClient, r.TreeID)
 		resp := tc.GetLatest(0)
 		if resp.Status != codes.OK {
-			return nil, fmt.Errorf("error with GetLatest(): resp code is %d", resp.Status)
+			return nil, fmt.Errorf("error fetching latest tree head for inactive shard %d: resp code is %d, err is %w", r.TreeID, resp.Status, resp.Err)
 		}
 		result := resp.GetLatestResult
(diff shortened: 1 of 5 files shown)


Hayden-IO:
Pending https://github.com/sigstore/rekor/pull/2330

See the last commit for relevant changes.

pkg/api/api.go
```diff
@@ -132,6 +140,26 @@ func NewAPI(treeID uint) (*API, error) {
 		return nil, fmt.Errorf("unable get sharding details from sharding config: %w", err)
 	}
 
+	cachedCheckpoints := make(map[int64]string)
+	for _, r := range ranges.GetInactive() {
+		tc := trillianclient.NewTrillianClient(ctx, logClient, r.TreeID)
+		resp := tc.GetLatest(0)
+		if resp.Status != codes.OK {
+			return nil, fmt.Errorf("error with GetLatest(): resp code is %d", resp.Status)
```
bobcallaway: maybe make this error message a bit clearer and/or wrap resp.Err?

bobcallaway (dismissed):

pkg/api/api.go
Hayden-IO: done

bobcallaway (approved):

PR #2350, opened 2025-01-29
Switch to `avast/retry-go` for HTTP retry

### LINKED ISSUE #2342: MPL-licensed `github.com/hashicorp/go-retryablehttp` is not allowed in CNCF
Hey folks, we (Kubernetes) use the rekor client directly from go in kubernetes-sigs/release-sdk: https://github.com/kubernetes-sigs/release-sdk/blob/83243fb51416adb402fbbb1e611e2861e8e8d008/sign/sign.go#L726

The issue is that [`GetRekorClient`  imports go-retryablehttp](https://github.com/sigstore/rekor/blob/1cb78ca673d02fb8302b27d6983e69b5281031ea/pkg/client/rekor_client.go#L27), which is MPL licensed and therefore not allowed in the CNCF. 

Would it be generally possible to switch to an alternative?  

Refers to https://github.com/kubernetes-sigs/release-sdk/issues/197

saschagrunert:
#### Summary
The hashicorp dependencies are MPL licensed and therefore not allowed in the CNCF. This means we now use the `avast/retry-go` implementation to achieve the same goal.

#### Release Note
Use  `github.com/avast/retry-go` instead of `github.com/hashicorp/go-retryablehttp` for HTTP request retry.

#### Documentation
None

Fixes https://github.com/sigstore/rekor/issues/2342

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,12 +56,11 @@ require (
 	cloud.google.com/go/pubsub v1.45.3
 	github.com/AdamKorcz/go-fuzz-headers-1 v0.0.0-20230919221257-8b5d3ce2d11d
 	github.com/DATA-DOG/go-sqlmock v1.5.2
+	github.com/avast/retry-go/v4 v4.6.0
 	github.com/cyberphone/json-canonicalization v0.0.0-20220623050100-57a0ce2678a7
 	github.com/go-redis/redismock/v9 v9.2.0
 	github.com/go-sql-driver/mysql v1.8.1
 	github.com/golang/mock v1.6.0
-	github.com/hashicorp/go-cleanhttp v0.5.2
-	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/jmoiron/sqlx v1.4.0
 	github.com/redis/go-redis/v9 v9.7.0
 	github.com/sassoftware/relic/v7 v7.6.2
@@ -132,7 +131,9 @@ require (
 	github.com/google/pprof v0.0.0-20240727154555-813a5fbdbec8 // indirect
 	github.com/google/s2a-go v0.1.9 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
+	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
+	github.com/hashicorp/go-retryablehttp v0.7.7 // indirect
 	github.com/hashicorp/go-rootcerts v1.0.2 // indirect
 	github.com/hashicorp/go-secure-stdlib/parseutil v0.1.7 // indirect
 	github.com/hashicorp/go-secure-stdlib/strutil v0.1.2 // indirect
--- a/pkg/client/options.go
+++ b/pkg/client/options.go
@@ -15,10 +15,16 @@
 package client
 
 import (
+	"crypto/tls"
+	"errors"
+	"fmt"
 	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
 	"time"
 
-	"github.com/hashicorp/go-retryablehttp"
+	"github.com/avast/retry-go/v4"
 )
 
 // Option is a functional option for customizing static signatures.
@@ -30,7 +36,6 @@ type options struct {
 	RetryWaitMin        time.Duration
 	RetryWaitMax        time.Duration
 	InsecureTLS         bool
-	Logger              interface{}
 	NoDisableKeepalives bool
 	Headers             map[string][]string
 }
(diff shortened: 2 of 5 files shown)


pkg/client/options.go
```diff
@@ -113,33 +108,43 @@ func WithHeaders(h map[string][]string) Option {
...
+			return err
+		}
+		if res.StatusCode != 200 {
+			return fmt.Errorf("%d: %s", res.StatusCode, res.Status)
+		}
+		return nil
+	},
+		retry.Attempts(rt.options.RetryCount),
+		retry.Delay(rt.options.RetryWaitMin),
+		retry.MaxDelay(rt.options.RetryWaitMax),
```
saschagrunert, line 135: This may behave different now. Worth double checking if that's intentional.

saschagrunert:
Green CI :)

cpanato (dismissed):

pkg/client/options.go
```diff
@@ -81,14 +86,9 @@ func WithRetryWaitMax(t time.Duration) Option {
...
-// WithLogger sets the logger; it must implement either retryablehttp.Logger or retryablehttp.LeveledLogger; if not, this will not take effect.
-func WithLogger(logger interface{}) Option {
-	return func(o *options) {
-		switch logger.(type) {
-		case retryablehttp.Logger, retryablehttp.LeveledLogger:
-			o.Logger = logger
-		}
-	}
+// WithLogger sets the logger; this method is deprecated and will not take any effect.
+func WithLogger(interface{}) Option {
```
bobcallaway:
```suggestion
func WithLogger(_ interface{}) Option {
```

pkg/client/options.go
```diff
@@ -113,33 +113,75 @@ func WithHeaders(h map[string][]string) Option {
...
 		}
 	}
-	return rt.RoundTripper.RoundTrip(req)
+
+	err = retry.Do(func() (err error) {
+		res, err = rt.inner.RoundTrip(req)
+		if retryErr := shouldRetry(res, err); retryErr != nil {
+			return retryErr
+		}
+		return nil
```
bobcallaway:
```suggestion
		return shouldRetry(res, err)
```

pkg/client/options.go
```diff
@@ -113,33 +113,75 @@ func WithHeaders(h map[string][]string) Option {
...
+		retry.MaxDelay(rt.options.RetryWaitMax),
+	)
+
+	return res, err
 }
 
-func createRoundTripper(inner http.RoundTripper, o *options) http.RoundTripper {
+var tooManyRedirectyRe = regexp.MustCompile(`stopped after \d+ redirects\z`)
+
+func shouldRetry(resp *http.Response, err error) error {
```
Hayden-IO, line 144: Is this a change in behavior or is this copied from go-retryable's implementation?

pkg/client/rekor_client.go
```diff
@@ -36,25 +36,26 @@ func GetRekorClient(rekorServerURL string, opts ...Option) (*client.Rekor, error
...
+			DualStack: true,
+		}).DialContext,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConnsPerHost:   -1,
+		DisableKeepAlives:     !o.NoDisableKeepalives,
+		TLSClientConfig:       &tls.Config{InsecureSkipVerify: o.InsecureTLS}, //nolint:gosec
```
Hayden-IO: Previously this was only set if `o.InsecureTLS`. Should we maintain that, or is there no change in behavior? I'm not sure if there's a default tls config we're overriding.

pkg/client/options.go
saschagrunert: It's not copied over, but it should reflect the same retry behavior.

pkg/client/rekor_client.go
saschagrunert: Changed it to only apply `if o.InsecureTLS`, but I assume it's the same as before in the end.

go.mod
```diff
@@ -132,7 +131,9 @@ require (
 	github.com/google/pprof v0.0.0-20240727154555-813a5fbdbec8 // indirect
 	github.com/google/s2a-go v0.1.9 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
+	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
```
bobcallaway, line 134: looks like these are still indirects (probably because of us supporting hashivault)... is that still an issue for CNCF?

saschagrunert: I think so, but the code does not seem to be imported from kubernetes-sigs/release-sdk, we may follow-up on that, though.

Hayden-IO:
If you'd like to still get this in, I'd suggest a different change - let's just implement a retrying client using the std library. If we can drop a dep and not pick up a new one at the same time, that'd be nice.

Hayden-IO:
Well, I ended up just trying to do this - https://github.com/sigstore/rekor/pull/2933

Lemme know if this would work?

PR #2351, opened 2025-01-30
Fix 'Using config file:' log message

saschagrunert:
#### Summary
Without applying this patch, the log message will be:
```
Using config file:%!(EXTRA string=/tmp/rekor_test.q1FmiY.rekor.yaml)
```

#### Release Note
Fixed "Using config file:" log message format.

#### Documentation
None


cpanato (approved):
thanks

PR #2360, opened 2025-02-08
use go1.23.6 to build rekor 

cpanato:


bobcallaway:
should we also update the Dockerfile?

cpanato:
> should we also update the Dockerfile?

Done, forgot that one

thanks!

PTAL

bobcallaway (approved):

PR #2429, opened 2025-04-04
pkg/api: better logs when algorithm registry rejects a key

ret2libc:
Right now the logs are quite obscure when the registry rejects a key. This change provides more info about the rejected algorithm, for better observability.

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -320,7 +320,7 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 			publicKeyInfo = fmt.Sprintf("public key: %T", pKey)
 		}
 		algorithmInfo := fmt.Sprintf("entry algorithm %v/%v not allowed", publicKeyInfo, checkedAlgorithmResult.hash)
-		return nil, handleRekorAPIError(params, http.StatusBadRequest, errors.New("entry algorithms are not allowed"), fmt.Sprintf(validationError, algorithmInfo))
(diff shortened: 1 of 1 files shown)


pkg/api/entries.go
```diff
@@ -287,12 +294,33 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
...
+			}
+		case *rsa.PublicKey:
+			publicKeyInfo = fmt.Sprintf("rsa public key: %T/%v", pKey, pKey.Size())
+		case *ed25519.PublicKey:
+			publicKeyInfo = fmt.Sprintf("ed25519 public key: %T", pKey)
+		default:
+			publicKeyInfo = fmt.Sprintf("public key: %T", pKey)
+		}
+		algorithmInfo := fmt.Sprintf("entry algorithm %v/%v not allowed", publicKeyInfo, checkedAlgorithmResult.hash)
+		return nil, handleRekorAPIError(params, http.StatusBadRequest, errors.New("entry algorithms are not allowed"), fmt.Sprintf(validationError, algorithmInfo))
```
Hayden-IO: The last two params should be flipped, the error logged is the second to last param, the error message for the client is the last param.

If you want a more descriptive error message, we'll need to update [the e2e test](https://github.com/sigstore/rekor/blob/30c2d4e11c480402d93258b463d2ae03a2029ad1/tests/client-algos-e2e-test.sh#L171) too

Hayden-IO (approved):

PR #2438, opened 2025-04-11
emit unpopulated values when marshalling TLE

bobcallaway:
if log index is 0 (or there are no hashes in an inclusion proof - which is possible only for the first entry in a log), we're not marshalling values to disk. The default values are implicit but I think being explicit here is better than omitting the value completely

### DIFF
--- a/pkg/tle/tle_test.go
+++ b/pkg/tle/tle_test.go
@@ -19,6 +19,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/go-openapi/strfmt"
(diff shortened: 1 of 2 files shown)


pkg/tle/tle_test.go
```diff
@@ -223,3 +224,90 @@ func TestGenerateTransparencyLogEntry(t *testing.T) {
...
+				IntegratedTime: swag.Int64(123),
+				LogID:          swag.String("deadbeef"),
+				LogIndex:       swag.Int64(0),
+				Verification: &models.LogEntryAnonVerification{
+					InclusionProof: &models.InclusionProof{
+						Checkpoint: swag.String("checkpoint"),
+						Hashes:     []string{},
+						LogIndex:   swag.Int64(0),
+						RootHash:   swag.String("abcdefaa"),
```
Hayden-IO, line 256: I'm trying to find the commit now, but are we handling an empty log root hash by providing the default sha256([]byte{})?

Hayden-IO (approved):

pkg/tle/tle_test.go
Hayden-IO: Ah we return https://github.com/sigstore/rekor/pull/1311/, that's fine

PR #2459, opened 2025-04-22
better mysql healthcheck

ramonpetgrave64:
#### Summary

Improves the healthcheck for mysql in the docker-compose.yml file.

See https://github.com/docker-library/mysql/issues/930.

With the previous healthcheck, the mysql service may be running, but the actual database may not be ready yet.

```shell
 ➜  rekor git:(main) docker compose up --wait
[+] Running 5/6
 ✔ Network rekor_default                  Created          0.1s 
 ✔ Container rekor-redis-server-1         Healthy         94.4s 
 ✔ Container rekor-mysql-1                Healthy         94.4s 
 ✔ Container rekor-trillian-log-signer-1  Healthy         94.6s 
 ✔ Container rekor-trillian-log-server-1  Healthy         94.3s 
 ⠦ Container rekor-rekor-server-1         Waiting         96.1s 
container rekor-rekor-server-1 is unhealthy
```

With the newer healthcheck mysql takes longer to be healthy, but `docker compose up --wait` succeeds.

```shell
➜  rekor git:(main) ✗ docker compose up --wait
 ✔ Network rekor_default                  Created          0.2s 
[+] Running 5/6kor-redis-server-1         Started          1.4s 
[+] Running 6/6r_default                  Created          0.2s 
 ✔ Network rekor_default                  Created          0.2s 
 ✔ Container rekor-redis-server-1         Healthy        128.3s 
 ✔ Container rekor-mysql-1                Healthy        128.4s 
 ✔ Container rekor-trillian-log-server-1  Healthy        128.2s 
 ✔ Container rekor-trillian-log-signer-1  Healthy        128.2s 
 ✔ Container rekor-rekor-server-1         Healthy        133.5s
```

**TODO**

- [ ] add this same healthceck to fulcio's.

#### Release Note

Improve docker-compose.yml healthcheck for mysql.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,7 +24,8 @@ services:
       - MYSQL_PASSWORD=zaphod
     restart: always # keep the MySQL server running
     healthcheck:
-      test: ["CMD", "/etc/init.d/mysql", "status"]
+      # better healthcheck for mysql. See https://github.com/docker-library/mysql/issues/930.
+      test: ["CMD", "mysqladmin", "-h", "localhost", "-u$MYSQL_USER",  "-p$MYSQL_ROOT_PASSWORD", "-s", "ping"]
       interval: 30s
       timeout: 3s
       retries: 5
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,9 +26,9 @@ services:
     healthcheck:
       # better healthcheck for mysql. See https://github.com/docker-library/mysql/issues/930.
       test: ["CMD", "mysqladmin", "-h", "localhost", "-u$MYSQL_USER",  "-p$MYSQL_ROOT_PASSWORD", "-s", "ping"]
-      interval: 30s
+      interval: 10s
       timeout: 3s
-      retries: 5
+      retries: 15
       start_period: 90s
   redis-server:
     image: docker.io/redis:6.2


docker-compose.yml
```diff
@@ -24,7 +24,8 @@ services:
       - MYSQL_PASSWORD=zaphod
     restart: always # keep the MySQL server running
     healthcheck:
-      test: ["CMD", "/etc/init.d/mysql", "status"]
+      # better healthcheck for mysql. See https://github.com/docker-library/mysql/issues/930.
+      test: ["CMD", "mysqladmin", "-h", "localhost", "-u$MYSQL_USER",  "-p$MYSQL_ROOT_PASSWORD", "-s", "ping"]
       interval: 30s
```
Hayden-IO: Can we change the interval to 10s and increase retries?

Hayden-IO:
These startup times are getting really high, this shouldn't take 2 minutes to startup. Do we need to be lowering intervals to be more frequent?

docker-compose.yml
ramonpetgrave64: changed.

Hayden-IO:
Running on macOS, `docker compose up --wait` takes about ~15s so I'm not sure why you're seeing those numbers. 

However I did see an issue - when `rekor-server` first starts up it reports unhealthy because the log server is not yet up so it panics. The server recovers and restarts, and eventually becomes healthy.

ramonpetgrave64:
Yes, it's slower on my linux machine, maybe because of my partition type.

 - https://forums.docker.com/t/mysql-slow-performance-in-docker/37179/9

Hayden-IO (approved):
I'll merge this, but I think this doesn't work unless there's artificial delays on starting up the log server.

PR #2473, opened 2025-05-06
Fix docker compose up --wait failing when Trillian server isn't healthy

Hayden-IO:
As noted in https://github.com/docker/compose/issues/12424, compose --wait doesn't seem to honor healthchecks with restart:always, when the server crashes and restarts a few times and eventually becomes healthy. This was happening with Rekor:

* MySQL was not yet healthy because the healthcheck wasn't working as expected. https://github.com/docker-library/mysql/issues/930#issuecomment-1463541157 suggested using 127.0.0.1 instead of localhost
* trillian-log-server was not yet healthy even when MySQL reported as healthy, causing trillian-log-server to crash and restart a few times. There was no healthcheck for either Trillian service because the image we're using is based on Distroless, which has no curl/wget.
* rekor-server tried to start up with an unhealthy trillian-log-server, and crashed. The healthcheck reported as unhealthy, and even though the server eventually became healthy because of the restart:always policy, the healthcheck reported the startup as unhealthy.

This change adds healthchecks to trillian-log-server and log-signer by pulling the binaries out of the images and putting them into Debian 12 containers that include curl, so we can curl the /healthz endpoint. This also fixes the MySQL healthcheck as noted above. Now, docker compose up --wait properly waits for a healthy MySQL before starting trillian-log-server, and a healthy Trillian before starting Rekor.

Also fix minor Dockerfile linting errors.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,9 +24,9 @@ services:
       - MYSQL_PASSWORD=zaphod
     restart: always # keep the MySQL server running
     healthcheck:
-      # better healthcheck for mysql. See https://github.com/docker-library/mysql/issues/930.
-      test: ["CMD", "mysqladmin", "-h", "localhost", "-u$MYSQL_USER",  "-p$MYSQL_ROOT_PASSWORD", "-s", "ping"]
-      interval: 10s
+      # better healthcheck for MySQL. See https://github.com/docker-library/mysql/issues/930.
+      test: "mysqladmin -h 127.0.0.1 --user=$$MYSQL_USER --password=$$MYSQL_ROOT_PASSWORD -s ping"
+      interval: 5s
       timeout: 3s
       retries: 15
       start_period: 90s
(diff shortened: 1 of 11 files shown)


cpanato (dismissed):

docker-compose.yml
```diff
@@ -24,9 +24,9 @@ services:
       - MYSQL_PASSWORD=zaphod
     restart: always # keep the MySQL server running
     healthcheck:
-      # better healthcheck for mysql. See https://github.com/docker-library/mysql/issues/930.
-      test: ["CMD", "mysqladmin", "-h", "localhost", "-u$MYSQL_USER",  "-p$MYSQL_ROOT_PASSWORD", "-s", "ping"]
-      interval: 10s
+      # better healthcheck for MySQL. See https://github.com/docker-library/mysql/issues/930.
+      test: ["CMD", "mysqladmin", "-h", "127.0.0.1", "-u$MYSQL_USER",  "-p$MYSQL_ROOT_PASSWORD", "-s", "ping"]
```
bobcallaway: this healthcheck is not working as you'd expect: see https://github.com/sigstore/rekor/actions/runs/14851848568/job/41697670054?pr=2473#step:8:824

```sh
WARN[0000] The "MYSQL_USER" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MYSQL_ROOT_PASSWORD" variable is not set. Defaulting to a blank string. 
```

I'd change this to: `test: "mysqladmin -h 127.0.0.1 --user=$$MYSQL_USER --password=$$MYSQL_ROOT_PASSWORD -s ping"` which will correctly interpolate the value of the environment variables

Hayden-IO: Thanks, this resolved that warning. Updated here and in the backfill compose script.

cpanato (dismissed):

cpanato (approved):

PR #2535, opened 2025-07-14
backoff pubsub emulator to last-known good

bobcallaway:
there's a bug in the pubsub emulator due to a dependency mismatch; the fix probably won't be out until next week so backing this off to unblock other PRs


cpanato (dismissed):

Hayden-IO:
LGTM, but there’s some issue with the harness test suite, I’m looking into it. I added dockerfiles for the trillian services recently, but I’m not immediately seeing the issue.

Hayden-IO (approved):

PR #2548, opened 2025-07-29
enable retries and timeouts on GCP KMS calls

bobcallaway:
the current default for the GCP KMS client is to never retry; this allows us to configure this option for better resiliency for intermittent failures

this required bumping to go 1.24 as well


Hayden-IO (commented):
LGTM, just need to fix lint.

Hayden-IO (dismissed):

Hayden-IO (approved):

PR #2564, opened 2025-08-07
move to per-shard trillian client manager

bobcallaway:
this allows users to configure a different trillian gRPC backend per-shard instead of assuming all shards are hosted by the same trillian instance. It is backwards compatible (if only the base level values for `trillian-log-server.{address, port, tls, tls_ca_cert}` are specified, they are used for all shards unless overridden in the shard config).

include some other minor fixes:

- initialize using int64 to avoid redundant casting
- ensure consistency proof endpoint returns 400 for unknown shard


pkg/api/api.go
```diff
@@ -123,28 +68,54 @@ var AllowedClientSigningAlgorithms = []v1.PublicKeyDetails{
...
+		TinkKeysetPath:         viper.GetString("rekor_server.tink_keyset_path"),
+		GCPKMSRetries:          viper.GetUint("rekor_server.signer.gcpkms.retries"),
+		GCPKMSTimeout:          viper.GetUint("rekor_server.signer.gcpkms.timeout"),
+	}
+	ranges, err := sharding.NewLogRanges(ctx, shardingConfig, treeID, signingConfig)
+	if err != nil {
+		return nil, fmt.Errorf("unable get sharding details from sharding config: %w", err)
+	}
+
+	grpcConfigs := make(map[int64]trillianclient.GRPCConfig)
```
Hayden-IO: nit, can this be named to more clearly signify it's only for inactive shards?

Alternatively, can the defaultGRPCConfig go into this map as well? I'm not sure if that makes the code downstream much cleaner though.

pkg/api/api.go
```diff
@@ -123,28 +68,54 @@ var AllowedClientSigningAlgorithms = []v1.PublicKeyDetails{
...
+	for _, r := range ranges.GetInactive() {
+		if r.GRPCConfig != nil {
+			grpcConfigs[r.TreeID] = *r.GRPCConfig
+		}
+	}
+	tcm := trillianclient.NewClientManager(grpcConfigs, defaultGRPCConfig, viper.GetString("trillian_log_server.grpc_default_service_config"))
```
Hayden-IO: Should `grpc_default_service_config` be a part of a trillianclient.config struct?

pkg/trillianclient/manager.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, fmt.Errorf("failed to connect to RPC server: %w", err)
+	}
+
+	return conn, nil
+}
+
+// Close stops clients and closes underlying gRPC connections.
+func (cm *ClientManager) Close() error {
+	var err error
+
```
Hayden-IO, line 202: Should you lock `clientMu` as well, in case there's any threads trying to access a client?

Hayden-IO (dismissed):
Looks good overall! As long as we thoroughly test this against both active and inactive shards, seems good.

pkg/api/api.go
bobcallaway: I could make this similar to address/port/etc, thanks for the suggestion

pkg/api/api.go
bobcallaway: i did that initially with treeID = -1 but it felt too dirty; happy to rename it though

Hayden-IO (approved):

PR #2580, opened 2025-08-12
AddLeaf: use cheaper gRPC endpoint since we already have the inclusion proof

bobcallaway:
`AddLeaf` queues the proposed entry for inclusion into the Trillian log, and then polls to see that an inclusion proof exists for that hash. Finally, we need the integration timestamp to pass back along to the client as a timestamp, and the only way to get that value is to fetch the entire entry again. We currently do that by calling `GetEntryAndProof`, which results in a Trillian having to recompute the inclusion proof (which it's already given us) which is more expensive than just reading the leaf itself using `GetLeavesByRange`. 

This swaps over to the less-complex `GetLeavesByRange` for a lower overhead implementation with equivalent function.

### DIFF
--- a/pkg/trillianclient/trillian_client.go
+++ b/pkg/trillianclient/trillian_client.go
@@ -16,6 +16,7 @@
 package trillianclient
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"fmt"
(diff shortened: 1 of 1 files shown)


Hayden-IO (approved):
I assume we have sufficient test coverage in the e2e tests?

bobcallaway:
> I assume we have sufficient test coverage in the e2e tests?

yup, this would be covered by all of our existing e2e write tests.

PR #2581, opened 2025-08-13
use less expensive gRPC call to implement GetLeafAndProofByHash

bobcallaway:
Similar to #2580, `GetLeafAndProofByHash` needs to call two Trillian endpoints:

- `GetInclusionProofByHash` to get the index
- another method to get the leaf content

We were previously calling `GetEntryAndProof` via `GetLeafAndProofByIndex` but since we already have the proof from `GetInclusionProofByHash`, this is a cheaper sequence of calls for Trillian with equivalent functionality.

### DIFF
--- a/pkg/trillianclient/trillian_client.go
+++ b/pkg/trillianclient/trillian_client.go
@@ -168,7 +168,7 @@ func (t *TrillianClient) AddLeaf(ctx context.Context, byteValue []byte) *Respons
 
 	leafIndex := proofs[0].LeafIndex
 	// fetch the leaf without re-requesting a proof (since we already have it)
-	leafOnlyResp := t.GetLeafWithoutProof(ctx, leafIndex)
+	leafOnlyResp := t.getStandaloneLeaf(ctx, leafIndex, resp.QueuedLeaf.Leaf.MerkleLeafHash, proofs[0], proofResp.getProofResult.SignedLogRoot)
 	if leafOnlyResp.Err != nil {
 		return &Response{
 			Status:       status.Code(leafOnlyResp.Err),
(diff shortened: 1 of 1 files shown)


pkg/trillianclient/trillian_client.go
```diff
@@ -229,7 +229,42 @@ func (t *TrillianClient) GetLeafAndProofByHash(ctx context.Context, hash []byte)
...
+	}
+
+	if leafOnlyResp.GetLeavesByRangeResult == nil || len(leafOnlyResp.GetLeavesByRangeResult.Leaves) == 0 {
+		err := fmt.Errorf("no leaf returned for index %d", leafIndex)
+		return &Response{
+			Status: codes.NotFound,
+			Err:    err,
+		}
+	}
+	leaf := leafOnlyResp.GetLeavesByRangeResult.Leaves[0]
```
Hayden-IO: Should we be checking that only a single leaf is returned?

pkg/trillianclient/trillian_client.go
```diff
@@ -229,7 +229,42 @@ func (t *TrillianClient) GetLeafAndProofByHash(ctx context.Context, hash []byte)
...
+	// fetch the leaf without re-requesting a proof (since we already have it)
+	leafOnlyResp := t.GetLeafWithoutProof(ctx, leafIndex)
+	if leafOnlyResp.Err != nil {
+		return &Response{
+			Status: status.Code(leafOnlyResp.Err),
+			Err:    leafOnlyResp.Err,
+		}
+	}
+
+	if leafOnlyResp.GetLeavesByRangeResult == nil || len(leafOnlyResp.GetLeavesByRangeResult.Leaves) == 0 {
```
Hayden-IO: Can we pull 242 to 258 into its own method to reuse it in AddLeaf?

Hayden-IO (dismissed):

pkg/trillianclient/trillian_client.go
bobcallaway: ah yes, good call.

Hayden-IO (approved):

PR #2583, opened 2025-08-14
Add STH caching with configurable timeouts and intelligent retry

bobcallaway:
This PR adds an optional cached Signed Tree Head (STH) path to the Trillian client to reduce repeated root-fetch overhead on hot paths while preserving correctness guarantees.

When `--trillian_log_server.cache_sth=True`, the client maintains a background root updater, stores the latest verified root for fast concurrent reads, and wakes waiting callers when the tree advances. The non-cached behavior remains available as the default path.

## What changed

- Added a cache-enabled Trillian client implementation with background root updates.
- Added a simple stateless client implementation and interface-based selection.
- Wired server/config plumbing for the cache_sth feature flag.
- Updated manager lifecycle/close behavior for safe client reuse and shutdown.
- Added focused tests for updater retries, manager caching/close semantics, and simple-client behavior.
- Added load-test scaffolding for evaluating throughput/latency impact.

## Why
The previous flow repeatedly fetched/validated roots during high concurrency, creating avoidable latency and contention. Caching the latest verified root allows callers to read quickly while still reacting to tree advancement.

## Impact

- Lower per-request overhead on read/write paths under load.
- Better scaling behavior when many callers contend on the same tree.
- Performance improvements are observable in local/load-test runs, but exact gains depend on deployment topology and workload mix.

## Safety / rollout

- Feature is gated behind `trillian_log_server.cache_sth` (experimental) command line argument.
- Existing behavior is preserved when the flag is disabled.
- Includes test coverage for concurrency and transient update failures.


pkg/trillianclient/trillian_client.go
```diff
@@ -75,18 +185,170 @@ func unmarshalLogRoot(logRoot []byte) (types.LogRootV1, error) {
...
+			metricUpdaterErrors.WithLabelValues(fmt.Sprintf("%d", t.logID)).Inc()
+			// Reset backoff on persistent failure and continue to next iteration
+			bo.Reset()
+			continue
+		}
+
+		// Success - reset backoff for next potential failure
+		bo.Reset()
+
+		if nr == nil {
```
Hayden-IO, line 235: How can `nr` be nil here? There should only be one thread running `updater()`, correct?

pkg/trillianclient/trillian_client.go
```diff
@@ -75,18 +185,170 @@ func unmarshalLogRoot(logRoot []byte) (types.LogRootV1, error) {
...
+// per advancement.
+func (t *TrillianClient) updater() {
+	// Create backoff for retry logic with reasonable defaults
+	bo := backoff.Backoff{
+		Min:    100 * time.Millisecond, // Start with 100ms
+		Max:    30 * time.Second,       // Cap at 30s
+		Factor: 2.0,                    // Double each time
+		Jitter: true,                   // Add randomization
+	}
+	for {
```
Hayden-IO, line 205: What's the period that this runs? Is this constant, or did you want a delay between each iteration?

pkg/trillianclient/trillian_client.go
```diff
@@ -75,18 +185,170 @@ func unmarshalLogRoot(logRoot []byte) (types.LogRootV1, error) {
...
+		lrBytes, mErr := nr.MarshalBinary()
+		if mErr != nil {
+			log.Logger.Debugw("failed to marshal updated log root", "treeID", t.logID, "err", mErr)
+			metricUpdaterErrors.WithLabelValues(fmt.Sprintf("%d", t.logID)).Inc()
+			continue
+		}
+		slr := &trillian.SignedLogRoot{LogRoot: lrBytes}
+
+		// publish new snapshot and notify waiters
+		t.mu.Lock()
```
Hayden-IO, line 251: Why the lock here? The snapshot update is atomic.

pkg/trillianclient/trillian_client.go
```diff
@@ -75,18 +185,170 @@ func unmarshalLogRoot(logRoot []byte) (types.LogRootV1, error) {
...
+		Factor: 2.0,                    // Double each time
+		Jitter: true,                   // Add randomization
+	}
+	for {
+		// Wrap the WaitForRootUpdate call with backoff retry
+		var nr *types.LogRootV1
+		err := bo.Retry(t.bgCtx, func() error {
+			select {
+			case <-t.stopCh:
+				return fmt.Errorf("client stopped")
```
Hayden-IO: Should this be returning nil rather than an error, since this is expected on server shutdown?

pkg/trillianclient/trillian_client.go
```diff
@@ -75,18 +185,170 @@ func unmarshalLogRoot(logRoot []byte) (types.LogRootV1, error) {
...
-		LogId: t.logID,
-	}
-	resp, err := t.client.GetLatestSignedLogRoot(ctx, rqst)
+// ensureStarted initializes the shared LogClient and starts the updater once.
+//
+// Locking strategy: this uses a double-checked initialization gate so network
+// calls do not hold the mutex. Only one goroutine performs the initial RPCs,
+// others wait on the condition variable until initialization completes. This
+// avoids head-of-line blocking while keeping state updates atomic.
+func (t *TrillianClient) ensureStarted(ctx context.Context) error {
```
Hayden-IO, line 148: Could you have this called when initializing the client, rather than from each client function?

Other option for handling concurrency, have you looked into `sync.OnceFunc` to ensure this runs only once?

pkg/trillianclient/trillian_client.go
```diff
@@ -220,137 +450,200 @@ func (t *TrillianClient) GetLeafAndProofByHash(ctx context.Context, hash []byte)
...
+		if err := t.waitForRootAtLeast(ctx, root.TreeSize+1); err != nil {
+			return &Response{Status: status.Code(err), Err: err}
+		}
+	}
+}
+
+// waitForRootAtLeast blocks until t.lastRoot.TreeSize >= size, or context/client closes.
+func (t *TrillianClient) waitForRootAtLeast(ctx context.Context, size uint64) error {
+	start := time.Now()
+	t.mu.Lock()
```
Hayden-IO, line 563: Can this read a reader lock?

pkg/trillianclient/trillian_client.go
```diff
@@ -220,137 +450,200 @@ func (t *TrillianClient) GetLeafAndProofByHash(ctx context.Context, hash []byte)
...
+			return status.Error(codes.Canceled, "client closed")
+		default:
+		}
+		cur := t.snapshot.Load().(rootSnapshot)
+		if cur.root.TreeSize >= size {
+			elapsed := float64(time.Since(start).Milliseconds())
+			metricWaitForRootAtLeast.WithLabelValues(fmt.Sprintf("%d", t.logID), "true").Observe(elapsed)
+			return nil
+		}
+		t.cond.Wait()
```
Hayden-IO, line 575: What happens when an error occurs in `updater()`? Let's say Trillian is down. From what I can tell, a request blocks until `Broadcast` is called, but `Broadcast` is only called on successful update of the tree head.

Should an error in `updater()` trigger a Broadcast so waiters unblock, and then observe an error in the context and return? Or can we have a timeout on waiting?

pkg/trillianclient/trillian_client.go
bobcallaway: the return value here stops the Retry loop from continuing, which is what we'd want in a shutdown scenario.

pkg/trillianclient/trillian_client.go
bobcallaway: `WaitForRootUpdate` has it's own backoff mechanism, which exponentially backs of if it doesn't see a root update.

pkg/trillianclient/trillian_client.go
bobcallaway: based on the current implementation of `WaitForRootUpdate`, this can never happen. but in case that were to change, if we don't have a new root then there's nothing to update.

pkg/trillianclient/trillian_client.go
bobcallaway: yes, the snapshot update is atomic but the locking here forces a serialized execution between the check in `waitForRootAtLeast` and the update in `updater()`. 

if we didn't hold the lock here, it's possible that the goroutine scheduler could fire the `t.cond.Broadcast()` after the size comparison but strictly before `waitForRootAtLeast` had called `t.cond.Wait()` which would mean that the reader wouldn't see the broadcast notification that the root had changed.

pkg/trillianclient/trillian_client.go
bobcallaway: nope, this requires a full lock given semantics of `sync.Cond`.

pkg/trillianclient/trillian_client.go
bobcallaway: I could do it there, but if there is a transient outage putting it inside of a `sync.OnceFunc` means we should really panic the entire process since we could never re-initialize it.

PR #2605, opened 2025-09-05
use pubsub client to check IAM permissions

bobcallaway:


Hayden-IO (approved):
FWIW, we could also consider disabling this feature. It’s not used by anyone in prod.

PR #2667, opened 2025-11-04
log: remove zap & go-chi dependecy from pkg/types

tonistiigi:
#### Summary

Rekor currently depends on Uber’s Zap logging framework (with go-chi middleware integration). It’s unclear to me why this particular stack was introduced, as other Sigstore projects don’t seem to use it.

However, this dependency leaks into library packages such as those under `pkg/types`, which are meant to be reused by other projects like `sigstore-go` when running TLog verification. This adds an unnecessary heavy dependency and also causes out-of-place log output in importing applications.

This PR fixes that by adding a lightweight, dependency-free internal logger (`pkg/internal/log`) as the default for these library packages. The internal logger is silent by default, which is suitable for libraries. Any user of Zap or the Rekor CLI utilities can continue to configure logging via `pkg/log` without any code or behavior changes(Rekor CLI still gets all the same logs as before). If an importing application wants to capture logs from Rekor libs, it can configure the `pkg/log` package as before.

Dependency footprint change for `sigstore-go/verifier`:
```
112 files changed, 34 insertions(+), 14361 deletions(-)
```

### DIFF
--- a/pkg/internal/log/logger.go
+++ b/pkg/internal/log/logger.go
@@ -1,0 +1,36 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


Hayden-IO (dismissed):
Thanks

Hayden-IO:
Can you add the license header to the new file?

tonistiigi:
> Can you add the license header to the new file?

Done

Hayden-IO (approved):

PR #2668, opened 2025-11-04
pkg: separate pki types from implementations

tonistiigi:
#### Summary

`pkg/pki` package defines both the interface types for `PublicKey` and `Signature`, linked to many external packages, and also all the implementations for pki via a static factory map.

This separates the types to separate packages so the packages that use them can be included without a big dependency chain. The types are aliased to the old pkg/pki package so that this change wouldn't break any backwards compatibility.

This could be cleaned up more in a further refactor when backwards compatibility is not an issue.

Footprint change for `sigstore-go/verifier` 
```
 80 files changed, 18 insertions(+), 13144 deletions(-)
```
Most notably, this drops the inclusion of `spf13/cobra` CLI framework and its dependencies that somehow is pulled in via x509tools dependency 🤯

### DIFF
--- a/pkg/pki/pkitypes/types.go
+++ b/pkg/pki/pkitypes/types.go
@@ -16,11 +16,25 @@
 package pki
 
 import (
-	pkitypes "github.com/sigstore/rekor/pkg/pki/pkitypes"
+	"io"
+
+	"github.com/sigstore/rekor/pkg/pki/identity"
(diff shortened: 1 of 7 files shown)


Hayden-IO (dismissed):
Thank you! And thanks for making this backwards compatible as well

Hayden-IO:
Just need a rebase

Hayden-IO (approved):

PR #2669, opened 2025-11-04
pkg: remove viper config from spec definitions

tonistiigi:
#### Summary

CLI config reading routines from Viper should not be included in spec definition types under pkg because
    - this creates unnecessarily huge dependency
    - it creates an unexpected reconfiguration/attack method to applications importing the types.

Instead, read viper config in Rekor CLI and pass to the types packages to reconfigure them. The default size limit remains unchanged.

Optionally, I could instead create a `limits` pkg with no dependencies that spec types import for shared implementation. In that case, let me know where you want that pkg defined.

Footprint reduction to `sigstore-go/verifier` from this change:
```
 218 files changed, 33 insertions(+), 35989 deletions(-)
```

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -27,6 +27,9 @@ import (
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/sigstore/rekor/pkg/api"
 	"github.com/sigstore/rekor/pkg/log"
+	cose "github.com/sigstore/rekor/pkg/types/cose/v0.0.1"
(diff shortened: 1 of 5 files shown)


Hayden-IO (dismissed):
I'm good with this approach, I don't think we need a shared limits package. Thanks!

Hayden-IO:
Just need to fix the test.

tonistiigi:
@haydentherapper I fixed the test, but I'm a bit confused now. Was this expected to fail by default when packages were used without the flag definition in `rekor-server/app`? That's what the test seemed to assume.

Hayden-IO (approved):

PR #2671, opened 2025-11-04
chore: remove double imported mapstructure pkg

tonistiigi:
#### Summary

Removed imports of `mitchellh/mapstructure`. The package is already imported from `go-viper` after the owner change.


Hayden-IO (approved):
Thanks!

PR #2672, opened 2025-11-05
e2e: don't mix e2e and regular utilities

tonistiigi:
#### Summary

Separate utilities and test data made exclusively for E2E testing from other packages.

E2E utilities should be only used in test files and should not be defined mixed with the common functions in pkg/util. Go modules/vendoring makes no filtering based on build tags and so all the dependencies needed for E2E become dependencies to any package that internally imports pkg/util.


Hayden-IO (commented):
Thank you again for cleaning all of this up!

Hayden-IO (approved):

PR #2755, opened 2026-02-17
Type assert the entry bundle when verifying inclusion proof

Hayden-IO:
This guards against a body unexpectedly being a different type.

Thanks to @1seal for reporting this.

### DIFF
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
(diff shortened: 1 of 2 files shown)


pkg/verify/verify.go
```diff
@@ -158,7 +156,11 @@ func VerifyInclusion(ctx context.Context, e *models.LogEntryAnon) error {
 	}
 
 	// Verify the inclusion proof.
-	entryBytes, err := base64.StdEncoding.DecodeString(e.Body.(string))
+	b, ok := e.Body.(string)
+	if !ok {
+		return fmt.Errorf("entry body must be a string, was %s", reflect.TypeOf(e.Body))
```
bobcallaway: Can't you just do `%T` and not import `reflect`?

Hayden-IO: yep, changed

bobcallaway (approved):

PR #2762, opened 2026-02-23
feat: add checkpointing into the backfill so we can resume from the l…

### LINKED ISSUE #2279: Avoid reprocessing data in backfill-index.

knrc:
…ast completed, contiguous index

Closes #2279 

#### Summary

This pull request adds support for checkpointing the progress of the backfill process, allowing the process to resume from the last known, contiguous index.  This addresses the issue of the backfill reprocessing information from the beginning each time it runs.

#### Release Note

Added support for checkpointing progress during the backfill operation, rerunning the backfill will resume from the last known, contiguous index instead of starting at the beginning.

### DIFF
--- a/cmd/backfill-index/main.go
+++ b/cmd/backfill-index/main.go
@@ -19,10 +19,18 @@
 	we need to backfill missing entries into the database for the search API.
 	It can also be used to populate an index storage backend from scratch.
 
+	The tool supports automatic checkpointing for both Redis and MySQL backends to allow
+	resuming interrupted backfills and to maintain progress across scheduled runs.
+	Use --checkpoint-interval to control how frequently checkpoints are saved
+	(e.g., --checkpoint-interval 100 saves every 100 entries). On subsequent runs,
+	the tool automatically continues from the last checkpoint unless --reset-checkpoint is set.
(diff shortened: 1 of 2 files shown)


Hayden-IO:
@knrc Could we simply write a log index periodically to disk to simplify this? Backfilling is idempotent, so we don't need to be overly precise with the latest index.

knrc:
> @knrc Could we simply write a log index periodically to disk to simplify this? Backfilling is idempotent, so we don't need to be overly precise with the latest index.

Yes, that's possible but I felt it would be better to keep the checkpoint within the same storage so they are managed together.  The code is already taking advantage of the idempotency when tracking/updating the checkpoint, and does expect to replay on out of sequence updates or errors.

Hayden-IO:
Given the amount of code needed for this, I'd strongly prefer a much simpler solution, especially given the shift to Rekor v2.

knrc:
Closing, I believe this codebase is moving more to a maintenance only phase.

PR #2766, opened 2026-03-04
optimize memory for DSSE v0.0.1 processing

Hayden-IO:
#### Summary

This change reduces the memory footprint of DSSE v0.0.1 entries by:
- Extracting index keys (subjects and SLSA materials) during initial unmarshaling
- Clearing the large base64-encoded envelope and decoded payload early
- Avoiding redundant payload decoding and JSON unmarshaling in IndexKeys

These optimizations prevent out-of-memory errors when processing large attestations.


bobcallaway (approved):
i attempted to do this via streaming, `io.MultiReader`, etc - and this kept coming back as the optimal approach under benchmarking

PR #2775, opened 2026-03-13
Handle client connection issues better

jku:
This is a potential fix for some observability issues:
* we return and log generic 500 errors for GRPC issues where we can reasonably use a better HTTP code (without bloating every callsite with grpc error handling)
* Internally we log ERRORs for client connection issues: That seems unnecessary

These are separate issues, I can file separate PRs.

The recoverer fix is mostly from AI but makes sense to me and should handle the `panic detected: write tcp 10.1.4.25:3000->35.191.75.48:57722: i/o timeout` error that started this

### DIFF
--- a/pkg/api/error.go
+++ b/pkg/api/error.go
@@ -16,12 +16,15 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"regexp"
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/go-openapi/strfmt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/generated/restapi/operations/entries"
@@ -30,6 +33,29 @@ import (
 	"github.com/sigstore/rekor/pkg/generated/restapi/operations/tlog"
 	"github.com/sigstore/rekor/pkg/log"
 )
+
+func mapGRPCToHTTP(code int, err error) int {
+	// Only try to be smart if current code is a generic 500
+	if code != http.StatusInternalServerError {
+		return code
+	}
+
+	// Look for a GRPC error (even a wrapped one) to resolve a more useful HTTP code.
+	// The list of handled codes is intentionally limited to specific cases
+	for currErr := err; currErr != nil; currErr = errors.Unwrap(currErr) {
+		if st, ok := status.FromError(currErr); ok {
+			switch st.Code() {
+			case codes.Canceled:
+				return 499 // Client Closed Request
+			case codes.DeadlineExceeded:
+				return http.StatusGatewayTimeout
+			default:
+				return code
+			}
+		}
+	}
+	return code
+}
 
 const (
 	trillianCommunicationError     = "unexpected error communicating with transparency log"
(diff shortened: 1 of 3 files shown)


pkg/api/error.go
```diff
@@ -60,6 +86,8 @@ func errorMsg(message string, code int) *models.Error {
 var re = regexp.MustCompile("^(.*)Params$")
 
 func handleRekorAPIError(params interface{}, code int, err error, message string, fields ...interface{}) middleware.Responder {
+	code = mapGRPCToHTTP(code, err)
+
 	if message == "" {
```
loosebazooka, line 91: apparently 499s have no default message in the [http.StatusText](https://go.dev/src/net/http/status.go). So if we want to print `Client Closed Context` to logs, we would need to check if code is 499 and then set our own message.

pkg/api/error_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestMapGRPCToHTTP(t *testing.T) {
+	tests := []struct {
+		code int
+		err  error
+		want int
+	}{
+		{http.StatusOK, status.Error(codes.Canceled, "context canceled"), http.StatusOK},
+		{http.StatusInternalServerError, status.Error(codes.Canceled, "context canceled"), 499},
+		{http.StatusInternalServerError, status.Error(codes.DeadlineExceeded, "deadline exceeded"), http.StatusGatewayTimeout},
+		{http.StatusInternalServerError, status.Error(codes.DataLoss, "dataloss"), http.StatusInternalServerError},
```
loosebazooka, line 34: add test for wrapped errors?
```
  {http.StatusInternalServerError, fmt.Errorf("outer: %w", status.Error(codes.Canceled, "ctx")), 499}, 
```

pkg/generated/restapi/configure_rekor_server.go
```diff
@@ -397,7 +400,22 @@ func recoverer(next http.Handler) http.Handler {
 					fields = append(fields, zap.ByteString("request_headers", request))
 				}
 
-				log.ContextLogger(ctx).With(fields...).Errorf("panic detected: %v", rvr)
+				// Check if the panic is due to a connection issue: Don't log these
+				// cases as serious errors
+				isNetworkError := false
```
loosebazooka, line 405: This is wading into "style" territory and I don't care either way, but we can compress this with a lambda or a separate function

```go
  isClientConnError := func(err error) bool {                                                                                                                                                                                                 
      var netErr net.Error                                                                                                                                                                                                                    
      return go_errors.Is(err, io.EOF) || go_errors.Is(err, syscall.EPIPE) || go_errors.Is(err, syscall.ECONNRESET) || (go_errors.As(err, &netErr) && netErr.Timeout())                                                                       
  }                                                                                                                                                                                                                                           
                  
  if err, ok := rvr.(error); ok && isClientConnError(err) {                                                                                                                                                                                   
      log.ContextLogger(ctx).With(fields...).Debugf("client connection closed: %v", rvr)
  } else {                                                                                                                                                                                                                                    
      log.ContextLogger(ctx).With(fields...).Errorf("panic detected: %v", rvr)
  }
```

loosebazooka (approved):
yeah this is sorely needed going through the logs

bobcallaway:
fixed in #2931

PR #2782, opened 2026-03-18
Run liveness checks on uncrowded 2112 port

loosebazooka:
- Readiness should still run on 3000

#### Summary
This might help with liveness checks that are being blocked by an overloaded port 3000 if we end up with a high volume of regular requests and the pod because starved.

Also move off the default global servermux and use an explicit one for utils (metrics, ping)

Readiness and Liveness still both hit `ping` just on different ports. It's possible they should have different names to prevent confusion.

Locally testing this, there are no (but practically probably just fewer) healthz related resets compared to running on 3000, when overloading the local instance running with docker compose.

### DIFF
--- a/cmd/rekor-server/app/serve.go
+++ b/cmd/rekor-server/app/serve.go
@@ -131,10 +131,13 @@ var serveCmd = &cobra.Command{
 		api.ConfigureAPI(treeID)
 		server.ConfigureAPI()
 
-		http.Handle("/metrics", promhttp.Handler())
+		utilMux := http.NewServeMux()
+		utilMux.Handle("/metrics", promhttp.Handler())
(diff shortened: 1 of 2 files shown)


Hayden-IO:
Isn't the issue that the server is processing too many requests? Will the port change help?

loosebazooka:
It's likely I misunderstood how gc assist debt works. It sounds like there might be a very minuscule improvement here due to a lighter middleware stack on port 3000, but not as much as claimed. Each connection is in its own go routine so gc assist debt isn't isolated to http server instances (an assumption that fueled a lot of this experiment). It appears that this may have a little effect and the real reason docker-compose tests were working better was because the compound effect of a few mitigation strategies in [pgi#3612](https://github.com/sigstore/public-good-instance/pull/3612) 

I think I'll just close this for now.

PR #2795, opened 2026-04-13
add defensive check to ensure tid is in config ahead of getting client

bobcallaway:
we know in the configuration which tree IDs are expected to be in trillian so there's no reason to establish a client when we can check locally

### DIFF
--- a/pkg/api/entries_test.go
+++ b/pkg/api/entries_test.go
@@ -1,0 +1,32 @@
+package api
+
+import (
+	"context"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


Hayden-IO (dismissed):

pkg/api/entries_test.go
```diff
@@ -0,0 +1,32 @@
+package api
```
Hayden-IO, line 16: need license

Hayden-IO (approved):

PR #2797, opened 2026-04-13
restapi: include inactiveShards in the homepage total count

### LINKED ISSUE #2281: Info page count does not account for inactive shards
**Description**

See [this javascript](https://github.com/sigstore/rekor/blob/56ea4b5c6382c83795a59253882273ef15a37675/pkg/generated/restapi/rekorHomePage.html#L32-L41), which does not parse the inactive shards object, only the current shard.

**Version**

alliasgher:
## Summary

The HTML homepage at `/` reads `j.treeSize` from `/api/v1/log/` to display "Currently storing N items" but ignores `j.inactiveShards`. After a shard rotation the total count is the sum of the active tree plus every inactive shard, so the page under-reports by the size of all previous shards.

Sum the inactive shards' `treeSize` values into the displayed count, defensively guarding against a missing or non-array `inactiveShards` field.

Fixes #2281

Signed-off-by: Ali <alliasgher123@gmail.com>

### DIFF
--- a/pkg/generated/restapi/rekorHomePage.html
+++ b/pkg/generated/restapi/rekorHomePage.html
@@ -34,7 +34,14 @@ const url = '/api/v1/log/';
 function update() {
   fetch(url, {headers: {'Accept': 'application/json'}}).then((resp) => {
     resp.json().then((j) => {
-      let count = j.treeSize;
+      // Total across the active shard plus any inactive shards; otherwise
+      // the page under-reports after a shard rotation.
+      let count = Number(j.treeSize) || 0;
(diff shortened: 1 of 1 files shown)


bobcallaway:
it looks like you combined https://github.com/sigstore/rekor/pull/2796 and this fix - can you please separate them?

alliasgher:
Separated — updated this PR to only contain the homepage inactiveShards fix. The retry error handler change is in #2796 alone @bobcallaway

bobcallaway (approved):

PR #2799, opened 2026-04-17
Verify returned entry matches client inputs in rekor-cli verify

bobcallaway:
When `rekor-cli verify` searches by artifact, the client now locally canonicalizes the proposed entry and confirms the server's response matches before reporting success.

For all lookup paths (uuid, log-index, artifact), the client now verifies the entry body's leaf hash matches the UUID in the response.

Previously the only binding between the user's local inputs and the verified entry was the server's honesty in answering the search query. The inclusion proof, checkpoint, and SET were all verified, but only against the server-returned body — never against the locally-derived expected entry. This adds client-side verification that closes that gap.

### DIFF
--- a/cmd/rekor-cli/app/verify.go
+++ b/cmd/rekor-cli/app/verify.go
@@ -16,6 +16,9 @@
 package app
 
 import (
+	"bytes"
+	"context"
+	"encoding/base64"
 	"encoding/hex"
 	"fmt"
 	"math/bits"
@@ -104,6 +107,8 @@ var verifyCmd = &cobra.Command{
 		uuid := viper.GetString("uuid")
 		logIndex := viper.GetString("log-index")
 
+		var proposedEntry models.ProposedEntry
+
 		if uuid != "" {
 			searchLogQuery.EntryUUIDs = append(searchLogQuery.EntryUUIDs, uuid)
 		} else if logIndex != "" {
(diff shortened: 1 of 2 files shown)


Hayden-IO:
This issue really makes me think we should be deprecating this client, or at the very least anything with verification. It’s not integrated into the conformance test suite (nor should it be because it shouldn’t be yet another Go client).

bobcallaway:
> This issue really makes me think we should be deprecating this client, or at the very least anything with verification. It’s not integrated into the conformance test suite (nor should it be because it shouldn’t be yet another Go client).

Let's go ahead and fix it and discuss deprecation separately. I know this is packaged and shipped in some distributions.

cmd/rekor-cli/app/verify.go
```diff
@@ -158,12 +163,22 @@ var verifyCmd = &cobra.Command{
 			entry = v
 		}
 
+		if err := verifyBodyMatchesUUID(entry, o.EntryUUID); err != nil {
```
Hayden-IO, line 166: What's the purpose of this check? `o.EntryUUID` and `entry` both come from the server.

Hayden-IO: Should this instead compare `uuid` to the `entry`'s hash when `uuid` is provided?

Hayden-IO (commented):
I think this is a net-positive, but it's still incomplete because for two of the three cases, providing an index or leaf hash, you aren't verifying the leaf hash against the original entry. It's a fine improvement and I'm good with merging, but it's not sufficient for verification. We should clearly define the trust model for using this tool.

cmd/rekor-cli/app/verify.go
bobcallaway: the check is to ensure the response from the server is coherent (body matches leaf hash)

the check on line 171 does what you asked (compares client provided uuid to the entry in the response's hash

Hayden-IO: The underlying issue is the server shouldn’t return a leaf hash because a client should calculate it itself, but we aren’t gonna the API, this is fine

Hayden-IO (approved):

PR #2812, opened 2026-04-28
fix(trillianclient): strip dns:/// scheme from TLS ServerName in gRPC dial

kdacosta0:
#### Summary

When the Trillian gRPC address includes a `dns:///` resolver scheme (used to enable client-side `round_robin` load balancing with headless Kubernetes Services), the `dial()` function passes the raw address, scheme included, into `tls.Config.ServerName` and the gRPC channel authority

This breaks TLS: the x509 certificate verification matches SANs against `"dns"` (the URI scheme) instead of the actual hostname, failing with:

```
x509: certificate is valid for trillian-logserver.ns.svc, not dns
```

The fix strips the `dns:///` prefix into a `cleanHostname` for TLS `ServerName` and `grpc.WithAuthority()`, while preserving the full scheme in the `grpc.NewClient` target so the DNS resolver stays active

`strings.TrimPrefix` is a no-op when the prefix is absent, so callers passing plain hostnames are unaffected

Tested on an OpenShift cluster with internal TLS:
- before: every Trillian gRPC call fails with the x509 error above
- after: TLS handshake succeeds and round-robin load balancing works across all trillian-logserver pod IPs

#### Release Note

Fixed a bug where using the `dns:///` gRPC resolver scheme in the Trillian server address caused TLS certificate verification to fail. The URI scheme was incorrectly used as the TLS ServerName instead of the actual hostname. This affected deployments using client-side gRPC load balancing (`round_robin`) with TLS enabled.

### DIFF
--- a/pkg/trillianclient/manager.go
+++ b/pkg/trillianclient/manager.go
@@ -158,14 +158,19 @@ func CreateAndInitTree(ctx context.Context, config GRPCConfig) (*trillian.Tree, 
 	return t, nil
 }
 
+// cleanDialHostname strips gRPC resolver scheme prefixes (e.g. "dns:///")
+// from a hostname, returning the bare hostname suitable for TLS SNI and
+// certificate verification. The original address with its scheme must be
+// preserved for the grpc.NewClient target so the chosen resolver stays active.
+func cleanDialHostname(hostname string) string {
+	return strings.TrimPrefix(hostname, "dns:///")
+}
+
 func dial(hostname string, port uint16, tlsCACertFile string, useSystemTrustStore bool, serviceConfig string) (*grpc.ClientConn, error) {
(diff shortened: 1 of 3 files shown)


bobcallaway:
could you add a test case here to ensure this doesn't break going forward?

bobcallaway:
> could you add a test case here to ensure this doesn't break going forward?

sorry, i should have been clearer. in `tests/e2e_test.go`, `TestDialE2E` I think can be easily altered to add setting `dns://localhost` as a value (instead of hardcoding it to `localhost` which will provide functional coverage of this.

bobcallaway (approved):

bobcallaway:
can you rebase this please and i'll merge

kdacosta0:
@bobcallaway Rebased and pushed, lmk if you need anything else

PR #2814, opened 2026-05-03
Support restricting kinds on insertion

Hayden-IO:
This adds a server flag to restrict the set of kinds that the server will support for insertion. Note that this does not affect the read path, meaning previously inserted entries are still readable.

This will be used in the public instance to limit the set of kinds to the ones that are actively used to minimize the API attack surface.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -27,6 +27,7 @@ import (
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/sigstore/rekor/pkg/api"
 	"github.com/sigstore/rekor/pkg/log"
+	"github.com/sigstore/rekor/pkg/types"
 	cose "github.com/sigstore/rekor/pkg/types/cose/v0.0.1"
 	intoto001 "github.com/sigstore/rekor/pkg/types/intoto/v0.0.1"
 	intoto002 "github.com/sigstore/rekor/pkg/types/intoto/v0.0.2"
@@ -135,6 +136,11 @@ Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().Int("max_attestation_size", 100*1024, "max size for attestation storage, in bytes")
 
 	rootCmd.PersistentFlags().StringSlice("enabled_api_endpoints", operationIDs, "list of API endpoints to enable using operationId from openapi.yaml")
+
+	rootCmd.PersistentFlags().StringSlice("enabled_entry_types", types.ListSupportedKinds(),
+		"comma-separated list of entry type kinds to enable on the server (e.g. rekord,intoto,dsse,hashedrekord). "+
+			"If unset (default), all types compiled into the binary are enabled. Any kind specified here must be compiled "+
+			"into the binary; unknown kinds will cause startup to fail.")
 
 	rootCmd.PersistentFlags().Uint64("max_request_body_size", 0, "maximum size for HTTP request body, in bytes; set to 0 for unlimited")
 	rootCmd.PersistentFlags().Uint64("max_jar_metadata_size", 1048576, "maximum permitted size for jar META-INF/ files, in bytes; set to 0 for unlimited")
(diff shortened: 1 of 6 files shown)


cmd/rekor-server/app/root.go
```diff
@@ -136,6 +136,11 @@ Memory and file-based signers should only be used for testing.`)
 
 	rootCmd.PersistentFlags().StringSlice("enabled_api_endpoints", operationIDs, "list of API endpoints to enable using operationId from openapi.yaml")
 
+	rootCmd.PersistentFlags().StringSlice("pluggable_types", []string{},
```
Hayden-IO: Note I chose to not deal with version, because only the intoto kind has two versions.

pkg/types/entries.go
```diff
@@ -87,6 +120,9 @@ func CreateVersionedEntry(pe models.ProposedEntry) (EntryImpl, error) {
 	} else {
 		return nil, fmt.Errorf("unknown kind '%v' specified", kind)
 	}
+	if !isKindAllowedForSubmission(kind) {
+		return nil, fmt.Errorf("entry kind '%v' is not enabled for submission on this server", kind)
+	}
```
bobcallaway: move up under `kind := pe.Kind()` to fail faster

cmd/rekor-server/app/root.go
```diff
@@ -136,6 +136,11 @@ Memory and file-based signers should only be used for testing.`)
...
+		"comma-separated list of pluggable type kinds to enable on the server (e.g. rekord,intoto,dsse,hashedrekord). "+
+			"If unset (default), all types compiled into the binary are enabled. Any kind specified here must be compiled "+
+			"into the binary; unknown kinds will cause startup to fail.")
+
```
bobcallaway: given the `enabled_api_endpoints` arg above, should this become `enabled_entry_types` (the term "pluggable" isn't super consistent with our docs) and the default value should auto-populate to the result of `types.TypeMap`'s keys?

Hayden-IO: Good call on the naming.

For the default value, yea, I was debating between empty=all vs explicitly including all. I think I'd have to specify each type string though and not rely on the `TypeMap` keys, which is why I went with empty=all. Whether `TypeMap` is populated is based on `init` order - I can test this but it feels a little brittle to rely on.

bobcallaway: the types are all imported in the same file therefore init order would have already run by the time this code would run, so it should be fine IMO

Hayden-IO: Updated, e2e should pass

Hayden-IO: Removed the added code in entries.go as well that let all kinds through if no allowlist is set.

cmd/rekor-server/app/serve.go
```diff
@@ -150,3 +167,25 @@ var serveCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(serveCmd)
 }
+
+// filterEntryTypes filters the list of compiled-in entry types
+// down to the set requested via the --enabled_entry_types flag.
+// If any requested type is not present, an error is returned.
+// Note that filtering only affects the write path, not the read paths.
+func filterEntryTypes(allTypes map[string][]string, requestedTypes []string) (map[string][]string, error) {
```
bobcallaway: can this just return a `[]string` - you're creating a map just to iterate it above which isn't really necessary

Hayden-IO: the map's values are used to log versions as well

bobcallaway: you have them in entryTypeMap, i guess this is a nit but would reduce a lot of this code

bobcallaway (dismissed):

cmd/rekor-server/app/serve.go
Hayden-IO: I changed it to return a string list, though it's still about the same amount of code to handle duplicates

bobcallaway (approved):

PR #2813, opened 2026-04-30
Update 413 status code error handling

stgrace:
#### Summary

Fixes #2808

Error handling for 413 errors seems to not properly return an error (anymore). When trying to create an entry with too large of a body the following generic errors were returned:

```
Error: signing <image>: Post "https://rekor.sigstore.dev/api/v1/log/entries": POST https://rekor.sigstore.dev/api/v1/log/entries giving up after 4 attempt(s)
error during command execution: signing <image>: Post "https://rekor.sigstore.dev/api/v1/log/entries": POST https://rekor.sigstore.dev/api/v1/log/entries giving up after 4 attempt(s)
```

This PR solves it by returning the proper http error code and a useful message:

```
Error: signing <image>: [POST /api/v1/log/entries][413] createLogEntry default {"code":413,"message":"Request Entity Too Large. Limit set to 10597713"}
error during command execution: <image>: [POST /api/v1/log/entries][413] createLogEntry default {"code":413,"message":"Request Entity Too Large. Limit set to 10597713"}
```

This was tested by running in debug mode with all dependencies running in docker via the docker compose and using cosign v2.6.2:

```
cosign attest --rekor-url=http://localhost:3000 --replace --timeout 10m0s --yes --predicate scan2.json --type spdx <image>
```

#### Release Note

* Fixed error handling for 413 status codes.

#### Documentation

N/A

PR #2841, opened 2026-06-03
Return clear HTTP 413 when request body exceeds max_request_body_size

### LINKED ISSUE #2808: Rekor error handling on max_request_body_size
**Description**

I was trying to upload an attestation using the public Rekor instance, but it kept failing with the following error message:
```
Error: signing <image>: Post "https://rekor.sigstore.dev/api/v1/log/entries": POST https://rekor.sigstore.dev/api/v1/log/entries giving up after 4 attempt(s)
error during command execution: signing <image>: Post "https://rekor.sigstore.dev/api/v1/log/entries": POST https://rekor.sigstore.dev/api/v1/log/entries giving up after 4 attempt(s)
```

After doing quite some extensive investigation by building cosign myself and hosting Rekor locally, I finally figured out that I receive the same error when setting `max_request_body_size`, which defaults to 0 in the code (i.e. no limit).

So not only is the limit for the public instance undocumented, the error itself does not provide any explanation.

I propose some documentation on the limit of the public instance + a clearer error message when the limit is reached.

Happy to create a PR if you guys agree.

**Version**

Cosign: v2.6.3
Rekor: v1.5.1

arpitjain099:
#### Summary

Fixes #2808.

When `max_request_body_size` is configured, the `maxBodySize` middleware in `pkg/generated/restapi/configure_rekor_server.go` wraps the request body with `http.MaxBytesReader`. Once the limit is exceeded the read fails, but the way that failure was surfaced to clients depended entirely on how the error happened to be wrapped before it reached `logAndServeError`:

- a bare `*http.MaxBytesError` fell through to an opaque `500` with body `http: request body too large`
- a top-level `*errors.ParseError` produced a `400` that leaked the same internal `http: request body too large` text
- only a `CompositeError` holding a `ParseError` was converted to `413`, and even then the message was a generic `Request Entity Too Large` that never named the limit

That is why the reporter (uploading with cosign against the public instance) saw only a generic `giving up after N attempt(s)` retry failure with no hint that the body size limit was the cause, and why the limit itself was effectively undiscoverable.

This change detects the underlying `*http.MaxBytesError` across all of those shapes through a single helper (`asMaxBytesError`) and responds with `413 Request Entity Too Large`, including the configured limit in bytes (read straight from `MaxBytesError.Limit`) so operators and clients can see exactly why the request was rejected. `*errors.ParseError` is handled explicitly because it does not implement `Unwrap`, so `errors.As` cannot reach its `Reason` on its own; everything else (bare error, or one nested in a `CompositeError`, which does unwrap) is caught by `errors.As`. The detailed error is still logged server-side, and it is now logged at warn rather than error since it is a client error.

This edit is in the hand-maintained region of `configure_rekor_server.go` (the file carries the go-swagger `This file is safe to edit. Once it exists it will not be overwritten` marker, and the prior 413 handling lived in this same region).

How to verify:

```
go test ./pkg/generated/restapi/ -run 'TestMaxBodySize|TestAsMaxBytesError' -v
```

Behaviour for an over-limit body with the fix applied:

- before: `500 {"code":500,"message":"http: request body too large"}` (or a `400` leaking the same text, depending on the consumer)
- after: `413 {"code":413,"message":"request body exceeds max_request_body_size limit of 1024 bytes"}`

The new tests cover each error shape, an under-limit request (must stay a success, not a false-positive 413), and the detection helper directly.

#### Release Note

Rekor now returns a clear `413 Request Entity Too Large` response that names the configured byte limit when a request body exceeds `max_request_body_size`, instead of an opaque `500`/`400` with no explanation. The detailed error continues to be logged server-side only.

#### Documentation

None.

### DIFF
--- a/pkg/generated/restapi/configure_rekor_server.go
+++ b/pkg/generated/restapi/configure_rekor_server.go
@@ -341,6 +341,14 @@ func wrapMetrics(handler http.Handler) http.Handler {
 
 func logAndServeError(w http.ResponseWriter, r *http.Request, err error) {
 	ctx := r.Context()
+	// If the request body exceeded max_request_body_size, surface a clear
+	// 413 Request Entity Too Large that names the limit instead of letting the
+	// MaxBytesReader error fall through to an opaque 500. The conversion happens
+	// before logging so the request is logged as the client error it is.
+	if maxBytesErr := asMaxBytesError(err); maxBytesErr != nil {
+		err = errors.New(http.StatusRequestEntityTooLarge,
+			"request body exceeds max_request_body_size limit of %d bytes", maxBytesErr.Limit)
+	}
 	if apiErr, ok := err.(errors.Error); ok {
 		code := apiErr.Code()
 		if code >= http.StatusBadRequest && code < http.StatusInternalServerError {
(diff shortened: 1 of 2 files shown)


Hayden-IO:
There is already an open PR for this - https://github.com/sigstore/rekor/pull/2813

arpitjain099:
> There is already an open PR for this - https://github.com/sigstore/rekor/pull/2813

@Hayden-IO I see. Noted. Thanks!

PR #2864, opened 2026-07-01
return 499 if client disconnects instead of 500

bobcallaway:
If a client abruptly disconnects from the server, we currently log and return a 500 error; this swaps the behavior to a 499 response code which is the industry norm for this scenario.


loosebazooka:
A small test verifying the behavior would be nice. mostly just around the context being cancelled?

loosebazooka (approved):

PR #2883, opened 2026-07-09
Migrate off deprecated golang.org/x/crypto/openpgp

### LINKED ISSUE #286: deal with deprecation of x/crypto/openpgp
**Description**

Per https://github.com/golang/go/issues/44226 `golang.org/x/crypto/openpgp` is now frozen and deprecated. We should look to switch to a different implementation to ensure we stay up to date as best we can with this technology.

imjasonh:
## Summary
- Replace deprecated `golang.org/x/crypto/openpgp` with `github.com/ProtonMail/go-crypto` (closes #286).
- Bump `sassoftware/relic` to v8 and switch RPM verification from `cavaliercoder/go-rpm` to `sassoftware/go-rpmutils` so the old openpgp package is fully removed from the dependency graph.
- Adapt PGP verify to ProtonMail's `Config.Time` API (verify at signature creation time so expired keys still work for log replay / existing fixtures), regenerate RPM test fixtures with OpenPGP v4 signatures (v3 RPM sigs are not supported by go-rpmutils without a soft fork), and update email-subject expectations where ProtonMail parses UIDs differently.

## Test plan
- [ ] `go test ./pkg/pki/pgp/ ./pkg/types/rpm/... ./pkg/types/helm/... ./pkg/pki/pkcs7/`
- [ ] Confirm `go list -deps ./...` has no `golang.org/x/crypto/openpgp`
- [ ] Spot-check RPM upload/verify with a modern (v4-signed) package

Made with [Cursor](https://cursor.com)


bobcallaway:
Thanks for the PR - im reviewing to ensure this is as backward compatible as possible, and (more importantly) doesn't break canonicalization logic.

imjasonh:
> Thanks for the PR - im reviewing to ensure this is as backward compatible as possible, and (more importantly) doesn't break canonicalization logic.

Yeah totally makes sense. I admit I don't know enough about this area to speak confidently, but I figured it would be easier to check and talk about it over code instead of in an issue.

I won't feel at all bad if you close this and do it some other way :) 

edit: If there are before/after tests you'd like to see added to this too, to add confidence, just let me know.

pkg/types/rpm/v0.0.1/entry.go
```diff
@@ -259,20 +245,24 @@ func (v *V001Entry) Canonicalize(ctx context.Context) ([]byte, error) {
...
 	}
-	if sha1sum := rpmObj.GetBytes(0, 1012); sha1sum != nil {
-		canonicalEntry.Package.Headers["RPMSIGTAG_SHA1"] = hex.EncodeToString(sha1sum)
+	if sha1sum, err := rpmObj.GetStrings(rpmutils.SIG_SHA1); err == nil && len(sha1sum) > 0 {
+		canonicalEntry.Package.Headers["RPMSIGTAG_SHA1"] = sha1sum[0]
 	}
-	if sha256sum := rpmObj.GetBytes(0, 1016); sha256sum != nil {
-		canonicalEntry.Package.Headers["RPMSIGTAG_SHA256"] = hex.EncodeToString(sha256sum)
+	if sha256sum, err := rpmObj.GetStrings(rpmutils.SIG_SHA256); err == nil && len(sha256sum) > 0 {
+		canonicalEntry.Package.Headers["RPMSIGTAG_SHA256"] = sha256sum[0]
```
bobcallaway: There's a bug in the original code - the header offsets appear to be wrong when reading SHA1 and SHA256 values, so no values would have been written (MD5 appears to be correct). Your code fixes that issue but breaks the canonicalization consistency we need (if clients were using the Canonicalize() method to verify a leaf hash/inclusion proof, they'd end up with a different leaf hash)

we should just comment out those headers with a comment to this effect

imjasonh: https://github.com/sigstore/rekor/pull/2883/changes/fd2c86a5fb06c0e3ebe75dc6c10cb5f22797504d omits these for compatibility, with a test. Let me know if there's anything else you'd like in this change.

bobcallaway (approved):

shanduur:
Is there a release coming with the fix?

PR #2885, opened 2026-07-10
Change default CLI upload type to hashedrekord

Hayden-IO:
#### Summary

This change sets the default `--type` flag to `hashedrekord` in rekor-cli, along with changing the default `--pki-format` flag to `x509` (which is certificates and public keys).

This is done in a backwards compatible way, in that a client can continue to pass an artifact via --artifact and it will automatically be hashed if the type is hashedrekord.

While the PKI flag change is a breaking change, I doubt anyone was using pgp, which was the default.


Hayden-IO:
Verify this works with:

```
openssl genrsa -out private.pem 4096
openssl rsa -in private.pem -pubout -out public.pem
echo "hello world" > artifact.txt
openssl dgst -sha256 -sign private.pem -out signature.sig artifact.txt
rekor-cli upload --artifact artifact.txt --signature signature.sig --public-key public.pem
```

Hayden-IO:
This is finally ready, had to update a bunch of e2e tests.

cmd/rekor-cli/app/pflag_groups.go
```diff
@@ -132,6 +140,30 @@ func CreatePropsFromPflags() *types.ArtifactProperties {
...
+				artifactReader, err = util.FileOrURLReadCloser(context.Background(), props.ArtifactPath.String(), nil)
+			} else {
+				artifactReader, err = os.Open(filepath.Clean(props.ArtifactPath.Path))
+			}
+			if err == nil {
+				defer artifactReader.Close()
+				artifactBytes, err := io.ReadAll(artifactReader)
+				if err == nil {
+					h := sha256.Sum256(artifactBytes)
+					props.ArtifactHash = hex.EncodeToString(h[:])
```
bobcallaway: nit: you could use `io.Copy` into the hasher to avoid reading the entire artifact into memory here

Hayden-IO: done

bobcallaway (approved):
we'll want to doc this in the next changelog

PR #2896, opened 2026-07-23
create interface for TrillianClient

bobcallaway:
this is step 1 of the plan to offer an alternative Trillian client in Rekor that caches the STH and reduces polling to Trillian on each request.

- renames `trillianClientManager.GetTrillianClient` to `trillianClientManager.GetClient`
- creates an `interface` which will provide a consistent facade for consumers
- changes `GetLatest` to stop accepting an initial tree size (since all callers passed `0`)


internal/trillianclient/client.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/google/trillian"
+	"github.com/google/trillian/types"
+	"google.golang.org/grpc/codes"
+)
+
+// Client defines the public API for interacting with a Trillian log.
+// One implementation exists:
+//   - directTrillianClient: stateless, per-RPC client (default)
+type Client interface {
```
Hayden-IO, line 38: Do you want to put this interface and the response struct under an `internal/` directory to avoid exporting them?

pkg/trillianclient/test_helpers_test.go
```diff
@@ -0,0 +1,62 @@
...
+// always terminate once the dial context is canceled, so they are benign. This
+// does not mask a real client leak: a leaked grpc client keeps persistent
+// transport goroutines (loopyWriter, http2 reader) alive, which goleak still flags.
+var grpcDialIgnores = []goleak.Option{
+	goleak.IgnoreAnyFunction("net.(*sysDialer).dialParallel"),
+	goleak.IgnoreAnyFunction("net.(*sysDialer).dialSerial"),
+}
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m, grpcDialIgnores...)
```
Hayden-IO, line 42: What is this for?

bobcallaway: i was seeing some leaks when doing testing with trillian b/c they are still using the older `grpc.Dial` call in their mock setup.

internal/trillianclient/client.go
Hayden-IO: It's more idiomatic to have a top level internal folder rather than nested under pkg, https://github.com/golang-standards/project-layout

Hayden-IO (approved):

PR #2905, opened 2026-07-30
Remove unmaintained scripts/createdb.sh dev helper

### LINKED ISSUE #1066: Errors when using scripts/createdb.sh
**Description**
There are various errors when I try to use `scripts/createdb.sh`. Some of them may be related. I'm using the latest commit with: Server version: 8.0.30 MySQL Community Server - GPL. I am able to get past some of them by fiddling with the database, but it seems like the `createdb.sh` script should make the experience smoother. I can try to create PRs to fix these if there is a consensus around the solutions.

### 1. mysql command does not work without -u and -p flags
This could be seen as a problem with the configuration of the database, but the correct configs to use `createdb.sh` are also not specified. The script fails on [this line](https://github.com/sigstore/rekor/blob/main/scripts/createdb.sh#L26) because mysql access is denied. Instead of using `mysql`, the command `mysql -u root -p` works (when also entering a preset password).

Output:
```
$ ./createdb.sh 
Creating test database and test user account
ERROR 1045 (28000): Access denied for user 'lily'@'localhost' (using password: NO)
Loading table data..
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'test'@'localhost' (using password: YES)

$ sudo ./createdb.sh 
Creating test database and test user account
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
Loading table data..
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'test'@'localhost' (using password: YES)
```

Proposed solution:
- Update the script to be able to use a preset user and password if the database requires it, or
- Document how to set up the db to not require a password

### 2. default password does not meet default db password requirements
The test password specified in the script will error out because it is not long or complex enough when used with default password settings.

Output:
```
mysql > CREATE USER IF NOT EXISTS 'test'@'localhost' IDENTIFIED BY 'zaphod';
ERROR 1819 (HY000): Your password does not satisfy the current policy requirements
```

Proposed solution:
- Update the script's default password to something more complex with special characters, or
- Document how to disable the db password checks for testing

### 3. ROOTPASS is specified but never used
It is specified [here](https://github.com/sigstore/rekor/blob/main/scripts/createdb.sh#L21). Is this deliberate? Or is this supposed to be passed in with the `mysql` command on [line 26](https://github.com/sigstore/rekor/blob/main/scripts/createdb.sh#L26)?

### 4. `mysql -u $USER -p$PASS -D $DB < ./storage.sql` command does not work

Output:
```
$ mysql -u test -p <pw> -D test < ./storage.sql
[1] 472589
bash: -D: command not found...
```

This command works instead (when entering pw manually at the prompt):
`mysql -u test -p test < ./storage.sql`

Proposed solution:
Change the script to use a version of the command above.

pujitha24:
Motivation:
scripts/createdb.sh (and its companion scripts/storage.sql) is a MySQL
dev-database setup script that was copied over from the trillian
project rather than being specific to rekor. Issue #1066 reports
several bugs in it: it fails without manually-supplied MySQL
credentials, its default password doesn't satisfy MySQL's default
password policy, it declares a ROOTPASS variable that is never used,
and its final `mysql -u $USER -p$PASS -D $DB` invocation is broken
because `-p$PASS` swallows the following `-D` argument as part of the
password.

Rekor maintainer bobcallaway commented on the issue: "this script
comes from trillian, and really isn't something custom to Rekor. We
should probably remove it and just refer folks to use what is in the
trillian repo." Trillian carries its own maintained equivalent at
scripts/resetdb.sh in google/trillian.

Approach:
Delete scripts/createdb.sh and scripts/storage.sql. A repo-wide search
found no other references to either file (no docs, Makefile, CI
workflow, or other script mentions them), so removal doesn't leave any
dangling references. This is a dev-convenience script only; it is not
invoked by any build, test, or CI path, so removing it does not change
any user-visible or CI behavior.

Validation:
- go build ./... passes
- Repo-wide grep for "createdb" and "storage.sql" returns no
  remaining references after the removal

Fixes #1066

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>


Hayden-IO (approved):
I'm not aware of any usage of this script either, seems good to clean up!

PR #2907, opened 2026-08-02
index client: fix shadowed err, wrong err label

bobcallaway:
`err` was being shadowed in the assignment in L405, so the success metrics only reported parsing failures, not the db txn.

`addToIndex` was always wrapping its errors with "redis client", even if that was not the configured provider. this was likely leftover from when redis was the only impl

### DIFF
--- a/pkg/api/entries.go
+++ b/pkg/api/entries.go
@@ -402,7 +402,7 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 				log.ContextLogger(ctx).Errorf("getting entry index keys: %v", err)
 				return
 			}
-			if err := addToIndex(context.Background(), keys, entryID); err != nil {
+			if err = addToIndex(context.Background(), keys, entryID); err != nil {
 				log.ContextLogger(ctx).Errorf("adding keys to index: %v", err)
 			}
 		}()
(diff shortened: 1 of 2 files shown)


pkg/api/entries.go
```diff
@@ -402,7 +402,7 @@ func createLogEntry(params entries.CreateLogEntryParams) (models.LogEntry, middl
 				log.ContextLogger(ctx).Errorf("getting entry index keys: %v", err)
 				return
 			}
-			if err := addToIndex(context.Background(), keys, entryID); err != nil {
+			if err = addToIndex(context.Background(), keys, entryID); err != nil {
```
Hayden-IO, line 405: I'm not following this change, why would we write to the instance of `err` declared on line 400 rather than a new instance bound to the conditional?

bobcallaway: because we want the value to be correct in the execution of `defer func()`, and per [spec](https://go.dev/ref/spec#Short_variable_declarations:~:text=Unlike%20regular%20variable,to%20the%20original) the one on L400 is a redeclaration of the one on L393. 

as previously written, `if err := addToIndex(...` is a new, scoped instance that only lives inside the `if` inner block, and the check on L396 would never see the return value of the `addToIndex` call

Hayden-IO: Thanks, missed the defer block.

Hayden-IO (approved):

PR #2914, opened 2026-08-03
separate read/write sql traffic for indexservice into separate pools

bobcallaway:
This refactors IndexStorageProvider in `mysql.go` to separate database connection management into two dedicated `*sqlx.DB` pools: `readDB` and `writeDB`.

1. Prevents Read Starvation / Denial of Service (DoS):
Rekor dispatches entry index writes asynchronously per entry while lookups (`LookupIndices`) run synchronously on the HTTP request path (`/api/v1/index/retrieve`). With a single shared pool, write bursts could consume all open connections. Because `database/sql` selects waiting goroutines randomly, read requests faced unbounded latency or request context timeouts. Splitting pools guarantees dedicated capacity for reads.

2. Backward-Compatible Connection Sizing:
Legacy connection limits (`search_index.mysql.max_open_connections / max_idle_connections`) are automatically split 70% write / 30% read in `indexstorage.go` to preserve the total connection budget without exceeding server limits.

3. DB Observability per Pool:
A custom Prometheus collector `metrics.go` was added to export Prometheus metrics tagged with `pool="read"` and `pool="write"`.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -149,8 +149,21 @@ Memory and file-based signers should only be used for testing.`)
 	rootCmd.PersistentFlags().String("search_index.mysql.dsn", "", "DSN for index storage using MySQL")
 	rootCmd.PersistentFlags().Duration("search_index.mysql.conn_max_idletime", 0*time.Second, "maximum connection idle time")
 	rootCmd.PersistentFlags().Duration("search_index.mysql.conn_max_lifetime", 0*time.Second, "maximum connection lifetime")
-	rootCmd.PersistentFlags().Int("search_index.mysql.max_open_connections", 0, "maximum open connections")
-	rootCmd.PersistentFlags().Int("search_index.mysql.max_idle_connections", 0, "maximum idle connections")
+	rootCmd.PersistentFlags().Int("search_index.mysql.read.max_open_connections", 0, "maximum open connections in the index read pool; 0 for unlimited")
+	rootCmd.PersistentFlags().Int("search_index.mysql.read.max_idle_connections", 0, "maximum idle connections retained in the index read pool")
+	rootCmd.PersistentFlags().Int("search_index.mysql.write.max_open_connections", 0, "maximum open connections in the index write pool; 0 for unlimited")
+	rootCmd.PersistentFlags().Int("search_index.mysql.write.max_idle_connections", 0, "maximum idle connections retained in the index write pool")
+
+	// superseded by the per-pool flags above; these sized the single pool that predates the
+	// read/write split and are now split 70/30 between the write and read pools
+	rootCmd.PersistentFlags().Int("search_index.mysql.max_open_connections", 0, "deprecated: total provided will be split 70/30 across write/read pools; use per-pool flags instead")
+	rootCmd.PersistentFlags().Int("search_index.mysql.max_idle_connections", 0, "deprecated: total provided will be split 70/30 across write/read pools; use per-pool flags instead")
+	for _, f := range []string{"search_index.mysql.max_open_connections", "search_index.mysql.max_idle_connections"} {
+		if err := rootCmd.PersistentFlags().MarkDeprecated(f, "value is split 70/30 between the write and read pools; "+
+			"use the search_index.mysql.read.* and search_index.mysql.write.* equivalents"); err != nil {
+			log.Logger.Fatal(err)
+		}
+	}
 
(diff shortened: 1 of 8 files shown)


cmd/rekor-server/app/root.go
```diff
@@ -149,8 +149,21 @@ Memory and file-based signers should only be used for testing.`)
...
 	rootCmd.PersistentFlags().Duration("search_index.mysql.conn_max_idletime", 0*time.Second, "maximum connection idle time")
 	rootCmd.PersistentFlags().Duration("search_index.mysql.conn_max_lifetime", 0*time.Second, "maximum connection lifetime")
-	rootCmd.PersistentFlags().Int("search_index.mysql.max_open_connections", 0, "maximum open connections")
-	rootCmd.PersistentFlags().Int("search_index.mysql.max_idle_connections", 0, "maximum idle connections")
+	rootCmd.PersistentFlags().Int("search_index.mysql.read.max_open_connections", 0, "maximum open connections in the index read pool; 0 for unlimited")
+	rootCmd.PersistentFlags().Int("search_index.mysql.read.max_idle_connections", 0, "maximum idle connections retained in the index read pool")
+	rootCmd.PersistentFlags().Int("search_index.mysql.write.max_open_connections", 0, "maximum open connections in the index write pool; 0 for unlimited")
+	rootCmd.PersistentFlags().Int("search_index.mysql.write.max_idle_connections", 0, "maximum idle connections retained in the index write pool")
+
+	// superseded by the per-pool flags above; these sized the single pool that predates the
```
Hayden-IO, line 157: Would it be cleaner to just drop these flags? A breaking change at server startup seems fine. And I'm skeptical anyone would be setting these flags, as this is most meaningful for tuning high-qps deployments.

pkg/indexstorage/mysql/mysql.go
```diff
@@ -39,47 +40,85 @@ const (
...
-	if _, err := provider.db.Exec(createTableStmt); err != nil {
-		return nil, fmt.Errorf("create table if not exists failed: %w", err)
+	provider.collector = &dbStatsCollector{readDB: provider.readDB, writeDB: provider.writeDB}
+	if err := prometheus.DefaultRegisterer.Register(provider.collector); err != nil {
+		provider.collector = nil
+		var alreadyRegistered prometheus.AlreadyRegisteredError
+		if !errors.As(err, &alreadyRegistered) {
+			return nil, errors.Join(fmt.Errorf("registering db stats collector: %w", err), provider.Shutdown())
+		}
+		log.Logger.Warnf("search index db stats already registered by another provider; this provider's pools will not be reported")
```
Hayden-IO: When does this occur?

pkg/indexstorage/mysql/options.go
```diff
@@ -79,38 +82,69 @@ func WithConnMaxLifetime(lifetime time.Duration) RequestConnMaxLifetime {
...
+func WithReadMaxOpenConns(openConns int) RequestMaxOpenConns {
+	return RequestMaxOpenConns{role: poolRead, openConns: openConns}
+}
+
+// WithWriteMaxOpenConns specifies the maximum number of open connections in the write pool
+func WithWriteMaxOpenConns(openConns int) RequestMaxOpenConns {
+	return RequestMaxOpenConns{role: poolWrite, openConns: openConns}
+}
+
+// poolRole identifies which of the provider's two connection pools is being configured.
```
Hayden-IO, line 137: This is just for metrics, correct?

pkg/indexstorage/mysql/mysql.go
bobcallaway: it wont, removed

cmd/rekor-server/app/root.go
bobcallaway: cleaner, yes - but i was trying to avoid this since it is best practice to set them

pkg/indexstorage/mysql/options.go
bobcallaway: `poolRole` is primarily the dispatch key for option targeting: `applyMaxIdleConns/applyMaxOpenConns` compare `role == r.role` so a `WithReadMaxOpenConns` option is a no-op against the write pool. It also tags the openPool error messages. `String()` feeding
  the `pool=` metric label is the secondary use.

cmd/rekor-server/app/root.go
Hayden-IO: Could we delete them with the next 1.x release?

Hayden-IO (approved):

cmd/rekor-server/app/root.go
bobcallaway: yeah, i think so.

PR #2918, opened 2026-08-05
remove unused semver type version map

bobcallaway:
none of the types ever made use of semver matching. to prune the dependency graph, re-implementing this with a simple map lookup.


Hayden-IO (approved):
Nice

PR #2919, opened 2026-08-06
build(deps): Migrate gopkg.in/yaml.v2 to sigs.k8s.io/yaml

chimanjain:
#### Summary

Migrate the deprecated `gopkg.in/yaml.v2` to `sigs.k8s.io/yaml`.
#### Release Note

NONE


pkg/sharding/ranges_test.go
```diff
@@ -43,9 +43,9 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/stretchr/testify/require"
+	"go.yaml.in/yaml/v3"
```
Hayden-IO: We're already using `sigs.k8s.io/yaml`, can we rely on that instead?

Hayden-IO (commented):
thanks!

pkg/sharding/ranges_test.go
chimanjain: Yes, I think it would work, but it is not a drop-in replacement, shall I try it?

chimanjain: I've refactored the code to use the `sigs.k8s.io/yaml` instead, please review.

chimanjain: I will update the description and title also.

bobcallaway (approved):

PR #2920, opened 2026-08-06
feat: serve the API over TLS with a configurable TLS policy

### LINKED ISSUE #2904: Support serving the Rekor API over TLS, with a configurable TLS policy
## Summary

`rekor-server serve` can only expose the API over plaintext HTTP. go-swagger, which generates Rekor's server, already implements an HTTPS listener and a `configureTLS` hook, but Rekor never enables the former or populates the latter.  This issues asks for both, serving TLS on the API listener, and a way to set a minimum TLS version and TLS 1.2 cipher suites on it.

## Current behaviour

The listener scheme is [hardcoded](https://github.com/sigstore/rekor/blob/main/cmd/rekor-server/app/serve.go#L140) in , so `https` can never be enabled:

```go
// cmd/rekor-server/app/serve.go
```
Certificate material can already be supplied today:

- Via CLI flag: `--tls-certificate`, `--tls-key`, `--tls-ca`, `--tls-port` are registered in `pkg/generated/restapi/server.go`'s `init()`. Cobra picks these up automatically, so they already show up in `rekor-server --help`.
- Via env var, as a fallback: `TLS_CERTIFICATE`, `TLS_PRIVATE_KEY`, `TLS_CA_CERTIFICATE`, `TLS_PORT`.

Either way, the value correctly lands on the server struct  but with no `https` listener enabled, it's inert.

The same applies to `--scheme`: it does populate `EnabledListeners` when `restapi.NewServer` builds the server, but `serve.go` immediately overwrites it with `["http"]`, discarding whatever scheme was requested. So the gap isn't that the flags don't reach the CLI, it's that hardcoded assignment always wins.

There's also no flag or config key feeding go-swagger's TLS policy hook:

    // pkg/generated/restapi/configure_rekor_server.go
    func configureTLS(_ *tls.Config) {
        // Make all necessary changes to the TLS configuration here.
    }

Note: `trillian_log_server.tls` and `redis_server.enable-tls` configure Rekor as a client to those backends; neither affects the API listener above.

## Motivation

- The final hop to the Rekor process is unavoidably plaintext in reencrypt/passthrough load-balancing setups without a service mesh.
- Compliance baselines (FIPS 140-x, FedRAMP, PCI-DSS) require a stated TLS version/cipher floor on the listener itself, not just "TLS somewhere in front of it."
- On an untrusted internal network, a plaintext listener exposes log traffic to passive observation and lets an on-path attacker tamper with requests/responses in flight (detectable by a correct client via inclusion proofs, but still a confidentiality/availability gap worth closing).

## Proposed change

1. **Make the listener scheme configurable** (e.g. `--scheme`), feeding go-swagger's existing `EnabledListeners`/`TLSCertificate`/`TLSCertificateKey`/`TLSCACertificate` fields from Rekor's own config, instead of hardcoding `http`.
2. **Wire a TLS policy** — add `--tls-min-version` and `--tls-cipher-suites`, applied inside `configureTLS`. Cipher-suite selection only affects TLS 1.2 handshakes (TLS 1.3 suites aren't configurable in Go); the min-version flag mainly matters for pinning 1.3 or making the floor explicit and auditable.

Defaults to current behaviour (HTTP only), so no impact on existing deployments.

## Scope

- **In:** serving TLS on the API listener; min-version and TLS 1.2 cipher-suite configuration.
- **Optional:** mTLS via `--tls-ca`, already supported by go-swagger once the listener exists.
- **Out:** the separate `:2112` metrics endpoint; Rekor's outbound TLS to Trillian/Redis (already configurable); certificate rotation (go-swagger loads the key pair once at startup).

## Acceptance criteria

1. No config change → behaves exactly as today (plaintext HTTP).
2. Scheme + cert/key configured. HTTPS served on the configured port; plaintext rejected on it.
3. Configured min version / TLS 1.2 cipher list is enforced against a client that doesn't meet it.
4. `https` with no certificate fails fast with a clear error, not a broken listener.
5. New flags and the TLS 1.3 cipher-suite caveat are documented with the other rekor-server settings.

## Prior art

`sigstore/timestamp-authority` also uses go-swagger and doesn't hardcode the listener. `scheme := viper.GetStringSlice("scheme")` feeds `EnabledListeners` directly (Viper-only, no registered flag). This request goes one step further and makes the scheme and TLS policy first-class `rekor-server` configuration.

anithapriyanatarajan:
## Summary

Closes #2904.

`rekor-server serve` could only expose the API over plaintext HTTP. The listener scheme was hardcoded to `["http"]` in `cmd/rekor-server/app/serve.go`, so go-swagger's existing HTTPS listener and its `configureTLS` hook were never
reached, even though the `TLSCertificate` / `TLSCertificateKey` / `TLSCACertificate` / `TLSPort` fields already existed on the generated server.

This PR makes the scheme configurable and wires a TLS policy, defaulting to the current behaviour (HTTP only) so existing deployments are unaffected.

## Changes

- **Configurable scheme** via `--scheme`, feeding go-swagger's `EnabledListeners` instead of the hardcoded slice.
- **TLS material as first-class Rekor config**: `--tls-certificate`, `--tls-key`, `--tls-ca` (optional mutual TLS), `--tls-port`. Existing `TLS_*` environment  fallbacks are preserved and only overridden when the matching config key is set.
- **TLS policy**: `--tls-min-version` (1.2 or 1.3) and `--tls-cipher-suites`, applied through the existing `configureTLS` hook. Parsing/validation lives in a new `internal/tlspolicy` package.
- **Fail-fast validation**: unknown TLS versions and insecure/unknown cipher suite names are rejected at startup, and `--scheme https` without a certificate/key exits with a clear error rather than starting a broken listener.
- **Docs**: commented, disabled-by-default TLS block added to `rekor-server.yaml`; the TLS-1.3 default and the TLS-1.2-only cipher-suite caveat are documented in the flag help.

Cipher-suite selection applies to TLS 1.2 only.  The minimum version defaults to TLS 1.3 for the HTTPS listener when unset.

## Backward compatibility

No behavioural change unless `https` and a certificate are explicitly configured.
HTTP-only deployments continue to work exactly as before; the policy flags are optional.

## How to test

- `go test ./internal/tlspolicy/...` - unit tests for version/cipher parsing and
  policy application.
- Manually: `rekor-server serve --scheme http --scheme https --tls-certificate server.crt --tls-key server.key` and confirm the API is served over HTTPS on the TLS port and that a sub-minimum client handshake is refused.

## Release Note

```release-note
Added TLS support for the Rekor API listener and a configurable TLS policy. New `serve` flags: `--scheme` selects the listeners (add `https` to serve TLS), `--tls-certificate`/`--tls-key`/`--tls-ca`/`--tls-port` supply the certificate material, `--tls-min-version` sets the minimum TLS version (1.2 or 1.3, defaults to 1.3), and `--tls-cipher-suites` restricts the allowed TLS 1.2 cipher suites. All flags are opt-in; existing HTTP-only deployments are unaffected.
```


cmd/rekor-server/app/serve.go
```diff
@@ -137,7 +139,36 @@ var serveCmd = &cobra.Command{
...
+		}
+
+		policy, err := tlspolicy.Parse(viper.GetString("tls-min-version"), viper.GetStringSlice("tls-cipher-suites"))
+		if err != nil {
+			log.Logger.Fatalf("invalid TLS policy: %v", err)
+		}
+		tlspolicy.Set(policy)
+
+		if slices.Contains(schemes, "https") && (server.TLSCertificate == "" || server.TLSCertificateKey == "") {
+			log.Logger.Fatal("scheme 'https' requires both --tls-certificate and --tls-key (or the TLS_CERTIFICATE and TLS_PRIVATE_KEY environment variables)")
```
Hayden-IO, line 170: To confirm, `server.TLSCertificate`/`server.TLSCertificateKey` would already be populated if the environment var is set?

internal/tlspolicy/policy.go
```diff
@@ -0,0 +1,129 @@
+//
+// Copyright 2025 The Sigstore Authors.
```
Hayden-IO: 2025 -> 2026

internal/tlspolicy/policy_test.go
```diff
@@ -0,0 +1,154 @@
+//
+// Copyright 2025 The Sigstore Authors.
```
Hayden-IO: 2025 -> 2026

internal/tlspolicy/policy.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// Policy is a parsed, validated TLS policy for the API listener.
+type Policy struct {
+	MinVersion   uint16
+	CipherSuites []uint16
+}
+
+var (
+	mu      sync.RWMutex
```
Hayden-IO: is a mutex needed?

internal/tlspolicy/policy.go
```diff
@@ -0,0 +1,129 @@
...
+	if len(names) == 0 {
+		return nil, nil
+	}
+
+	secure := make(map[string]uint16)
+	for _, c := range tls.CipherSuites() {
+		secure[c.Name] = c.ID
+	}
+	insecure := make(map[string]struct{})
+	for _, c := range tls.InsecureCipherSuites() {
```
Hayden-IO: can we just treat these as unknown, and throw an error if it's not in the list from `CipherSuites`?

internal/tlspolicy/policy.go
```diff
@@ -0,0 +1,129 @@
...
+			return nil, fmt.Errorf("unknown TLS cipher suite %q: valid values are %s", name, strings.Join(secureCipherSuiteNames(), ", "))
+		}
+		ids = append(ids, id)
+	}
+	return ids, nil
+}
+
+func secureCipherSuiteNames() []string {
+	names := make([]string, 0)
+	for _, c := range tls.CipherSuites() {
```
Hayden-IO: small cleanup, you already iterated this list on line 96, so you can build the list there

Hayden-IO (dismissed):

cmd/rekor-server/app/root.go
```diff
@@ -180,6 +180,15 @@ Memory and file-based signers should only be used for testing.`)
 	keyAlgorithmHelp := fmt.Sprintf("signing algorithm to use for signing/hashing (allowed %s)", strings.Join(keyAlgorithmTypes, ", "))
 	rootCmd.PersistentFlags().StringSlice("client-signing-algorithms", keyAlgorithmTypes, keyAlgorithmHelp)
 
+	// TLS on the API listener; add "https" to --scheme to serve TLS.
+	rootCmd.PersistentFlags().StringSlice("scheme", []string{"http"}, "listeners to enable on the API; add 'https' to serve TLS (e.g. --scheme http --scheme https)")
```
```suggestion
	rootCmd.PersistentFlags().StringSlice("scheme", []string{"http"}, "listeners to enable on the API; add 'https' to serve TLS (e.g. --scheme http,https)")
```

cmd/rekor-server/app/root.go
```diff
@@ -180,6 +180,15 @@ Memory and file-based signers should only be used for testing.`)
...
+	rootCmd.PersistentFlags().String("tls-certificate", "", "the certificate file to use for serving TLS on the API listener")
+	rootCmd.PersistentFlags().String("tls-key", "", "the private key file (without passphrase) to use for serving TLS on the API listener")
+	rootCmd.PersistentFlags().String("tls-ca", "", "the certificate authority file used to verify client certificates for mutual TLS on the API listener")
+	rootCmd.PersistentFlags().Int("tls-port", 0, "the port to serve TLS on for the API listener; defaults to a random port when unset")
+	rootCmd.PersistentFlags().String("tls-min-version", "", "minimum TLS version for the API listener (1.2 or 1.3); when unset defaults to 1.3")
```
```suggestion
	rootCmd.PersistentFlags().String("tls-min-version", "1.3", "minimum TLS version for the API listener (1.2 or 1.3); when unset defaults to 1.3")
```

cmd/rekor-server/app/root.go
```diff
@@ -180,6 +180,15 @@ Memory and file-based signers should only be used for testing.`)
...
+	rootCmd.PersistentFlags().StringSlice("tls-cipher-suites", nil, "allowed TLS 1.2 cipher suite names (crypto/tls spelling) for the API listener; empty keeps the built-in default. Has no effect on TLS 1.3.")
```
bobcallaway: id add 1.2 into the arg name

bobcallaway: also, is there a way to list out the valid values here? redirecting to go doc isn't a great UX imo

cmd/rekor-server/app/serve.go
anithapriyanatarajan: Yes. The env vars populate the cert and key.

internal/tlspolicy/policy.go
anithapriyanatarajan: Thank you for pointing out. Have removed this now.

cmd/rekor-server/app/root.go
```diff
@@ -180,6 +180,15 @@ Memory and file-based signers should only be used for testing.`)
 	keyAlgorithmHelp := fmt.Sprintf("signing algorithm to use for signing/hashing (allowed %s)", strings.Join(keyAlgorithmTypes, ", "))
 	rootCmd.PersistentFlags().StringSlice("client-signing-algorithms", keyAlgorithmTypes, keyAlgorithmHelp)
 
+	// TLS on the API listener; add "https" to --scheme to serve TLS.
```
bobcallaway, line 184: macro-comment: some of those flags are already defined in pkg/generated/restapi/server.go - have you reconciled how they interact?

cmd/rekor-server/app/serve.go
```diff
@@ -137,7 +139,36 @@ var serveCmd = &cobra.Command{
...
+		if v := viper.GetString("tls-certificate"); v != "" {
+			server.TLSCertificate = v
+		}
+		if v := viper.GetString("tls-key"); v != "" {
+			server.TLSCertificateKey = v
+		}
+		if v := viper.GetString("tls-ca"); v != "" {
+			server.TLSCACertificate = v
+		}
+		if v := viper.GetInt("tls-port"); v != 0 {
```
bobcallaway:
```suggestion
		if v := viper.GetUInt("tls-port"); v != 0 {
```

internal/tlspolicy/policy.go
```diff
@@ -0,0 +1,109 @@
...
+		return Policy{}, err
+	}
+	return Policy{MinVersion: mv, CipherSuites: cs}, nil
+}
+
+// parseMinVersion maps "1.2"/"1.3" to their crypto/tls constants, defaulting to
+// TLS 1.3.
+func parseMinVersion(v string) (uint16, error) {
+	switch strings.TrimSpace(v) {
+	case "", "1.3":
```
bobcallaway: if you set the default as 1.3, remove the empty string here

cmd/rekor-server/app/root.go
anithapriyanatarajan: Done. Thank you.

cmd/rekor-server/app/root.go
anithapriyanatarajan: Fixed to default 1.3

internal/tlspolicy/policy.go
anithapriyanatarajan: Done. Thank you

cmd/rekor-server/app/root.go
anithapriyanatarajan: image
 
listed the values in --help. Hope this is acceptable

anithapriyanatarajan: > id add 1.2 into the arg name

Could we keep the name generic (--tls-cipher-suites)?. Today Go ignores this field for 1.3 for TLS 1.3 suite set , so the flag only meaningfully applies to TLS 1.2  but that's a Go implementation detail that could change. If Go ever makes 1.3 suites configurable, the generic name still fits.

cmd/rekor-server/app/root.go
anithapriyanatarajan: Verified that, though both the go-swagger-generated code (server.go) and the app command code (root.go) declare the same flags, pflag does not register a flag name twice, and the app flags are registered first, so those are the ones used on the CLI, and the swagger-generated same-named flags are ignored.

`tls-min-version` and `tls-cipher-suites` are not part of the generated server, so they don't get swagger's env handling. Instead they're validated in tlspolicy and applied through the configureTLS hook, and can be set via CLI flag, config file, or env var (TLS_MIN_VERSION / TLS_CIPHER_SUITES).

cmd/rekor-server/app/root.go
Hayden-IO: Does 1.3 ignore the cipher suite? What's [this list](https://pkg.go.dev/crypto/tls#:~:text=//%20TLS%201.3%20cipher%20suites.%0A%09TLS_AES_128_GCM_SHA256%20%20%20%20%20%20%20uint16%20%3D%200x1301%0A%09TLS_AES_256_GCM_SHA384%20%20%20%20%20%20%20uint16%20%3D%200x1302%0A%09TLS_CHACHA20_POLY1305_SHA256%20uint16%20%3D%200x1303) referring to?

cmd/rekor-server/app/root.go
```diff
@@ -180,10 +181,33 @@ Memory and file-based signers should only be used for testing.`)
...
+	tlsCipherSuitesHelp := fmt.Sprintf("allowed cipher suites for the API listener (comma-separated); applies to TLS 1.2 only and has no effect on TLS 1.3, whose suites are fixed by Go. Empty keeps the Go default. Allowed values: %s", strings.Join(tlspolicy.CipherSuiteNames(), ", "))
+
+	rootCmd.PersistentFlags().StringSlice("scheme", []string{"http"}, "listeners to enable on the API; add 'https' to serve TLS (e.g. --scheme http,https)")
```
Hayden-IO, line 187: Do we need to support both concurrently, or should it be one or the other?

anithapriyanatarajan: Yeah, you are right. Even though they are on different ports, there is practically no point instantiating both a secure and an insecure endpoint. I will add a validation to make it either/or.

cmd/rekor-server/app/root.go
anithapriyanatarajan: Based on comments [here](https://pkg.go.dev/crypto/tls#Config.CipherSuites) ciphersuites are not configurable for TLS1.3. Hence, in this case, even if some value is set, it will be ignored for TLS1.3

cmd/rekor-server/app/root.go
```diff
@@ -181,10 +182,31 @@ Memory and file-based signers should only be used for testing.`)
 	keyAlgorithmHelp := fmt.Sprintf("signing algorithm to use for signing/hashing (allowed %s)", strings.Join(keyAlgorithmTypes, ", "))
 	rootCmd.PersistentFlags().StringSlice("client-signing-algorithms", keyAlgorithmTypes, keyAlgorithmHelp)
 
+	// TLS on the API listener.
+	tlsCipherSuitesHelp := fmt.Sprintf("allowed cipher suites for the API listener (comma-separated); applies to TLS 1.2 only and has no effect on TLS 1.3, whose suites are fixed by Go. Empty keeps the Go default. Allowed values: %s", strings.Join(tlspolicy.CipherSuiteNames(), ", "))
+
+	rootCmd.PersistentFlags().StringSlice("scheme", []string{"http"}, "listener scheme for the API; 'http' or 'https' (mutually exclusive). Set 'https' to serve TLS")
```
bobcallaway, line 188: if these are mutually exclusive this should be String() not StringSlice()

cmd/rekor-server/app/serve.go
```diff
@@ -139,7 +141,35 @@ var serveCmd = &cobra.Command{
...
+		// Override cert fields only when set, preserving the TLS_* env fallbacks.
+		if v := viper.GetString("tls-certificate"); v != "" {
+			server.TLSCertificate = v
+		}
+		if v := viper.GetString("tls-key"); v != "" {
+			server.TLSCertificateKey = v
+		}
+		if v := viper.GetString("tls-ca"); v != "" {
+			server.TLSCACertificate = v
+		}
```
bobcallaway, line 159: should we file.CleanPath() these?

cmd/rekor-server/app/serve.go
```diff
@@ -166,6 +196,31 @@ func init() {
...
 }
 
+// validateSchemes enforces the API listener scheme policy: only http, https,
+// and unix are recognized; plaintext http and TLS https are mutually exclusive
+// (serving both would leave a plaintext hop open); and https requires a
+// certificate/key pair.
+func validateSchemes(schemes []string, tlsCert, tlsKey string) error {
+	for _, s := range schemes {
+		switch s {
+		case "http", "https", "unix":
```
bobcallaway, line 206: what is unix here?

internal/tlspolicy/policy.go
```diff
@@ -0,0 +1,133 @@
...
+// starts serving, so no synchronization is required.
+var current Policy
+
+// Set stores the policy applied by Apply.
+func Set(p Policy) {
+	current = p
+}
+
+// Apply updates cfg from the stored policy, leaving unset fields unchanged.
+func Apply(cfg *tls.Config) {
```
bobcallaway, line 44: needs defensive check to see if cfg is `nil`

cmd/rekor-server/app/root.go
anithapriyanatarajan: fixed.

cmd/rekor-server/app/serve.go
anithapriyanatarajan: I am sorry.  This was generated based on go-swagger default. Has no meaning here. Hence removed.

cmd/rekor-server/app/serve.go
```diff
@@ -166,6 +196,22 @@ func init() {
...
 }
 
+// validateScheme accepts only http and https; https also requires a
+// certificate/key pair.
+func validateScheme(scheme, tlsCert, tlsKey string) error {
+	switch scheme {
+	case "http":
+		return nil
+	case "https":
+		if tlsCert == "" || tlsKey == "" {
```
bobcallaway, line 206: should you test existence of these files here to ensure they're valid before passing this down into an error case which might not be fatal?

bobcallaway: please address this and the lint error and this should be good to go

bobcallaway:
https://github.com/sigstore/rekor/actions/runs/34998262460/job/104482321669?pr=2920#step:6:249

anithapriyanatarajan:
> https://github.com/sigstore/rekor/actions/runs/34998262460/job/104482321669?pr=2920#step:6:249

This is fixed now and ready to review. :bowing_woman:

bobcallaway (approved):

PR #2922, opened 2026-08-09
remove repetitive viper lookups from api endpoints

bobcallaway:
viper lookups cause unnecessary allocation churn, and since they are write-once we should keep them as vars instead of re-looking up over & over

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -50,6 +50,12 @@ type API struct {
 	// so we can fetch the checkpoint on service startup to
 	// minimize signature generations
 	cachedCheckpoints map[int64]string
+	// Config snapshotted at startup. viper rebuilds a flat map of every bound
+	// pflag on each lookup of a dotted key, so reading these per-request
+	// allocates for values that cannot change once the server is running.
+	checkpointHostname    string
+	publishEventsProtobuf bool
+	publishEventsJSON     bool
 }
 
 func (api *API) ActiveTreeID() int64 {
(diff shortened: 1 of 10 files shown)


pkg/types/jar/v0.0.1/entry.go
```diff
@@ -45,13 +45,21 @@ import (
...
 )
 
 const (
 	APIVERSION = "0.0.1"
 )
 
+// Zero means unlimited. rekor-server overrides this from config at startup;
+// leaving the default at zero preserves the unbounded behavior other callers
+// (e.g. rekor-cli) get when the flag was never registered with viper.
+var maxJarMetadataSize uint64
```
Hayden-IO: Could this be under an internal package instead, something like `internal/config/limits.go`? I want to avoid exporting public setters.

In `cmd/rekor-server/app/root.go`, you could do something like `config.MaxJarMetadataSize = viper.GetUint64("max_jar_metadata_size")` and access `config.MaxJarMetadataSize` below.

pkg/api/api.go
```diff
@@ -186,6 +190,15 @@ var (
 	indexStorageClient       indexstorage.IndexStorage
 )
 
+// Immutable config snapshotted by NewAPI. viper rebuilds a flat map of every
+// bound pflag on each lookup of a dotted key, so reading these per-request
+// allocates for a value that cannot change after startup.
+var (
```
Hayden-IO: Should these be fields in the `API` struct rather than package variables?

Hayden-IO (approved):

Issue #68, opened 2020-12-07
ERROR: logging before flag.Parse


lukehinds:
When starting rekor, the following error appears:

```
go run cmd/server/main.go serve
2020-12-07T11:38:46.453Z	INFO	app/root.go:97	Using config file: /home/luke/go/src/github.com/lukehinds/rekor/rekor-server.yaml
ERROR: logging before flag.Parse: I1207 11:38:46.458754   34544 admin.go:151] Initialising Log 847182197821869483...
ERROR: logging before flag.Parse: I1207 11:38:46.461381   34544 admin.go:162] Initialised Log (847182197821869483) with new SignedTreeHead:
key_hint:"\x0b\xc1˸i5\r\xab" log_root:"\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00 \xe3\xb0\xc4B\x98\xfc\x1c\x14\x9a\xfb\xf4șo\xb9$'\xaeA\xe4d\x9b\x93L\xa4\x95\x99\x1bxR\xb8U\x16Nl=\x0b\x8b(D\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" log_root_signature:"0E\x02 d\xcd\xc0\xac\xbd\x99x\xc4%`[k\x8c\xbaz\xe4\xf0q\xe68?Rn\xb9\xbc)\x1f\x99͵\x9b\xa7\x02!\x00\xe3ښ\xef\x83\xe0\xe5\xbf;\xe2\xf9\xe2\x8b}\xaf\xb3b\xc3+T:\x0b\xa4\x9a\x85\xa6I\xe6\xed\xce\x1a\xd9"
2020-12-07T11:38:46.462Z	INFO	api/server.go:55	Starting server...
2020-12-07T11:38:46.462Z	INFO	api/server.go:61	Listening on 127.0.0.1:3000
```

Spent a while looking into this and trying to find the root cause, it seems to be an upstream issue in glog , see [here](https://github.com/GoogleContainerTools/kaniko/issues/478)

@dlorenc you have a better insight into go packaging, any ideas?

bobcallaway:
fixed in 7128cd1fa4dd35132a3207ce12093ea6549bf184

PR #2932, opened 2026-08-10
fix pprof 

bobcallaway:
Previously we enabled pprof in `init()`. However, the command line arguments had not been parsed then, so the value was always false. This fixes it appropriately.

### DIFF
--- a/cmd/rekor-server/app/root.go
+++ b/cmd/rekor-server/app/root.go
@@ -185,30 +185,43 @@ Memory and file-based signers should only be used for testing.`)
 	}
 
 	rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+}
 
-	log.Logger.Debugf("pprof enabled %v", enablePprof)
-	// Enable pprof
-	if enablePprof {
-		go func() {
-			mux := http.NewServeMux()
-
-			mux.HandleFunc("/debug/pprof/", pprof.Index)
-			mux.HandleFunc("/debug/pprof/{action}", pprof.Index)
-			mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
-
-			srv := &http.Server{
-				Addr:         ":6060",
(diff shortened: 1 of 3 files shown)


cmd/rekor-server/app/root.go
```diff
@@ -185,30 +185,43 @@ Memory and file-based signers should only be used for testing.`)
...
+		mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+		mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+		srv := &http.Server{
+			Addr: ":6060",
+			// No WriteTimeout: a CPU profile or trace streams for its full
+			// requested duration (30s by default, often longer), and any write
+			// deadline shorter than that truncates the response into an
+			// unparseable profile.
+			ReadHeaderTimeout: 10 * time.Second,
```
Hayden-IO, line 217: ReadTimeout or ReadHeaderTimeout?

cmd/rekor-server/app/serve.go
```diff
@@ -70,6 +70,8 @@ var serveCmd = &cobra.Command{
 		// Setup the logger to dev/prod
 		log.ConfigureLogger(viper.GetString("log_type"), viper.GetString("trace-string-prefix"))
 
+		StartPprof()
```
Hayden-IO, line 73: should this happen after `flag.CommandLine.Parse`?

bobcallaway: nope; flag.CommandLine.Parse([]string{}) is stdlib flag which was a glog workaround for #68 — it exists solely to mark the stdlib flag set as parsed so glog stops complaining.

cmd/rekor-server/app/root.go
bobcallaway: all pprof calls just do GET w/ no body so readheader is strictly more accurate afaik

Hayden-IO (approved):

PR #2943, opened 2026-08-17
move CI containers to GHCR

bobcallaway:
we can publish to ghcr to be consistent & save $

also adds missing backfill index ref


Hayden-IO (approved):
fwiw, i'm not sure if per-commit container builds are really needed, so we could also probably just remove this

PR #2946, opened 2026-08-21
dsse parsing cpu/memory optimization

bobcallaway:
Reuses the serialized DSSE envelope for parsing and hashing, and extracts in-toto subject/material digests in one JSON pass. Large-payload benchmarks show approximately 33% lower allocation and 34% lower execution time, while preserving best-effort indexing for non-object predicates.

```
  Payload       Before       After    Time reduction    Memory reduction
  ━━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━
   256 KiB     2.935 ms    1.925 ms             34.4%               33.2%
  ─────────  ───────────  ──────────  ────────────────  ──────────────────
     1 MiB    11.333 ms    7.482 ms             34.0%               33.3%
  ─────────  ───────────  ──────────  ────────────────  ──────────────────
    32 MiB     365.1 ms    240.9 ms             34.0%               33.3%

  Allocations also fell by roughly 9–10 per operation.
```

### DIFF
--- a/pkg/types/dsse/v0.0.1/entry.go
+++ b/pkg/types/dsse/v0.0.1/entry.go
@@ -64,17 +64,19 @@ type V001Entry struct {
 	isInsertable       bool
 }
 
-type indexKeyExtract struct {
-	Subject []struct {
-		Digest map[string]string `json:"digest"`
-	} `json:"subject"`
-	Predicate json.RawMessage `json:"predicate"`
+type digestExtract struct {
+	Digest map[string]string `json:"digest"`
 }
 
-type materialsExtract struct {
-	Materials []struct {
-		Digest map[string]string `json:"digest"`
-	} `json:"materials"`
+type subjectExtract struct {
+	Subject []digestExtract `json:"subject"`
+}
+
+type indexKeyExtract struct {
+	Subject   []digestExtract `json:"subject"`
+	Predicate *struct {
+		Materials []digestExtract `json:"materials"`
+	} `json:"predicate"`
 }
 
 func (v V001Entry) APIVersion() string {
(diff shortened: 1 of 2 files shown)


pkg/types/dsse/v0.0.1/entry.go
```diff
@@ -317,20 +322,31 @@ func (v *V001Entry) Unmarshal(pe models.ProposedEntry) error {
...
-		if err := json.Unmarshal(decodedPayload, &extract); err == nil {
-			for _, s := range extract.Subject {
-				for alg, ds := range s.Digest {
-					v.extractedIndexKeys = append(v.extractedIndexKeys, alg+":"+ds)
-				}
+		if err := json.Unmarshal(decodedPayload, &extract); err != nil {
+			// Predicate can legally have an arbitrary shape. Retry without decoding it
+			// so valid subject digests are still indexed.
+			var subjects subjectExtract
+			if err := json.Unmarshal(decodedPayload, &subjects); err == nil {
```
Hayden-IO, line 329: Why do we have to try to decode again? Isn't `indexKeyExtract` a superset of `subjectExtract`, the former just being a subject+predicate?

bobcallaway: im optimizing for the default case (well-formed provenance/sbom). if the payload does not have the exact contents of `"predicate": { "materials": [ { "subject": [ { "alg": "deadbeef..." }, ... ] } ] }` the first `Unmarshal` will fail. 

if the decodedPayload only has a `subject` array, the second parse will succeed with minimal memory footprint.

this mimics the prior logic which is why i wrote it this way

Hayden-IO (approved):

pkg/types/dsse/v0.0.1/entry.go
Hayden-IO: from offline chat - `materials` is defined by the predicate type, not by the envelope type. as a follow up, i'd prefer we match on well-known predicate types (e.g. slsa v0.2, slsa v1.0), and otherwise fallback to parsing only the subject when the predicate type is not known.

also materials was renamed to resolvedDependencies in the slsa v1.0 predicate.

PR #2962, opened 2026-08-24
use vendored pgp v3 signature parsing instead of fork

bobcallaway:


pkg/pki/pgp/internal/sigv3/LICENSE
Hayden-IO: nit, for project structure, can this go under either a top-level vendor/ or third_party/ directory, following https://github.com/golang-standards/project-layout advice? I believe vendor is preferred, as that is natively supported by Go tooling, but this isn't a fully vendored in dependency, so not sure if that's the right place.

pkg/pki/pgp/internal/sigv3/sigv3.go
```diff
@@ -0,0 +1,307 @@
...
+// against its exported API so no forked module is required.
+//
+// Only verification of a v3 signature made by a v4 public key is supported;
+// v3 public keys and v3 signature generation are deliberately omitted.
+package sigv3
+
+import (
+	"bytes"
+	"crypto"
+	"crypto/dsa" //nolint:staticcheck // required to verify legacy DSA signatures already in the log
```
Hayden-IO: Could we confirm this with the deps.dev data? I'd hope we don't have any DSA signatures in the log and we can just drop this. If it'll be too much work, it's fine, just another motivation for deprecation (see https://words.filippo.io/dsa/).

pkg/pki/pgp/internal/sigv3/LICENSE
Hayden-IO: Can this go under internal/ at the very least so it's not publicly exported?

PR #2988, opened 2026-09-18
fix(intoto): canonicalize the public key in v0.0.2 entries

### LINKED ISSUE #1170: PEM encoded certificate does not always end with a newline character when returned from Rekor
**Description**

For some entry types (it seems?), the uploaded PEM encoded certificate is not returned with an ending newline (this may require that the uploaded certificate did not end with a newline). I tried to recreate this with sigened blob/record type, but even if I uploaded a PEM encoded certificate with the ending newline missing, it was added back by Rekor (which I think is the right thing to do). The rekor cli also does this, so I had to `curl` a POST request with a certificate missing the ending newline. 

This entry https://rekor.sigstore.dev/api/v1/log/entries?logIndex=5450282 contains a certificate that does not end with a newline.
(Omitting a newline is actually breaking the POSIX standard, as a text file is a set of lines, where each line must end with a newline character.)

When uploading a (badly formatted) PEM encoded certificate where the ending newline character is missing, my belief is that Rekor *should* accept it and append the missing newline according to the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle), but it seems it only does so for certain kinds?

This was discovered by @bdehamer when working on implementing the sigstore bundle for JavaScript

**Version**

Public good Rekor as deployed at October 19th 2022.

Eljees:
Fixes #1170.

### The behaviour

`Canonicalize()` for `intoto` v0.0.2 builds the canonical entry by copying `Content.Envelope.Signatures` straight from the client-submitted object. The public key inside each signature therefore reaches storage exactly as the client sent it, so a certificate that is valid PEM but not canonical — most visibly one without a trailing newline — is persisted and returned that way.

Two places in the same codebase already avoid this and show the intended shape:

* `hashedrekord`'s `Canonicalize()` re-encodes the key through `keyObj.CanonicalValue()` every time;
* `CreateFromArtifactProperties()` in this very file does the same when the entry is built from artifact properties, so an entry can come out canonical or not depending on which path created it.

### Is it still live?

The issue is from 2022 and the thread ends on the open question of whether it still matters after the move to `dsse`. It does: I reproduced it on today's `main` (`b7e5e99`). The new test fails there and passes with this change; the rest of `pkg/types/intoto/...` is unaffected either way.

### The change

Re-derive each signature's key through `x509.NewPublicKey(...).CanonicalValue()` before assembling the canonical entry, and fail with a clear error if a signature or its key is missing. `Keyid` and `Sig` are carried over untouched. +28/−1 in `pkg/types/intoto/v0.0.2/entry.go`.

### The test

`TestV002Entry_Canonicalize_PEMTrailingNewline` submits an entry whose certificate lacks the trailing newline, canonicalizes it, and asserts the stored key is the canonical encoding rather than the bytes that came in. I checked the failure direction by keeping the test and reverting only `entry.go`: it fails on unpatched `main`, which is what makes it a regression test rather than a description of current behaviour.

`gofmt`, `go vet ./pkg/types/intoto/...` and `go test ./pkg/types/intoto/...` are clean on Go 1.26.7 against `main` at `b7e5e99`.


Hayden-IO:
Since this would change canonicalization logic, this would break verification for older entries. Therefore, we can't make this change.

Eljees:
Understood, and agreed: the canonical bytes are what gets hashed into the log, so re-encoding the key would change the leaf hash of every v0.0.2 entry whose client sent a non-canonical PEM, and inclusion proofs for those entries would stop verifying. That outweighs the consistency with `hashedrekord` this was going for. Closing.

If it helps, #1170 could probably be closed with the same reasoning, since any fix there has the same shape.

